PR #455, opened 2015-01-07
Back to specific directory for ADD command

huntc:
My PR https://github.com/sbt/sbt-native-packager/pull/424 moved to a `ADD` wildcard approach, and then removed the Dockerfile. There were two problems with this:

* `ADD` seems to behave weirdly regarding recursion. I recently observed that `/opt/docker` and `/docker` were both added to the output.
* I had incorrectly set the current working directory to the parent of the context directory, where it really should be the context directory itself. This only worked because there was a Dockerfile in the parent also. The fix is to ensure that the context directory is properly set.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -115,11 +115,12 @@ object DockerPlugin extends AutoPlugin {
         Some(Cmd("MAINTAINER", maintainer))
     }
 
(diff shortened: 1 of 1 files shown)


huntc:
Provisionally released as `1.0.0-5f0e74fcdfa43d9d654e590694bddef526fd94ac`

huntc:
`scripted docker/*` passes. Using Docker `1.4.1` but I do not believe any specific version is required.

huntc:
Closing - as I was using the wrong branch. Will open another.

Issue #451, opened 2015-01-05
SystemV init script is not idempotent


OlegIlyenko:
I'm creating a debian package with SystemV  (init.d) init script in it. The problem is that I able to `sudo service my-app start` several times in a row and it will start several processes. I expect it to start my service the first time and fail all consequent starts because service is already started and running.

The actual issue is in the way services are started. It uses `--exec` [here](https://github.com/sbt/sbt-native-packager/blob/master/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/systemv/start-debian-template#L33):

    start-stop-daemon --background --chdir ${{chdir}} --chuid "$DAEMON_USER" --make-pidfile --pidfile "$PIDFILE" --exec "$RUN_CMD" --start -- $RUN_OPTS

The problem is that in order to check whether the service is running, `--exec` not only checks the PID file, but also the `/proc/[PID]/exe` file (see [this article](https://chris-lamb.co.uk/posts/start-stop-daemon-exec-vs-startas) for more detail). This is a problem, because the actual process is a `java` process, but what I give to the `start-stop-daemon` command is an application name like `my-service`.

I looked at this issue and found 2 possible solutions. One would be to use `--startas` instead of `--exec` like described in article I linked above. Another solution would be to use  `exec -a <actual-service-name> "$@"` in the application script:

https://github.com/sbt/sbt-native-packager/blob/master/src/main/resources/com/typesafe/sbt/packager/archetypes/bash-template#L136

in order to change the process name.

muuki88:
Thanks for the detailed bug report! What are the difference between your solutions?

Using the `start-stop-daemon` solution would mean that only the debian/SystemV packages
would benefit, right? The new script would look something like

```bash
start-stop-daemon --background \
--chdir ${{chdir}} --chuid "$DAEMON_USER"  \
--make-pidfile --pidfile "$PIDFILE" \
--startas "$RUN_CMD" --start -- $RUN_OPTS
```

The second solution, using `exec -a` means that the bashscript itself starts a named
process, which _should_ make the `start-stop-daemon` behave as expected.

```bash
exec -a ${{app_name}} "$@" 
```

What would you suggest? PRs welcome :)

OlegIlyenko:
Hi, thanks for the fast feedback! Yes, you described it correctly. I generally would prefer the second solution (`exec -a`), because `--startas` has some downsides (and `exec -a` makes nice process names). 

[...]

PR #457, opened 2015-01-07
Using `--startas` instead of `--exec` now in order to make start script idempotent.

OlegIlyenko:
This will only use PID file to check whether application is started or not. `--exec` in contrast is also checks the `/proc/[PID]/exe` file which links to the java executable and not original start script (see [this article](https://chris-lamb.co.uk/posts/start-stop-daemon-exec-vs-startas)). 

For more detail please see the discussion on #451

I tried to use another solution (`exec -a "appName"`), but unfortunately it does not work as I expected. So as far as i can tell `--startas` is the only option.


muuki88:
Awesome, thanks :)

PR #471, opened 2015-01-26
Removed grep invocations from rpm postun script

jkramarz:
Let's do it in simpler & more robust way...


typesafehub-validator:
Hi @jkramarz,

kardapoltsev:
LGTM. Lets wait for travis (:

muuki88:
Nice. I think we should change this in [control-functions](https://github.com/sbt/sbt-native-packager/blob/master/src/main/resources/com/typesafe/sbt/packager/linux/control-functions) as well, or better remove this from `rpm` and just use the `add/delete Group/User` functions.

jkramarz:
@muuki88: Ok, I'll take a bit deeper look at this part of code, so let's close this PR without merging for now.

muuki88:
I think the refactoring didn't finish this last piece.

```bash
if getent passwd "${{daemon_user}}" > /dev/null;
```

is just redudant, so you can this check in the `control-functions` and
in the `rpm` scriptlets.

PR #478, opened 2015-02-04
[Fix #476] Fixing -jvm-debug flags so that they use args compatible with the latest JDKs

### LINKED ISSUE #476: Using the -jvm-debug flag should pass the -agentlib jvm flag instead of -Xrunjdwp flag
It looks like the jvm debugger options passed into the java application are of the type:

```bash
-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=12345  
```

which is the outdated JDK 1.4 way of doing things.

Shouldn't it be using these args instead:

```bash
-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=12345
```

I couldn't get remote debugging with IntelliJ working with my application which uses JDK 7 with the args that the native packager uses for `-jvm-debug`, but could get it working with the latter set of args

jateenjoshi:
When providing the -jvm-debug flag the sbt native packager's bash template currently passes the following jvm debug flags:

-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=[debug port]

This is outdated (it is for JVM 1.4.x) and for newer JVMs should be:

-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=[debug port]

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/bash-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/bash-template
@@ -148,8 +148,7 @@ addResidual () {
   residual_args+=( "$1" )
 }
 addDebugger () {
-  addJava "-Xdebug"
-  addJava "-Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=$1"
+  addJava "-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=$1"
 }
 # a ham-fisted attempt to move some memory settings in concert
 # so they need not be messed around with individually.


typesafehub-validator:
Hi @jateenjoshi,

jateenjoshi:
This change simply changes the -jvm-debug flag being passed through a bash script - there's no reason why this should fail a build - especially on a particular jdk8 build.

Is this a known issue with the build? The jdk8 build seems to be unable to resolve:

org.scala-lang#scala-actors;2.10.2!scala-actors.jar

muuki88:
Travis _sometimes_ fails. I retriggered the job. We don't have any tests that cover this, so everything is fine.

thanks for your work :)

jateenjoshi:
Thanks, np :)

Issue #385, opened 2014-10-23
Add command in Dockerfile


c4po:
Can we add a new setting key to hold some commands in Dockerfile?
for example, we need run some commands to create a couple of folders before start the application.
or change the application configure file based on docker environment value.

raittes:
+1

PR #486, opened 2015-02-11
Wip/docker refactoring

muuki88:
First rework of the docker API according to #453 

Reviews appreciated @fiadliel , @huntc , @hseeberger

PR #488, opened 2015-02-12
proposition for #385 -- adding raw commands

andypetrella:
Hello,

This is a small PR for the issue #385. I did it rather straightforward because it's a rather broad need.

I introduced `Raw` for a partial script to be included in the `Dockerfile` then two keys:

* `dockerRawWithOriginalUser` for commands **before** `USER $daemonUser`
* `dockerRawWithDaemonUser` for commands **after** `USER $daemonUser`

FYI: I need it the for my [SparkNotebook](https://github.com/andypetrella/spark-notebook) for which I'd like to be able to map volumes from the host (for instance). 
I had a [question](http://stackoverflow.com/questions/28405699/spark-notebook-worksheets-not-saved-with-docker) regarding this and I'm struggling with it for two days, and actually a solution would be to use a less restrictive [user](http://chapeau.freevariable.com/2014/08/docker-uid.html).

Note: I'm sure I did it on the right branch (... from master → 0.7, and there is a 0.8 one I see).

tell me what you think
thanks
andy

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -122,8 +122,8 @@ object DockerPlugin extends AutoPlugin {
     val dockerCommands = Seq(
       Cmd("ADD", s"$files /$files"),
       Cmd("WORKDIR", "%s" format dockerBaseDirectory),
-      ExecCmd("RUN", "chown", "-R", daemonUser, "."),
       Raw(dockerRawWithOriginalUser),
+      ExecCmd("RUN", "chown", "-R", daemonUser, "."),
       Cmd("USER", daemonUser),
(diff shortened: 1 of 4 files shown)


typesafehub-validator:
Hi @andypetrella,

muuki88:
Thanks for your pull request :) We are currently changing the docker internals here #486 
I think you can now achieve your goal with the new exposed API.

```scala
dockerCommands := {
  val cmds = dockerCommands.value
  // add, filter, remove docker commands as you like
  cmds
}

```

The [new documentation](https://github.com/sbt/sbt-native-packager/blob/wip/docker-refactoring/src/sphinx/formats/docker.rst) gives a bit more details.

andypetrella:
@muuki88 ah great, so should I work on this branch atm?

Also, do you have any mechanism (shortcuts) to create user and to set him the rights to execute the packaged app (a play one in my case)?

Thanks a lot for the information.
andy

PR #491, opened 2015-02-13
Revert "[fix #472] /etc/default/<package-name> should be shell script se...

### LINKED ISSUE #472: "/etc/default/<package-name>" should be shell script setting envars

muuki88:
Reverts sbt/sbt-native-packager#473


muuki88:
@dhardy92 while testing your pr I found that the unversial/docker packagings don't work out of the box anymore :( 

I'll revert your PR until we can figure out why.

Issue #82, opened 2013-11-20
Specify default memory parameters of java_application BASH script in the build


abrutus:
I downloaded a sample app for the latest version of play
packaged it using debian:packageBin and then installed it in my VM with 512 megs of ram
but when I try to run it I get this

    Error occurred during initialization of VM
    Could not reserve enough space for object heap

While you can specify `-mem` to the script to fix this, it'd be nice if the default configuration for memory requirements was configurable.

huntc:
I'd like to re-visit this issue. As more microservice style applications are developed then having a default of `-Xms1024m -Xmx1024m` seems excessive. I recently observed a medium sized EC2 machine running out of space within only 3 services running.

Why not just lean on the default `mx` of Java which is 64m respectively. `ms` could still equal `mx`.

Developers need to think about the memory required for their apps. Unfortunately the current settings do little to provoke that thought in practice.

muuki88:
@huntc the easiest way of doing the shell part would be to remove the `-mem` option. It adds a lot of possible errors in the bash script and doesn't add any convenience as the jvm figures out most of the stuff itself.

@jsuereth I really would like to remove this part as it only gets more tricky with each additional request.

kardapoltsev:
+1 for removing `-mem`.

PR #494, opened 2015-02-15
Added translation of `jvmopts` contents and enhanced example app.

metasim:
Added this so standard JVM options encoded in the `con/jvmopts` file are translated and passed on to `javapackager`.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/jdkpackager/JDKPackagerHelper.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/jdkpackager/JDKPackagerHelper.scala
@@ -10,7 +10,7 @@ import sbt._
  */
 object JDKPackagerHelper {
 
-  // Try to compute determine a default path for the java packager tool.
+  /** Attempts to compute the path to the `javapackager` tool. */
   def locateJDKPackagerTool(): Option[File] = {
     val jdkHome = sys.props.get("java.home").map(p ⇒ file(p))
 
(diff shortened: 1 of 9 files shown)


src/main/scala/com/typesafe/sbt/packager/jdkpackager/JDKPackagerHelper.scala
```diff
@@ -62,40 +42,79 @@ object JDKPackagerHelper {
...
-    def mainClassArg = mainClass
+    val mainClassArg = mainClass
       .map(c ⇒ Map("-appclass" -> c))
       .getOrElse(Map.empty)
 
-    //    val cpSep = sys.props("path.separator")
-    //    val cp = classpath.map(p ⇒ "lib/" + p)
-    //    val cpStr = cp.mkString(cpSep)
+    // Make a setting?
+    val jvmOptsFile = (sourceDir ** "jvmopts").getPaths.headOption.map(file)
```
muuki88: jvmOpts, etc-default and so forth have been part of a lot of discussions now.
We still need some work here, but this affects sbt-native-packager as a whole

src/main/scala/com/typesafe/sbt/packager/jdkpackager/JDKPackagerKeys.scala
```diff
@@ -38,5 +44,12 @@ trait JDKPackagerKeys {
       |    * `icns`: MacOS
       |    * `ico`: Windows
       |    * `png`: Linux
+      |
+      | Defaults to generic Java icon.
     """.stripMargin)
+
+  // To consider:
```
muuki88: See comment above

muuki88:
LGTM. Currently I have a lot more time, so I hope to get this released soon.

src/main/scala/com/typesafe/sbt/packager/jdkpackager/JDKPackagerHelper.scala
metasim: My vote would be for a setting that specifies the JVM options as a `Seq[String]`, and not rely on some external source file with some one-off format (fine with me if it's converted to a source file as a part of `universal:packageBin`.) Almost went that route with this plugin, but I worked hard to avoid introducing any new constructs here.

src/main/scala/com/typesafe/sbt/packager/jdkpackager/JDKPackagerKeys.scala
metasim: Would be interested in knowing if there's already been a debate between an external file vs. explicit settings in SBT. The latter also allows for platform specific scoping.

muuki88: There has been, ticket is still open #82

src/main/scala/com/typesafe/sbt/packager/jdkpackager/JDKPackagerHelper.scala
muuki88: >  but I worked hard to avoid introducing any new constructs here
And you have done a great job here :) 

The application configuration has a real big scope. Hopefully there is some time
to finish the discussion in #82 and find a proper solution. I like your idea of generating
the `jvmopts` or `etc-default` files from build settings.

PR #496, opened 2015-02-16
Fix #495. Enhanced `locateJDKPackagerTool` to work on Windows when SBT...

### LINKED ISSUE #495: Default value for `jdkPackagerTool` is computed incorrectly on Windows.
The `locateJDKPackagerTool()` did work on windows, I had to set the setting with:

```scala
jdkPackagerTool := Some(file("C:\\Program Files\\Java\\jdk1.8.0_31\\bin\\javapackager.exe"))
```

metasim:
...is running in the parallel installed JRE.


src/main/scala/com/typesafe/sbt/packager/jdkpackager/JDKPackagerHelper.scala
```diff
@@ -11,19 +15,45 @@ import sbt._
...
-      .map(f ⇒ f / "bin" / toolname)
-      .filter(_.exists())
+
+    // This approach to getting JDK bits is borrowed from: http://stackoverflow.com/a/25163628/296509
+    // Starting with an ordered list of possible java directory sources, create derivative and
+    // then test for the tool. It's nasty looking because there's no canonical way of finding the
+    // JDK from the JRE, and JDK_HOME isn't always defined.
+    Seq(
+      // Build-defined
+      javaHome,
```
muuki88: Nice, I didn't know that setting. One couldn't believe how hard it is to find the JDK on a system -.-

PR #523, opened 2015-03-17
Fix incorrect information in RPM Plugin docs.

myyk:
I believe this is what was meant to be said. To see what's on master as it is now see: http://www.scala-sbt.org/sbt-native-packager/formats/rpm.html?highlight=template#scriptlet-changes

It doesn't look correct to me.


muuki88:
That's true. Thanks :)

PR #526, opened 2015-03-17
Added support for environment configuration file.

knshiro:
Added the possibility to create a configuration file to be sourced before running the app.


muuki88:
Awesome :)
Tests and docs will be done in a second PR .
cc @kardapoltsev

PR #533, opened 2015-03-27
!Update default dockerBaseImage to official repo

danielwegener:
according to https://blog.docker.com/2015/03/updates-available-to-popular-repos-update-your-images/, all dockerfile/* repos will be replaced with an official repository. This may be a better default configuration.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -60,7 +60,7 @@ object DockerPlugin extends AutoPlugin {
   override def requires = universal.UniversalPlugin
 
   override lazy val projectSettings = Seq(
-    dockerBaseImage := "dockerfile/java:latest",
+    dockerBaseImage := "java:latest",
     dockerExposedPorts := Seq(),
     dockerExposedVolumes := Seq(),
     dockerRepository := None,


muuki88:
Awesome, thanks. Where you able to run the tests? `sbt "scripted docker/*"`

muuki88:
Note: Reading the comments on the docker repo, it seems like only open-jdk is supported. However updating is reasonable IMHO

danielwegener:
This was just a ping as I stumbled across the blog post. The PR comes from the github ui and I have not yet tried to sbt-native-packager at all :)
But I can try it later in the evening.

muuki88:
Ah, github is amazing :) 

I tested this on my machine with docker 1.5 and everything works fine

muuki88:
Thanks :)

danielwegener:
Awesome :)

dhpiggott:
Does the base image need to include a JDK? I've been using just the JRE image without any problems, but I note that the 'latest' tag of the new official Docker repo (https://registry.hub.docker.com/_/java/) is an alias for openjdk-7-*jdk* (https://github.com/docker-library/java/blob/4c5beb8bdb21c746bef683f18bdeddee157f61fd/openjdk-7-jdk/Dockerfile), whereas the 'latest' tag of the old Dockerfile repo (https://registry.hub.docker.com/u/dockerfile/java/) was an alias for openjdk-7-*jre*.

Might it be worth changing it to the 'jre' version, which is an alias for https://github.com/docker-library/java/blob/4c5beb8bdb21c746bef683f18bdeddee157f61fd/openjdk-7-jre/Dockerfile?

PR #535, opened 2015-03-28
Improvements to RPM documentation

dwhjames:
- `name in Rpm` is not actually used, instead it is the `packageName in Rpm` setting that is relevant.
- describe the relations between sbt settings and the rpm file name
- describe the relations between sbt settings and the rpm spec file
- remove listings of settings that are not in fact required settings
- change default value of `rpmRelease` from `0` to `1`


muuki88:
LGTM. Thanks a lot :)

Issue #436, opened 2014-12-13
systemd service is not ran as the created user


nefilim:
```
[Service]
Type=simple
WorkingDirectory=/usr/local/myapp
ExecStart=/usr/local/myapp/bin/myapp
Restart=always
RestartSec=60
User=myapp <== MISSING
```

is there a way to instruct SBT to add the ```User=``` directive?

PR #540, opened 2015-03-30
test added for daemon user in systemd #436

kardapoltsev:


muuki88:
LGTM. Travis seems to have a lot of issues fetching debian dependencies lately :(

PR #541, opened 2015-04-04
fix SystemV init script template for rpm packaging

dwhjames:
Main fix is to make use of the `daemon` library function in `start`, correcting broken invocation of `runuser`.

For processes that manage their own pid file, they should define the `PIDFILE` variable (sourced either through `/etc/default/…` or `/etc/sysconfig/…`). Otherwise, if undefined, this script will capture the pid and write it to `/var/run/${{app_name}}/running.pid`.

Other minor changes:

- don’t need to call `success`/`failure` in `start` as this is already done by `daemon`
- don’t need to create `/var/run/${{app_name}}` dir as this is already done by the linux package mappings and thus the dir is created at package install time.

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/etc-default-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/etc-default-template
@@ -22,6 +22,10 @@
 # -----------------
 # JAVA_OPTS="-Dpidfile.path=/var/run/${{app_name}}/play.pid $JAVA_OPTS"
 
+# Setting PIDFILE
+# ---------------
+# PIDFILE="/var/run/${{app_name}}/play.pid"
+
 # export env vars for 3rd party libs
 # ----------------------------------
 # COMPANY_API_KEY=123abc
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/systemv/start-rpm-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/systemv/start-rpm-template
@@ -34,7 +34,7 @@
 
 # Source from sysconfig
 # This order means system config appends/overrides package config
-[ -e /etc/sysconfig/$prog ] && . /etc/sysconfig/$prog
+[ -e /etc/sysconfig/${{app_name}} ] && . /etc/sysconfig/${{app_name}}
 
 exec="${{chdir}}/bin/${{exec}}"
 prog="${{app_name}}"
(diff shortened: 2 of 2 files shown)


dwhjames:
@muuki88, I’ve tested this with Centos 6.6 and a sample Play app. I can package up my semi-automated test environment and submit it as a PR, if you’d like.

I’ve tested the script in the two cases with respect to pid file creation.

First with the following in the `/etc/default/…`
```bash
JAVA_OPTS="-Dpidfile.path=/var/run/${{app_name}}/play.pid $JAVA_OPTS"
PIDFILE="/var/run/${{app_name}}/play.pid"
```

In this case, this value of `PIDFILE` is used by `daemon`, `kill_proc`, and `status`.

In the second case, I left `PIDFILE` undefined, which results in the init script setting
```bash
PIDFILE="/var/run/${{app_name}}/running.pid"
```
and appending `echo $! > $PIDFILE` to the command run by `daemon`.

src/main/resources/com/typesafe/sbt/packager/archetypes/etc-default-template
```diff
@@ -22,7 +22,11 @@
 # -----------------
 # JAVA_OPTS="-Dpidfile.path=/var/run/${{app_name}}/play.pid $JAVA_OPTS"
 
+# Setting PIDFILE
+# ---------------
+# PIDFILE="/var/run/${{app_name}}/play.pid"
```
muuki88: getting the config stuff right already paid off. Sweet!

src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/systemv/start-rpm-template
```diff
@@ -28,62 +28,48 @@
...
 
-    retval=$?   # last error code
-    PID=$!      # pid of last backgrounded process
-    [ $retval -eq 0 ] && touch ${lockfile} && success || failure
+# If program manages its own PID file then it
+# should declare its location in PIDFILE
+if [ -z "${PIDFILE:-}" ]; then
+    PIDFILE=/var/run/${{app_name}}/running.pid
+    # echo $! must run in the shell started by `runuser` in `daemon`
+    RUN_CMD="$RUN_CMD echo \$! > $PIDFILE"
```
muuki88: I like extending the `RUN_CMD` this way. Nice work

muuki88:
I will give this a try hopefully in the next days. Really appreciate your work. Thanks :)

dwhjames:
Thanks! It’s been a good opportunity to brush up on Linux and Bash.

PR #543, opened 2015-04-05
Optimize Travis script

dwhjames:
- silence info-level logging for dependency resolving and downloading in update
- split tasks for update, compile, test:compile, so that sbt will show task completion times for each
- join test and scripted tasks into one sbt invocation
- use bash && so that failure stops the build


muuki88:
Nice. Thanks :)

Issue #529, opened 2015-03-23
daemon user removed while perform rpm upgrading


jilen:
I have a play 2.3.7 project with following configs
```
maintainer in Linux := "xxx"

packageSummary in Linux := "yyy"

packageDescription := "zzz"

rpmRelease := {
  val sdf = new java.text.SimpleDateFormat("yyMMddHHmm")
  sdf.format(new java.util.Date())
}

rpmVendor := "aaa"

rpmUrl := Some("bbb")

rpmLicense := Some("MIT")

```
Note that `rpmRelease` is set to current time

when I try to upgrade the rpm package via `yum install  xxx_201503221000.rpm` , the daemon user is removed and not recreated after upgrading

muuki88:
What system do you use and which rpm-installer (e.g. yum) do you use?
Upgrades [should be handled](https://github.com/sbt/sbt-native-packager/blob/master/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/rpm/postun-template#L7) in the `postun` script.

jilen:
@muuki88  I am doing upgrade with `yum install`

dwhjames:
@jilen, could you give the fix in #544 a try?

jilen:
Confirmed. This is fixed `in play 2.4`

PR #544, opened 2015-04-05
fix for issue #529

dwhjames:
- use numeric rather than string test in conditional
- add missing semicolon


muuki88:
Things are still working for me and fix looks good, so I'll merge this.

Issue #563, opened 2015-04-26
JDKPackagerPlugin can't package ScalaFX application 


strobe:
Got this error even if main class was set at build.sbt manually (on OSX).

```
[error] Exception in thread "main" com.sun.javafx.tools.packager.PackagerException: Error: Unknown argument: hello.ScalaFXHelloWorld
[error] 	at com.sun.javafx.tools.packager.Main.main(Main.java:340)
```

as example of ScalaFX app see https://github.com/scalafx/scalafx-hello-world/blob/master/src/main/scala/hello/ScalaFXHelloWorld.scala

muuki88:
Is this a sbt-native-packager error? For me it looks more like a general sbt/javafx missconfiguration?

cc @metasim

metasim:
@strobe What version of Java are you using?

strobe:
I'm using this
```
java version "1.8.0_45"
Java(TM) SE Runtime Environment (build 1.8.0_45-b14)
Java HotSpot(TM) 64-Bit Server VM (build 25.45-b02, mixed mode)
```

I created simple project for test this issue 

https://github.com/strobe/ScalaFxPackaging

metasim:
@strobe Thanks a lot. Will take a look at it soon.

metasim:
I don't think the bug has anything to do with ScalaFX, but with the plugin improperly handling the case when no maintainer is specified.

@strobe  Working on a fix, but try setting the `maintainer` key and see if you get a package out of it.

@muuki88 Could you please change the title of this issue to: "JDKPackagerPlugin fails to build package if `maintainer` is not set."?

strobe:
@metasim
checked - yes it working well if maintainer was set

metasim:
PR submitted.

muuki88:
fix with #566

PR #566, opened 2015-04-28
Additional tests and fix for #563.

metasim:


muuki88:
Travis as problems again :/

muuki88:
Code LGTM

kardapoltsev:
LGTM too except `println` (: 
You could use `streams.log` instead.

metasim:
@kardapoltsev `println` was just in the scripted test, as it swallows log output by default, and more feedback was needed there. BTW, I was having problems with scripted; calling `> clean` wasn't cleaning, and calling `> set maintainer := "" ` caused an error. Anyone know where to file bugs against `scripted`?

PR #569, opened 2015-05-01
Wip/issue 21 multiple config files shortcuts

ezzarghili:
This fixes issue with multiple shortcuts when having multiple config files,tested with a play application using several config files.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/windows/WixHelper.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/windows/WixHelper.scala
@@ -130,7 +130,8 @@ object WixHelper {
       // Also, we need some mechanism to ensure the start menu folder is removed in the event
       // that we remove all menu items.
       case AddShortCuts(targets, workingDir) =>
-        val id = cleanStringForId("shortcut_" + makeGUID).takeRight(67 - targets.size.toString().size) // Room for "_SC"+incremental number
+        val targetSize = targets.size.toString.size
+        val id = cleanStringForId("shortcut_" + makeGUID).takeRight(67 - targetSize) // Room for "_SC"+incremental number
         val xml =
           <DirectoryRef Id="ApplicationProgramsFolder">
             <Component Id={ id } Guid={ makeGUID }>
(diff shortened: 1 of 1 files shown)


src/main/scala/com/typesafe/sbt/packager/windows/WixHelper.scala
```diff
@@ -130,16 +130,18 @@ object WixHelper {
       // Also, we need some mechanism to ensure the start menu folder is removed in the event
       // that we remove all menu items.
       case AddShortCuts(targets, workingDir) =>
-        val id = cleanStringForId("shortcut_" + makeGUID).takeRight(67) // Room for "_SC"
+        val id = cleanStringForId("shortcut_" + makeGUID).takeRight(67 - targets.size.toString().size) // Room for "_SC"+incremental number
         val xml =
           <DirectoryRef Id="ApplicationProgramsFolder">
             <Component Id={ id } Guid={ makeGUID }>
               {
+                var i = 0
```
muuki88: Codacy Issue found: [Avoid mutable local variables](https://www.codacy.com/app/nepomukseiler/sbt-native-packager/file/600521328/issues/source?bid=2040925&fileBranchId=2282229#l138)

muuki88: if you need an index use 

```
for( (target, i) <- targets.zipWithIndex) yield { ... }
```

src/main/scala/com/typesafe/sbt/packager/windows/WixHelper.scala
```diff
@@ -130,16 +130,18 @@ object WixHelper {
...
                 for (target <- targets) yield {
+                  i = i + 1
                   val name = simpleName(target)
                   val desc = "Edit configuration file: " + name
                   val cleanName = name.replaceAll("[\\.-\\\\//]+", "_")
-                  <Shortcut Id={ id + "_SC" } Name={ cleanName } Description={ desc } Target={ "[INSTALLDIR]\\" + target.replaceAll("\\/", "\\\\") } WorkingDirectory="INSTALLDIR"/>
+                  <Shortcut Id={ id + "_SC" + ("%0" + targets.size.toString().size + "d").format(i) } Name={ cleanName } Description={ desc } Target={ "[INSTALLDIR]\\" + target.replaceAll("\\/", "\\\\") } WorkingDirectory="INSTALLDIR"/>
```
muuki88: With scala `2.10` we can use string interpolation

```scala
s"$i${targets.size.toString().size}d")
```

muuki88:
If I get this correctly, the shortcuts are generated based on the number of targets available.
So I got **3** targets, I will get short cut ids like this

```
val id = makeGuid
val targetSize = targets.size.toString.size
$id_SC_0$targetSize
$id_SC_1$targetSize
$id_SC_2$targetSize
```

If yes, you can put the `targetSize` in a variable, making it clear that it get's reused.

Otherwise this LGTM

ezzarghili:
Thanks for the reviews, actually I am not familiar with scala that was really helpful,
in the other hand yes the shortcuts are generated based on the number of targets available
but, I am using a string format to get a fixed length id string
so for 10 targets I will get 
    $id_SC_01
    $id_SC_02
    ...
    $id_SC_10
and for 100 targets
    $id_SC_001
    ....
    $id_SC_100

PR #579, opened 2015-05-10
Remove `in Docker` from dockerExposedPorts in docs

GitsMcGee:
I found out the hard way that putting `dockerExposedPorts in Docker := Seq(9000, 9443)` into your build.sbt does nothing. The correct setting is `dockerExposedPorts := Seq(9000, 9443)`. I believe the docs, as they are, are misleading.


muuki88:
Thanks :)

Issue #577, opened 2015-05-07
Binary broken when run in path with spaces


francisdb:
Built a zip built with `universal:packageBin` on a sbt project with

```
.enablePlugins(JavaAppPackaging)
.enablePlugins(LauncherJarPlugin)
```

I extracted the resulting zip into `/Users/me/My Downloads/myapp`.
When starting the binary in OS X I get the following error:

```
Error: Invalid or corrupt jarfile /Users/me/My
```

(other ways of packaging and other operating systems might also have this issue

muuki88:
This is interesting. We have tests for this, so this could be a configuration or system issue. Can you share a sample project?

cc @jroper

francisdb:
I think the bug is here when adding `-jar ...`:
https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/jar/LauncherJarPlugin.scala#L45

muuki88:
Is jar name wrong?

francisdb:
no but probably need quotes as `$lib_dir` might contain spaces

muuki88:
Can you provide a pr with a regression test for this?

francisdb:
can you point me to the tests that already test this folder with spaces stuff?

muuki88:
Sure :)

https://github.com/sbt/sbt-native-packager/tree/master/src/sbt-test/jar/launcher-jar

There's a wiki page explaining how to run the tests. Basically it's 
`sbt "scripted jar/*" `

francisdb:
Ok, tried to reproduce this and force the test scripts to use a target/stage folder containing spaces but did not get there. Asserts were still expecting default target/stage folders.

The only solution I found to reproduce the issue is renaming
`src/sbt-test/jar/launcher-jar/`
to
`src/sbt-test/jar/launcher jar/`

but so far no luck on fixing the bash script, quotes don't help as then I get this:

```
error: Unable to access jarfile "/Users/me/My
```

muuki88:
So there needs to be a more sophisticated way of dealing with whitespaces on
windows? For testing you can do the following steps

- go to `test-project-simple` ( in the root folder of sbt-native-packager project)
- add all the settings from `sbt-test/jar/launcher-jar/` you need
- make the build fail in this test project
- change the windows plugin and test with the `test-project-simple` (if you keep the sbt console open you may have to type reload before testing)
[...]

PR #581, opened 2015-05-12
re #577 fix for running the bash file in a path with spaces

francisdb:
this fixes the issue for mac/linux
have not tested or modified anything windows related


muuki88:
LGTM. I'll see if I can run the windows tests

src/main/resources/com/typesafe/sbt/packager/archetypes/bash-template
```diff
@@ -235,9 +235,9 @@ run() {
   fi
   
   if [ -n "$custom_mainclass" ]; then
-    mainclass="$custom_mainclass"
+    mainclass=("$custom_mainclass")
```
muuki88: what do the round brackets achieve here?

francisdb: Needs to be an array (see also declare -a and [@] stuff)

muuki88:
@francisdb I'm checking now if the jar builds are generally broken on windows. On linux we seem to be fine, still I have a strange feeling about this. #586

francisdb:
@muuki88 since I did not touch windows stuff, the LauncherJarPlugin generally still works on non-with-spaces folders. The bat file still needs a fix to be run on paths with spaces (which is not uncommon for windows these days)

But since the linux/macos issue is fixed this can already be merged

muuki88:
okay. will do this after the weekend

PR #586, opened 2015-05-21
Add jar tests to windows build

muuki88:


muuki88:
I'm not able to figure out why these jar-tests break on windows, do you guys have any clues? @nazoking @bjuric

Issue #573, opened 2015-05-05
show dockerCommands fails on windows OS


rsemlal:
Steps to reproduce:
- Create simple sbt project with sbt-native-packager v1.0.0 and `enablePlugins(DockerPlugin)` in its `build.sbt`
- running an sbt REPL on the project, in a windows console
- submitting the sbt command `show dockerCommands`

The command fails with the following stack trace:
```
java.util.regex.PatternSyntaxException: Unexpected internal error near index 1
\
 ^
        at java.util.regex.Pattern.error(Unknown Source)
        at java.util.regex.Pattern.compile(Unknown Source)
        at java.util.regex.Pattern.<init>(Unknown Source)
        at java.util.regex.Pattern.compile(Unknown Source)
        at java.lang.String.split(Unknown Source)
        at java.lang.String.split(Unknown Source)
        at com.typesafe.sbt.packager.docker.DockerPlugin$.com$typesafe$sbt$packager$docker$DockerPlugin$$makeAdd(DockerPlugin.scala:152)
        at com.typesafe.sbt.packager.docker.DockerPlugin$$anonfun$projectSettings$8.apply(DockerPlugin.scala:79)
        at com.typesafe.sbt.packager.docker.DockerPlugin$$anonfun$projectSettings$8.apply(DockerPlugin.scala:70)
        at scala.Function1$$anonfun$compose$1.apply(Function1.scala:47)
        at sbt.$tilde$greater$$anonfun$$u2219$1.apply(TypeFunctions.scala:40)
        at sbt.std.Transform$$anon$4.work(System.scala:63)
        at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:226)
        at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:226)
        at sbt.ErrorHandling$.wideConvert(ErrorHandling.scala:17)
        at sbt.Execute.work(Execute.scala:235)
        at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:226)
        at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:226)
        at sbt.ConcurrentRestrictions$$anon$4$$anonfun$1.apply(ConcurrentRestrictions.scala:159)
        at sbt.CompletionService$$anon$2.call(CompletionService.scala:28)
        at java.util.concurrent.FutureTask.run(Unknown Source)
[...]

PR #588, opened 2015-05-26
Use docker on windows #573

sjeandeaux:
When native-packager add a folder in Dockerfile, it uses "File.separator" but this add doesn't depend to context of build.


muuki88:
thanks for your request. can you squash the commits to one commit?

sjeandeaux:
it's done.

PR #591, opened 2015-06-02
Documented packageTemplateMappings

kodemaniak:
Documented the `packageTemplateMappings` helper that allows to add empty directories to the mappings.


muuki88:
thanks a lot :)

PR #592, opened 2015-06-02
FIX #590 fixed irritating windows documentation

### LINKED ISSUE #590: Windows doc page has many RPM / Linux references
http://www.scala-sbt.org/sbt-native-packager/formats/windows.html

It looks like valid documentation, but then there's "A rpm package needs.." and it gets confusing.

Since I'm in the need of the documentation myself, I'm unable to fix this. If the docs are half-baked it's way better to just say so. Remove the RPM copy-paste remnants.

After 'Settings' it gets back to Windows, again. Strange.

muuki88:


metasim:
:-1:, because "incorrect" != "irritating". cf. Microsoft Word Manual for exemplar irritating windows documentation. ;-)

jsuereth:
:) LGTM

PR #595, opened 2015-06-05
grammar fix for website

SethTisue:
in prominent location on front page


muuki88:
that's.. ehm... thanks :)

PR #597, opened 2015-06-09
tar --force-local

sumkincpp:
MSYS/MinGW windows tar is not able to pack files with absolute paths because of the essence of the colons(i.e. "C:\Code\..."). Colons are treated by tar as links to remote machines. This is not the case of usage for sbt packager.

So, forcing local, from tar man page :
```
--force-local
    archive file is local even if it has a colon 
```

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/universal/Archives.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/universal/Archives.scala
@@ -179,7 +179,7 @@ object Archives {
 
       val tmptar = f / (relname + ".tar")
 
-      Process(Seq("tar", "-pcvf", tmptar.getAbsolutePath) ++ distdirs, Some(rdir)).! match {
+      Process(Seq("tar", "-pcvf", "--force-local", tmptar.getAbsolutePath) ++ distdirs, Some(rdir)).! match {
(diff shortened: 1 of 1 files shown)


muuki88:
@sumkincpp can you make sense of the error message? You can run the failing tests with

```
sbt "scripted universal/publish universal/test-packageName universal/test-zips-no-top-level-dir universal/test-zips"
```

sumkincpp:
@muuki88 uploaded output to http://paste.openstack.org/show/nSOFQdWY09nmeb87aFTt/

muuki88:
yep. Will you be able to fix these?

sumkincpp:
@muuki88 isn't `--force-local` an appropriate fix? Or should I write any tests?

muuki88:
Your fix broke some tests. Point is: "why do the break?". I'm not sure why this is happening and how, but obviously this additional parameter breaks them. At least they all fail for the same reason:

```
[error] tar: Removing leading `/' from member names
[error] tar: /tmp/sbt_abeab32/simple-test-0.1.0.tar: Cannot stat: No such file or directory
[error] tar: Exiting with failure status due to previous errors
```

without having these fixed, I cannot merge this PR.

PR #603, opened 2015-06-14
FIX #347 add ability to add arbitrary stuff from the classpath. Only

### LINKED ISSUE #347: Jar with no classes is not included with staged files?
We've got one library (stanford-corenlp-3.4-models.jar below) which has no classes, but is full of large data files.  It's so large that github won't let us check it in -- not that we really want to do that.

```
libraryDependencies ++= Seq(
  "edu.stanford.nlp" % "stanford-corenlp" % "3.4",
  "edu.stanford.nlp" % "stanford-corenlp" % "3.4" classifier "models"
)
```

When we run play stage or sbt stage, it ignores stanford-corenlp-3.4-models.jar.  It does include all the other code jars in the project, including stanford-corenlp-3.4.jar.   The other build targets: compile, run, test do find the models jar and include it in the classpath from its location in the appropriate local ivy cache,  but stage ignores the jar.  The jar isn't bundled with the other jars during staging or appended to the class path.

Is there something rejecting this jar because it has no referenced classes?   Is there a way to flag the jar and indicate it must be included with the other staged files even if it is never referenced.

We're using the play framework 2.2.1 with sbt 0.13.0.   I think we also tried sbt 0.13.5 and didn't see any difference.

[Transferred from https://github.com/sbt/sbt/issues/1575 at request of jsueret]

muuki88:
documented in scaladoc for now

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/MappingsHelper.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/MappingsHelper.scala
@@ -2,11 +2,22 @@ package com.typesafe.sbt
 package packager
 
 import sbt._
+import com.typesafe.sbt.packager.archetypes.JavaAppPackaging
 
 /** A set of helper methods to simplify the writing of mappings */
 object MappingsHelper {
 
-  /** return a Seq of mappings which effect is to add a whole directory in the generated package */
+  /** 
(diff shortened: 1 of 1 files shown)


jsuereth:
Nice!!!

PR #611, opened 2015-06-24
If the /var/run (aka /run) folder doesn't exist for the PID file, create it

ekuns:
CentOS 7 and many other recent Linuxes have switched to a non-persistent (tmpfs) /run folder, with /var/run being a symbolic link to the tmpfs /run mount.  The RPM creates the correct folder in /var/run for the PID file, but on reboot, that file is gone.  There are many ways to fix it.  This way will work on older systemd versions.

This change was tested on CentOS 7.

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/systemd/start-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/systemd/start-template
@@ -9,6 +9,10 @@ ExecStart=${{chdir}}/bin/${{exec}}
 Restart=always
 RestartSec=${{retryTimeout}}
 User=${{daemon_user}}
+ExecStartPre=/bin/mkdir -p /run/${{app_name}}
+ExecStartPre=/bin/chown ${{daemon_user}}:${{daemon_group}} /run/${{app_name}}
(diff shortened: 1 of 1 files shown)


muuki88:
Nice, LGTM. I'm thinking if we should use the same bash-line as in the Debian systemv script:

```bash
 [ -d "/var/run/${{app_name}}" ] || install -d -o "$DAEMON_USER" -g "$DAEMON_GROUP" -m755 "/var/run/${{app_name}}"
```

I don't care if it's a one liner , but using the same snippet for the same thing. Do you know if there are any differences between both bash snippets?

muuki88:
Ah, and we use this in [Upstart](You can optimise for performance and get scale thrown in (almost for free). Your monolith can scale further than you think) as well.

ekuns:
The one-liner is fine with me.  I like the idea of consistency.  Let me test it locally to see if it works and get back to you.  If someone else could test the same change in Upstart that would be cool.  I don't want to make such a change myself without an environment where I can test it.  After testing, should I check the change into my fork and issue a new pull request?  Can I amend this one?

muuki88:
Sure, just amend and force push.

I will test this locally as well ( just push it on this pr ). Thinks :)

muuki88:
Merging now, and  code-style change in a new PR.

Issue #607, opened 2015-06-22
Play docs are wrong


muuki88:
The `application.ini` doesn't support replacements:

http://www.scala-sbt.org/sbt-native-packager/topics/play.html

PR #613, opened 2015-06-24
Correct the way to give additional informations for Play apps

nremond:
As discussed on the Gitter channel.


muuki88:
Fixes #607

PR #614, opened 2015-06-26
Deployment section, documentation fix

szimano:
As far as i understand this is the proper way of deploying universal packages. If not, then ignore ;-)


muuki88:
That's true! Thanks :)

PR #615, opened 2015-06-29
use autopluging system in deployments docs

szimano:
another docs PR - currently the old way of adding ``deploymentSettings`` is mentioned in the docs, while it's already deprecated

### DIFF
--- a/src/sphinx/topics/deployment.rst
+++ b/src/sphinx/topics/deployment.rst
@@ -27,15 +27,13 @@ For an automatised build process are other plugins like the `sbt release plugin`
 
 Default Configuration
 ---------------------
-The easiest way is to add ``deploymentSettings`` to your ``build.sbt``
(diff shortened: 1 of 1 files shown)


muuki88:
thanks @szimano for keeping the docs up-to-date :)

src/sphinx/topics/deployment.rst
```diff
@@ -27,15 +27,13 @@ For an automatised build process are other plugins like the `sbt release plugin`
 
 Default Configuration
 ---------------------
-The easiest way is to add ``deploymentSettings`` to your ``build.sbt``
+The easiest way is to add ``UniversalDeployPlugin`` to your ``build.sbt``
 
 .. code-block:: scala
 
     import NativePackagerKeys._
```
muuki88: I think this can be removed as well

PR #627, opened 2015-07-17
Added EnvCmd

ipsq:
Added EnvCmd to add `ENV` to your Dockerfile since it differs a bit from default Cmd syntax.


jsuereth:
This LGTM.

PR #628, opened 2015-07-24
Added CombinedCmd and removed EnvCmd

ipsq:
I added `CombinedCmd` for `ONBUILD` instruction and possible future commands that will be added that will have a similar form. For readability I made it so that it uses a `CmdLike` object as an argument. `Cmd` could also be used with an extra argument but it would degrade readability. Removed `EnvCmd` because it introduced an inconsistency in the API with `ADD` and `COPY`. Made `Cmd` accept multiple arguments because this solves the `ADD` and `ENV` inconsistency. This fix won't break the old API because a string with a space in the middle is still acceptable only less readable.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/dockerfile.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/dockerfile.scala
@@ -44,22 +44,38 @@ case class ExecCmd(cmd: String, args: String*) extends CmdLike {
  *
  * @example
  * {{{
- *   val add = Cmd("ADD", "src/resource/LICENSE.txt /opt/docker/LICENSE.txt")
+ *   val add = Cmd("ADD", "src/resource/LICENSE.txt", "/opt/docker/LICENSE.txt")
+ * }}}
+ * 
+ * @example
+ * {{{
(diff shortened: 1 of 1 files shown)


src/main/scala/com/typesafe/sbt/packager/docker/dockerfile.scala
```diff
@@ -44,22 +44,38 @@ case class ExecCmd(cmd: String, args: String*) extends CmdLike {
...
+ * }}}
+ *
+ * @example
+ * {{{
+ *   val env = Cmd("ENV", "APP_SECRET", "7sdfy7s9hfisdufuusud")
  * }}}
  */
-case class Cmd(cmd: String, arg: String) extends CmdLike {
-  def makeContent = "%s %s\n" format (cmd, arg)
+case class Cmd(cmd: String, args: String*) extends CmdLike {
```
muuki88: I'm not 100% sure about this. `Cmd(cmd: String, first: String, rest: String*)` would represent the old interface more correctly. However with your suggestion zero-argument commands would be possible.

@jsuereth  what about binary  compatibility?

muuki88:
I'm merging this at this point as it's about time to release a small RC with some fixes. Would you mind making another pull request if we would break binary compatibility?

thanks a lot for you help :) (and for the nice docs!)

PR #629, opened 2015-07-28
Update start-template for systemd

kononencheg:
sighup for reload

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/systemd/start-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/systemd/start-template
@@ -6,6 +6,7 @@ Requires=${{start_facilities}}
 Type=simple
 WorkingDirectory=${{chdir}}
 ExecStart=${{chdir}}/bin/${{exec}}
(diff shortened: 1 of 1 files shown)


kardapoltsev:
Nice addition! It would be great if you add the same feature to other system loaders (`SystemV` and `Upstart`) (:

muuki88:
@kardapoltsev would this be simple to do? For upstart maybe, but for SystemV I think this is though :p 
For me this looks good to merge. Any further improvements would be awesome, but not required IMO.

kardapoltsev:
> And I'm not 100% sure what this addition does xD

It only sends `SIGHUP (1)` signal to application.
Additions would be great but definitely isn't required. For `SystemV` it could look like
```bash
 reload)
    status_of_proc -p "$PIDFILE" "$RUN_CMD" ${{app_name}} > /dev/null || /bin/kill -HUP "${PIDFILE}"
    ;;
```
I didn't test it.

kardapoltsev:
@kononencheg do you plan to add smth? If not it's fine to merge it as is IMHO.

kononencheg:
@kardapoltsev no, that's all.

kardapoltsev:
Thanks, this is a nice feature for some apps!

PR #631, opened 2015-07-28
FIX #621 --system option for user/groupadd replaced with -r.

### LINKED ISSUE #621: RPM install fails on RHEL 5
I tried installing an RPM from this plugin and got this error

```
Creating system user: mdes in mdes with mdes user-daemon and shell /bin/false
useradd: unknown group mdes
Error in PREIN scriptlet in rpm package mdes-1.0.0_SNAPSHOT-1.noarch
error: %pre(mdes-1.0.0_SNAPSHOT-1.noarch) scriptlet failed, exit status 6
error:   install: %pre scriptlet failed (2), skipping mdes-1.0.0_SNAPSHOT-1
Installed products updated.
```

It looks like the `mdes` group is not being created before attempting to create the user `mdes`.

Actually I'm happy with `root`... but please tell me you don't try to delete the user/group on uninstall!

kardapoltsev:
`--no-create-home` is a default option for system users. I removed it because it's not supported on CentOS.

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/linux/control-functions
+++ b/src/main/resources/com/typesafe/sbt/packager/linux/control-functions
@@ -26,7 +26,7 @@ addUser() {
     if ! getent passwd | grep -q "^$user:";
     then
 	echo "Creating system user: $user in $group with $descr and shell $shell"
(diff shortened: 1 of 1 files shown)


muuki88:
LGTM

PR #641, opened 2015-08-04
Upstart: source /etc/default/${{app_name}}

nafg:
This should fix the limitation mentioned in http://www.scala-sbt.org/sbt-native-packager/archetypes/java_server/customize.html#linux-configuration:

> The server archetype adds an additional way with an etc-default file placed in src/templates, which currently only works for SystemV.

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/upstart/start-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/upstart/start-template
@@ -6,7 +6,7 @@ author "${{author}}"
 # Stanzas
 #
 # Stanzas control when and how a process is started and stopped
-# See a list of stanzas here: http://upstart.ubuntu.com/wiki/Stanzas#respawn
+# See a list of stanzas here: http://upstart.ubuntu.com/wiki/Stanzas
 
 # When to start the service
 start on runlevel ${{start_runlevels}}
@@ -27,7 +27,11 @@ end script
 # set the working directory of the job processes
 chdir ${{chdir}}
 
+setuid ${{daemon_user}}
+setgid ${{daemon_group}}
+
 # Start the process
 script
-  exec sudo -u ${{daemon_user}} bin/${{exec}}
+  . /etc/default/${{app_name}}
+  bin/${{exec}}
 end script


kardapoltsev:
LGTM.

src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/upstart/start-template
```diff
@@ -27,7 +27,11 @@ end script
 # set the working directory of the job processes
 chdir ${{chdir}}
 
+setuid ${{daemon_user}}
```
muuki88: I think we had that at one point in time and it didn't work for some reason. On which systems did you test this?

muuki88:
Nice :) see my one comment. IIRC we had issues with the setuid not working, but I don't remember why.

Can you fix the test? `debian / test-executableScriptName` .  see the wiki page on how to run the tests.

muuki88:
@nafg you had time to look into this, or do you need any help?

nafg:
It was Ubuntu. I haven't had time.

src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/upstart/start-template
allnash: Just curious now if we are executing without sudo, unless I am wrong I think the command should be sudo bin/${{exec}}

I am seeing issues with the daemon_user not having enough permissions

allnash: Ive tried this central is not in the sudoers file. 

Also to make this work I had to do this 
sudo chown ${{daemon_user}}:${{daemon_group}} -R  /usr/share/${{app_name}}l/

muuki88: You don't want to do `sudo chown ${{daemon_user}}:${{daemon_group}} -R /usr/share/${{app_name}}l/` 
`/usr/share/<app>` is generally owned by `root`. If you need write access to disk you add `linuxPackageMappings` with exactly the permissions you need.

timcharper:
FYI, I updated the documentation in my PR saying that upstart specifically doesn't source `/etc/default/...`.

Issue #635, opened 2015-08-01
Overridable bash template


damirv:
I'm using the Play framework and I want to override the bash template. Is there currently a way to override the bash template without placing it under `src/templates/bash-template`? 

It seems from [this](https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaApp.scala#L105) that it uses `sourceDirectory` to determine the base. I tried setting `sourceDirectory in makeBashScript := baseDirectory.value / "app"` and that doesn't work (as I expected). Play defines the following: `sourceDirectory in Compile := baseDirectory.value / "app",`. Setting `sourceDirectory := baseDirectory.value / "app"` works but then I get also other folders in the package.

muuki88:
I'm afraid this is not possible without overriding the task that generates the bash script.
We could introduce a special setting, but I think this more cosmetic than actual functionality.

damirv:
With overriding the task, you mean something like [this](https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaApp.scala#L262)?

damirv:
I have this now (added to build.sbt) and it works (bash-template is located in the `conf` folder of Play). But if we would have `defaultTemplateLocation` as a setting, then it would be easier to override (because that's basically what I'm overriding in this function). Do you think this is useful?

```scala
{
  object Override extends JavaAppStartScript {
    val bashTemplate = "bash-template"

    def settings: Seq[Setting[_]] = Seq(
      makeBashScript <<= (baseDirectory, bashScriptDefines, target in Universal, executableScriptName, sourceDirectory) map makeCustomUniversalBinScript
    )

    import JavaAppPackaging.autoImport._

    def makeCustomUniversalBinScript(baseDir: File, defines: Seq[String], tmpDir: File, name: String, sourceDir: File): Option[File] =
      if (defines.isEmpty) None
      else {
        val defaultTemplateLocation = baseDir / "conf" / bashTemplate
        val defaultTemplateSource = getClass getResource bashTemplate

        val template = if (defaultTemplateLocation.exists)
          defaultTemplateLocation.toURI.toURL
        else defaultTemplateSource

        val scriptBits = JavaAppBashScript.generateScript(defines, template)
        val script = tmpDir / "tmp" / "bin" / name
        IO.write(script, scriptBits)
        // TODO - Better control over this!
[...]

PR #646, opened 2015-08-08
overridable bash and bat templates

damirv:
This implements feature request #635. 

I left the old method signatures in the [`JavaAppStartScript`](https://github.com/sbt/sbt-native-packager/blob/6cd1c3a04591dba721ea7e022c5d98c1230bbc7d/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaApp.scala#L282) trait, for backwards compatibility. Is this something that you could use? Should I add an example test project demonstrating this feature?


muuki88:
Thanks for resolving this feature request. Code looks solid. Could you add test and some docs on how to use?

damirv:
I've added tests and some documentation, let me know if I should add anything else.

muuki88:
Nice! One last thing, could you squash your commits into one?

damirv:
Done :)

muuki88:
Awesome :+1:

PR #652, opened 2015-08-14
ash support documentation

gzoller:


muuki88:
Thanks, looks good. The documentation states that the is a need to override the `bashScript`. I didn't get this at the other pull request. I think we should provide this by default, so it works out-of-the-box.

```scala
bashScriptTemplateLocation := {
   // generate the file, so it can be found
   val script = (target in Universal) / "tmp" / "ash-start-script"
   val content = .. // getResource from internal 
   // fill in the contents from the interally shipped jar
   IO.write(script, content)
   // return this as template location
   script
}
```

Simplifies usage and documentation :)

gzoller:
Makes sense.  Let me look at it.  That will probably require an ash-template file but that shouldn't be a big deal. I'll put it in this PR.

gzoller:
I've created an overridable ash-template file so it works more "out of the box".  That'll be better.  I am getting a Travis failure but the CI system is a little flakey today and I can't click into the details to know what's failing.

kardapoltsev:
I'm also couldn't get access to details.. :(
'There is not an active repository' Travis said.
@muuki88 is there any spercial security configuration at travis side?

muuki88:
@kardapoltsev yeah, this is for a few days now.

@jsuereth I fear this is something the repo-admin has to reconfigure?

src/main/scala/com/typesafe/sbt/packager/archetypes/AshScriptPlugin.scala
```diff
@@ -73,15 +73,14 @@ import SbtNativePackager.{ Universal, Debian }
...
   override def requires = JavaAppPackaging
 
-  //object autoImport extends JavaAppKeys
-
   import JavaAppPackaging.autoImport._
 
+  val ashTemplate = "ash-template"
+
   override def projectSettings = Seq(
+    bashScriptTemplateLocation := (sourceDirectory.value / "templates" / ashTemplate),
```
muuki88: I fear this won't be enough. You have to create the _ash template_. Something like

bashScriptTemplateLocation := {
   // generate the file, so it can be found
   val script = (target in Universal) / "tmp" / "ash-start-script"
   val content = .. // getResource from internal 
   // fill in the contents from the interally shipped jar
   IO.write(script, content)
   // return this as template location
   script
}

gzoller: I'm afraid I'm confused on this point.  I've got something like this in makeUniversalAshScript() in file  AshScriptPlugin.scala.  I've built it locally and it seems to work great for a project that consumes this feature.  Looks like this:

```scala
      val template = resolveTemplate(defaultTemplateLocation)
      val scriptBits = JavaAppAshScript.generateScript(defines, template)
      val script = tmpDir / "tmp" / "bin" / name   // fine since this is an ash-specific archetype
      IO.write(script, scriptBits)
      script.setExecutable(true)
      Some(script)
```

muuki88:
Travis CI flakiness :( Tests run now. I've tested this on my machine as well and everything works smoothly. Great job :)

muuki88:
@gzoller I forgot to ask. You mentioned that you use busybox within the docker build, right? Is there a busybox-java image that we could use as a default for the `AshScriptPlugin`?

gzoller:
Kinda...  It's not provided by busybox but a developer who goes by errordeveloper I think.

fiadliel:
Here's one image you might be able to try: https://hub.docker.com/r/develar/java/
The image is based on Alpine Linux (w/busybox), with the Oracle 8 JRE.

Alpine Linux seems to have quite a mini-community (I only came across it this month).
However, it's hard to know yet which will be well-supported in the long term, I haven't seen any image which I would be certain will be kept up to date (because they are single-person endeavours).

gzoller:
Exactly.  That's kinda why I didn't wire in a default at this point.  Wasn't really sure who'd stick until some time had passed.

muuki88:
Okay, that makes sense. Thanks @fiadliel for the research.

Issue #619, opened 2015-07-07
hardcoded RPM log directory


fommil:
In

https://github.com/sbt/sbt-native-packager/blob/master/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/systemv/start-rpm-template#L43

```
RUN_CMD="$exec >> /var/log/${{app_name}}/daemon.log 2>&1 &"
```

but the packager may have provided a custom log directory, which should be preferred.

fommil:
uugh, I was going to send you a quick PR on this but it's more involved than I thought. The logs directory needs to be available in the default replacements list and changing that looks like a big faff.

Workaround is to use a custom script, of course.

muuki88:
Currently I'm working on a major refactoring of the `maintainerScripts`. After that, it should be a lot easier.

giampaolotrapasso:
Any forecast about the resolution of this?

muuki88:
Unfortunately not. #625 is almost done, but needs some clean up. And then lets see if there is a good way fixing this.

louisq:
I was not aware that a issue was opened for this, but the #675 pull request addresses the custom log directory

muuki88:
ah, thanks :)

PR #656, opened 2015-08-31
Make the log file generated by application daemon configurable in RPM based SystemV

fsat:
Introduce a new settings key called `rpmDaemonLogFile` which will be used to configure
the name of the log file generated by application daemon in RPM based SystemV.

Make the default value of the `rpmDaemonLogFile` settings to be `${{app_name}}.log`.


fsat:
Fixes https://github.com/typesafehub/conductr/issues/793

muuki88:
#619 is related to this. If you could make this modifiable, then it could be configured for conductr and other things.

fsat:
@muuki88 - I have made this configurable with `daemon.log` as the default value to match existing behaviour. Updated the PR title & description to match as well.

src/main/scala/com/typesafe/sbt/packager/linux/Keys.scala
```diff
@@ -13,6 +13,7 @@ trait LinuxKeys {
   val daemonGroup = SettingKey[String]("daemon-group", "Group to start application daemon")
   val daemonGroupGid = SettingKey[Option[String]]("daemon-group-gid", "GID of daemonGroup")
   val daemonShell = SettingKey[String]("daemon-shell", "Shell provided for the daemon user")
+  val daemonLogFile = SettingKey[String]("daemon-log-file", "Name of the log file generated by the application daemon")
```
huntc: Should be an rpm only key

muuki88: for what reason? Not sure if other packages/system start daemons may use this as well. debian systemv for example.

muuki88:
@huntc if you have no objections, feel free to merge.

src/main/scala/com/typesafe/sbt/packager/linux/Keys.scala
huntc: Felix and I spoke about this and concluded that it was only needed for rpm. You can't have this choice for Debian as the log is already named with the dpkg-ing and there is no ability to change it.

huntc:
Thanks @muuki88. I believe Felix has some more changes to push so no merge is required immediately.

fsat:
@huntc & @muuki88 - I have created a rpm specific settings called `rpmDaemonLogFile` which has `daemon.log` as the default value.

This settings will allow overriding the log file name generated by the application daemon in rpm based install.

I have performed manual test for this change using the `test-project`. I have tested both default `rpmDaemonLogFile` value and overriding `rpmDaemonLogFile` scenario successfully.

src/main/scala/com/typesafe/sbt/packager/rpm/Keys.scala
```diff
@@ -60,4 +60,15 @@ trait RpmKeys {
...
   val rpmLint = TaskKey[Unit]("rpm-lint", "Runs rpmlint program against the genreated RPM, if available.")
 
+  val rpmDaemonLogFile = SettingKey[String]("rpm-daemon-log-file", "Name of the log file generated by application daemon")
+
+  val rpmScriptReplacements = SettingKey[Seq[(String, String)]](
+    "rpmScriptReplacements",
+    """|Replacements of template parameters used in linux rpm scripts.
+       |  Default supported templates:
+       |  rpmDaemonLogFile - name of the log file generated by application daemon
+    """.stripMargin
```
huntc: `rpmScriptReplacements` should be `rpm-script-replacements`

src/main/scala/com/typesafe/sbt/packager/rpm/RpmPlugin.scala
```diff
@@ -80,7 +80,11 @@ object RpmPlugin extends AutoPlugin {
     target in Rpm <<= target(_ / "rpm"),
     name in Rpm <<= name in Linux,
     packageName in Rpm <<= packageName in Linux,
-    executableScriptName in Rpm <<= executableScriptName in Linux
+    executableScriptName in Rpm <<= executableScriptName in Linux,
+    rpmDaemonLogFile := "daemon.log",
```
huntc: For consistency I recommend using the name as per Debian packaging, and not `daemon.log`, unless there is some Fedora style precedence for this. This change of ours was motivated by the fact that we came upon `daemon.log` unexpectedly.

@muuki88 WDYT?

fsat:
Marking this as 'WIP - DON'T MERGE' for now since there's still a change pending

src/main/scala/com/typesafe/sbt/packager/rpm/RpmPlugin.scala
```diff
@@ -80,7 +80,11 @@ object RpmPlugin extends AutoPlugin {
     target in Rpm <<= target(_ / "rpm"),
     name in Rpm <<= name in Linux,
     packageName in Rpm <<= packageName in Linux,
-    executableScriptName in Rpm <<= executableScriptName in Linux
+    executableScriptName in Rpm <<= executableScriptName in Linux,
+    rpmDaemonLogFile := s"${(packageName in Linux).value}.log",
```
fsat: Quoting @huntc + cc @muuki88 as commit has been updated: 

> For consistency I recommend using the name as per Debian packaging, and not daemon.log, unless there is some Fedora style precedence for this. This change of ours was motivated by the fact that we came upon daemon.log unexpectedly.
>
> @muuki88 WDYT?

src/main/scala/com/typesafe/sbt/packager/rpm/Keys.scala
```diff
@@ -60,4 +60,15 @@ trait RpmKeys {
   // Building
   val rpmLint = TaskKey[Unit]("rpm-lint", "Runs rpmlint program against the genreated RPM, if available.")
 
+  val rpmDaemonLogFile = SettingKey[String]("rpm-daemon-log-file", "Name of the log file generated by application daemon")
+
+  val rpmScriptReplacements = SettingKey[Seq[(String, String)]](
```
muuki88: Why introduce this new setting? `linuxScriptReplacements in Rpm` should be sufficient?

src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
```diff
@@ -128,6 +128,7 @@ object JavaServerAppPackaging extends AutoPlugin {
         makeStartScriptReplacements,
       linuxScriptReplacements += JavaServerLoaderScript.loaderFunctionsReplacement(serverLoading.value, ARCHETYPE),
       linuxScriptReplacements ++= bashScriptEnvConfigLocation.value.map(ENV_CONFIG_REPLACEMENT -> _).toSeq,
+      linuxScriptReplacements ++= rpmScriptReplacements.value,
```
muuki88: just add you `rpmDaemonLog` property here.

src/main/scala/com/typesafe/sbt/packager/rpm/Keys.scala
huntc: Agreed. Sorry for missing this.

huntc: I understand that `rpmScriptReplacements` is declared here because it is specific to RPM i.e. it isn't a script replacement destined for Linux.

If there was a key named just `scriptReplacements` then that'd be different...

Make sense @muuki88 ?

src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
huntc: I believe that the idea is that there could be many script replacements for RPM - just as there are for Linux. Consistency.

src/main/scala/com/typesafe/sbt/packager/rpm/Keys.scala
muuki88: yes, that makes sense @huntc and I like the idea for a general purpose setting `scriptReplacements` and scoping it to the different configurations. This will be easier if there were less settings ;) 

IMHO we should not introduce a new key as removing these is much harder.

src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
muuki88: I feel `linuxScriptReplacements in Rpm` is the way most things work in sbt-native-packager.

huntc: OK. @fsat Could you please use `linuxScriptReplacements in Rpm` once again.

fsat:
@muuki88 & @huntc - changes implemented, please review.

huntc:
LGTM

Issue #659, opened 2015-09-01
RPM relocation - requires local ownership of sbt-native-packager source templates


edwardcallahan:
When using rpmPrefix to make a package relocatable, one must also take ownership of the init start script and scriptlets. 

The current doc for rpmPrefix has had rpmChangelogFile inserted into the 'how to enable relocation' story. From https://github.com/sbt/sbt-native-packager/pull/242/files we learn that one must also copy the init script out of sbt-native-packager into the project source in order to get a functioning init script when relocated (the stock script uses the default path in exec).  This is presented under Template Changes with: "Apply the following changes to the default init start script.  You can find this in the sbt-native-packager source."  This is not only a template customization example, but is also required in order to have a relocatable package that will start.

Copying sbt-native-packager sources into a project source is highly undesirable. Could the templates use PACKAGE_PREFIX instead of chdir, for example?

muuki88:
I haven't used the prefix myself yet. So to get it working you have to

- set the `rpmPrefix` value
- override `scriptlets`
- copy the (systemV?) `initScript`

The `${PACKAGE_PREFIX}` has to be inserted "by hand", right?
First thing that comes to mind is, changing the `linuxScriptReplacements`. `{{chdir}}` could be set to the relocated location? Or what else does this relocation do?

edwardcallahan:
I got as far as:
- set defaultLinuxInstallLocation
- set rpmPrefix
- set linuxPackageSymlinks (tip: use `show linuxPackageSymlinks` to get them all!)
- set defaultLinuxLogsLocation
This resulted in a relocatable package, but with an invalid init script. 

As I understand the doc, yes, next I would need to copy (start-rpm-template)[https://github.com/sbt/sbt-native-packager/blob/master/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/systemv/start-rpm-template] into the project tree and edit it to use `${PACKAGE_PREFIX}`  instead of `${{chdir}}` for `exec`. With same for the scriptlet changes for post-rpm and reun-rpm.

Yes, setting `{{chdir}}` to the user specified location should also work. Relocation let's the user specify an alternative install location (i.e. `rpm -i --relocate /usr/share=/opt mypackage.rpm`). The relocation works, it's the reflection of the user chosen path in the (systemV) scripts that's been the sticky wicket.

fsat:
@muuki88 - maybe we can close this after 1.0.5 release since it's fixed by #661

[...]

PR #661, opened 2015-09-04
Support for relocatable RPMs

fsat:
Save the value of `RPM_INSTALL_PREFIX` into application's sysconfig file.

Upon application start up, look for this value, else use the `chdir` value supplied via sbt templates.

--

I have two options in detecting where the RPM package is installed.

**Option 1: Use `RPM_INSTALL_PREFIX` provided during installation process**

The `RPM_INSTALL_PREFIX` environment variable is only provided during the install time. This value would need to be persisted so it can be retrieved during application start-up.

**Option 2: Use rpm command**

The command `rpm -q --queryformat "%{INSTPREFIXES}" ${{app_name}}` will print out the location where the package is installed. This command works for both Redhat 6 and 7.

I have decided to go with Option 1 because we are using the values provided to us by the operating system instead of trying to deduce this value ourselves.

The downside of this approach is:
* I am persisting the value of `RPM_INSTALL_PREFIX` in the `/etc/sysconfig/${{app_name}}`.
* I need to replace this information at re-install time or upgrade time.

@muuki88 and @edwardcallahan - can I get your thoughts on my approach please? Happy to change if you think Option 2 is a better way to move forward.

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/rpm/preinst-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/rpm/preinst-template
@@ -9,3 +9,6 @@ then
     addGroup ${{daemon_group}} "${{daemon_group_gid}}"
     addUser ${{daemon_user}} "${{daemon_user_uid}}" ${{daemon_group}} "${{app_name}} user-daemon" "${{daemon_shell}}"
 fi
+
+[ -e /etc/sysconfig/${{app_name}} ] && sed -i 's/PACKAGE_PREFIX\=.*//g' /etc/sysconfig/${{app_name}}
+[ -n "$RPM_INSTALL_PREFIX" ] && echo "PACKAGE_PREFIX=${RPM_INSTALL_PREFIX}" >> /etc/sysconfig/${{app_name}}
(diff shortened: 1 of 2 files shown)


fsat:
Fixes https://github.com/typesafehub/conductr/issues/790

fsat:
Fixes #659

fsat:
I have done a manual test with the following steps:
* Override the `preinst` and `start` with the original content from `sbt-native-packager` + changes I have introduced applied.
* Generate rpm package
* Install rpm package into default path, e.g. `sudo rpm -i conductr-0.1.0-1.noarch.rpm`
* Ensure process are started correctly with package files created into the correct locations.
* Uninstall the package.
* Re-install rpm package into relocated path, e.g. `sudo rpm -i conductr-0.1.0-1.noarch.rpm --relocate /usr/share=/opt`
* Ensure process are started correctly with package files created into the correct locations.

The manual test has been done on Redhat 7.

huntc:
Using `-z` for tests is better.

Otherwise LGTM - @muuki88 - WDYT?

fsat:
Used `-n` instead since I'd like to check for non-empty string, also shorten the `if` clause where possible.

fsat:
Manual test also done successfully on Redhat 6.

muuki88:
LGTM.

PR #665, opened 2015-09-07
under upstart, use start-stop-daemon -c instead of sudo -u

urzhumskov:
Using ```start-stop-daemon``` instead of ```exec sudo -u``` allows, for example, setting environment variables in ```/etc/init/$service.override```. Using ```sudo``` resets the environment, so these overrides are lost.

Probably I miss some reasons why ```sudo``` was chosen, though.

An even more upstart-correct way of doing things would be just using ```exec``` with correct ```setuid/setgid``` stanzas, but this will prevent us from being able to run pre-start/post-stop sections as root, which might be useful for some.

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/upstart/start-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/upstart/start-template
@@ -24,10 +24,7 @@ pre-start script
 	[ -d /var/run/${{app_name}} ] || install -m 755 -o ${{daemon_user}} -g ${{daemon_group}} -d /var/run/${{app_name}}
 end script
 
-# set the working directory of the job processes
-chdir ${{chdir}}
-
 # Start the process
 script
-  exec sudo -u ${{daemon_user}} bin/${{exec}}
+  start-stop-daemon --start -c ${{daemon_user}} --exec {{chdir}}/bin/${{exec}}
 end script


muuki88:
@urzhumskov thanks for your contribution. On which systems did you test this? AFAIK we were using `sudo -u` because upstart could be used for `debian` based system as well as for `Fedora`, `RHEL` and `CentOS`, which don't have a `start-stop-daemon` facility.

I think we can only add this change if we split the upstart script, like the `systemV` script in `start-template-debian` and `start-template-rpm`.

urzhumskov:
@muuki88 I tested it on (pre-systemd) Ubuntu under assumption that wherever there is upstart, there is start-stop-daemon :) Now I see the reasoning behind using sudo, so disregard this; not a big issue anyway.

muuki88:
@urzhumskov thanks anyway for pull request. Keeping track with all best practices needs help :)

guersam:
How about `sudo su - ${{daemon_user}} -c bin/${{exec}}`? It exists in most platforms and would preserve the environment.

muuki88:
@guersam could you provide a PR for this? I won't have time to test this.

cc @kardapoltsev

theossystems:
How about `sudo -E` ?
>     -E, --preserve-env
                 Indicates to the security policy that the user wishes to preserve their existing environment variables.  The
                 security policy may return an error if the user does not have permission to preserve the environment.

PR #666, opened 2015-09-07
FIX #662 and make universal archive options customizable

### LINKED ISSUE #662: tar is expected to have --force-local option

muuki88:
cc @mackler

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/universal/Archives.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/universal/Archives.scala
@@ -14,9 +14,10 @@ object Archives {
    * @param name of output (without extension)
    * @param mappings included in the output
    * @param top level directory
+   * @param options  NOT USED
    * @return zip file
    */
-  def makeZip(target: File, name: String, mappings: Seq[(File, String)], top: Option[String]): File = {
+  def makeZip(target: File, name: String, mappings: Seq[(File, String)], top: Option[String], options: Seq[String]): File = {
     val zip = target / (name + ".zip")
 
     // add top level directory if defined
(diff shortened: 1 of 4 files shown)


src/main/scala/com/typesafe/sbt/packager/universal/Archives.scala
```diff
@@ -58,9 +60,10 @@ object Archives {
    *  @param name of output (without extension)
    *  @param mappings included in the output
    *  @param top level directory : NOT USED
+   *  @param options  NOT USED
    *  @return dmg file
    */
-  def makeDmg(target: File, name: String, mappings: Seq[(File, String)], top: Option[String]): File = {
+  def makeDmg(target: File, name: String, mappings: Seq[(File, String)], top: Option[String], options: Seq[String]): File = {
```
muuki88: Codacy Issue found: [Method is longer than 50 lines](https://www.codacy.com/app/nepomukseiler/sbt-native-packager/file/946190549/issues/source?bid=2040925&fileBranchId=2849134#l66)

mackler:
Thanks for doing this.  I can see I would have been unable to figure this out myself.

muuki88:
@sumkincpp we made the `--force-local` customizable. With the next release you will can add this option for your build.

PR #667, opened 2015-09-10
WIP - DON'T MERGE - Reinstate and deprecate older methods to preserve binary backward compatibility with 1.0.4

fsat:
The methods in question are the methods used to create various kinds of archives (e.g. zip or tar) from `com.typesafe.sbt.packager.universal.Archives`

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/universal/Archives.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/universal/Archives.scala
@@ -192,10 +192,24 @@ object Archives {
     file(f.getAbsolutePath + ".xz")
   }
 
-  val makeTxz = makeTarball(xz, ".txz") _
-  val makeTgz = makeTarball(gzip, ".tgz") _
+  val makeTxz = makeTarballWithOptions(xz, ".txz") _
+  val makeTgz = makeTarballWithOptions(gzip, ".tgz") _
 
   /**
+   * Helper method used to construct tar-related compression functions with `--force-local` and `-pvcf` option specified
+   * as default.
+   * @param target folder to build package in
+   * @param name of output (without extension)
+   * @param mappings included in the output
+   * @param top level directory
+   * @return tar file
+   *
+   */
+  def makeTarball(compressor: File => File, ext: String)(target: File, name: String, mappings: Seq[(File, String)], top: Option[String]): File =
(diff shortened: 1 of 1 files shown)


fsat:
@muuki88 - I'm putting back older methods in the `com.typesafe.sbt.packager.universal.Archives` and marking them as deprecated to preserve backward binary compatibility with 1.0.4.

There's one problem however, I can't put back older version of the `makeTarball` method.

I can’t overload `makeTarball` with older version of the method due to the fact `options` argument introduced in the second group of arguments.

Can I suggest we move the existing `makeTarball` to a new method called `makeTarballWithOptions`, and reinstate older `makeTarball` method? This way, the backward compatibility will be maintained, and we can have the fix for #662.

I'm happy to discuss further if you'd like.

//cc @huntc

src/main/scala/com/typesafe/sbt/packager/universal/Archives.scala
```diff
@@ -9,6 +9,21 @@ object Archives {
...
   /**
    * Makes a zip file in the given target directory using the given name.
+   * @param target folder to build package in
+   * @param name of output (without extension)
+   * @param mappings included in the output
+   * @param top level directory
+   * @return zip file
+   */
+  @deprecated(
+    "Use [[com.typesafe.sbt.packager.universal.Archives.makeZip(File, String, Seq[(File, String)], Option[String], Seq[String]): File]]",
```
fsat: @huntc - is there a way to reference overloaded method in Scaladoc? This doesn't seem to come out right.

huntc: dunno

src/main/scala/com/typesafe/sbt/packager/universal/Archives.scala
```diff
@@ -9,6 +9,21 @@ object Archives {
...
+    since = "1.0.5"
+  )
+  def makeZip(target: File, name: String, mappings: Seq[(File, String)], top: Option[String]): File =
+    makeZip(target, name, mappings, top, options = Seq.empty[String])
```
huntc: The type can be inferred for `empty`

muuki88:
`1.0.5-M2` is out

PR #675, opened 2015-10-02
defaultLinuxLogsLocation not applied to start-rpm-template

louisq:
Where an application is installed using an RPM and the defaultLinuxLogsLocation is set the rpm would correctly create the logging folders for the application at the new location. Unfortunately the daemon would still try to log in the default location which was hard coded which resulted in the daemon failing. 

    [louisq@localhost APPNAME]$ sudo rpm -Uvh target/rpm/RPMS/noarch/APPNAME-45-0.noarch.rpm 
    [sudo] password for louisq: 
    Preparing...                ########################################### [100%]
    Creating system group: APPNAME
    Creating system user: APPNAME in APPNAME with APPNAME user-daemon and shell /bin/false
       1:APPNAME                   ########################################### [100%]
    Adding APPNAME to autostart using chkconfig
    Starting APPNAME: [  OK  ]
    bash: /var/log/APPNAME/daemon.log: No such file or directory

This change removes the hard coding from the logging in the start-rpm-template to instead use the value which originates from the LinuxPlugin. If the value is overridden the daemon will now log to the specified location.

An RPM built with those changes was run on a centos 6.7 machine and the universal/* and rpm/* tests were all run successfully


muuki88:
LGTM.  Can you add a regression test for this? Here is a [small developer guide](https://github.com/sbt/sbt-native-packager/wiki/Developer-Guide) on how to run the tests. if you need any help, just ping here.

louisq:
Thank you. Will do that tomorrow when I come back from an obligation that I have

louisq:
Thank you for the link to the guide.
Added additional test that validates that the defaultLinuxLogLocation overrides the value in the init.d file generated with the rpm. Also added coverage for the defaultLinuxInstallLocation which can also be overridden in this file. Is there something else that you would like to see?

muuki88:
Awesome. Can you squash the commits into one? Then this is ready to merge :)

louisq:
Completed the rebase of the branch

PR #677, opened 2015-10-06
Added rpm-build as a Requirement for *.rpm

schmitch:
Currently rpm-build is a requirement of the rpm format aswell, it will fail hardly if you don't have it installed.


muuki88:
Thanks ;)

PR #680, opened 2015-10-23
Ensure pidfile dir on systemv rpm template exists

mcarolan:
On CentOS 7 /run is tmpfs, and /var/run is a symlink to /run.

Upon reboot the /var/run/${{app_name}} directory (created by the RPM)
will be blown away, causing the init script to fail - but still spawn
the service as a side effect.

This change will create the /var/run/${{app_name}} directory if it does
not already exist, and change the owner to $DAEMON_USER.


muuki88:
LGTM. Will merge after travis has finished

PR #685, opened 2015-11-02
Relocatable rpm symlink fix

fsat:
When creating relocatable RPM packages, the symlink needs to take into account the relocated file paths.

RPM does not provide out of the box support for this, hence the pull request to manage symlinks as part of `%post` and `%postun` scriptlet. This will ensure the symlink will be created and removed as part of post-install and post-uninstall step.

When creating the symlink, both the source path and the symlink itself need to take into account the relocated path.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
@@ -295,5 +295,4 @@ object JavaServerAppPackaging extends AutoPlugin {
       case script => TemplateWriter generateScriptFromString (content + script, replacements)
     }
   }
-
 }
--- a/src/main/scala/com/typesafe/sbt/packager/linux/LinuxPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/linux/LinuxPlugin.scala
@@ -79,7 +79,7 @@ object LinuxPlugin extends AutoPlugin {
       author = (maintainer in Linux).value,
       description = (packageSummary in Linux).value,
       execScript = (executableScriptName in Linux).value,
-      chdir = s"${defaultLinuxInstallLocation.value}/${(packageName in Linux).value}",
+      chdir = chdir(defaultLinuxInstallLocation.value, (packageName in Linux).value),
       logdir = defaultLinuxLogsLocation.value,
       appName = (packageName in Linux).value,
       version = sbt.Keys.version.value,
@@ -248,5 +248,8 @@ object LinuxPlugin extends AutoPlugin {
       packageMappingWithRename(remaining: _*) withUser user withGroup group withPerms "0644"
     )
   }
+
+  final def chdir(installLocation: String, packageName: String): String =
+    s"$installLocation/$packageName"
 
 }
--- a/src/main/scala/com/typesafe/sbt/packager/rpm/RpmMetadata.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/rpm/RpmMetadata.scala
@@ -2,9 +2,8 @@ package com.typesafe.sbt
 package packager
 package rpm
 
-import linux.{ LinuxPackageMapping, LinuxFileMetaData }
+import com.typesafe.sbt.packager.linux.{ LinuxPlugin, LinuxPackageMapping, LinuxFileMetaData, LinuxSymlink }
 import sbt._
-import com.typesafe.sbt.packager.linux.LinuxSymlink
 import java.io.File
 
 case class RpmMetadata(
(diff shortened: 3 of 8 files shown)


fsat:
@huntc and @muuki88 - can I get a review please?

src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
```diff
@@ -284,6 +284,8 @@ object JavaServerAppPackaging extends AutoPlugin {
     Some(script)
   }
 
+  protected def makePostInstallScript(): Option[String] = ???
+
```
huntc: Why was this removed?

src/main/scala/com/typesafe/sbt/packager/linux/LinuxPlugin.scala
```diff
@@ -249,4 +249,7 @@ object LinuxPlugin extends AutoPlugin {
     )
   }
 
+  final def chdir(installLocation: String, packageName: String): String =
+    s"$installLocation/$packageName"
+
```
huntc: I see little value in the addition of this function. It is another level of indirection and doesn't actually change the dir. :-)

huntc: Cancel that - I see that you're calling it from elsewhere

huntc:
LGTM

src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
fsat: Oops... I thought I've put that in, I'll put it back.

src/main/scala/com/typesafe/sbt/packager/linux/LinuxPlugin.scala
fsat: Yes, the `chdir` is actually not change directory - it's the `chdir` variable in the RPM scriptlet which is the actual directory where the application is installed.

fsat:
Manual test successfully done on Redhat 7 VM for both relocated and default installation.

Test is done using ConductR rpms:
* During installation: ensure correct installation path, correct symlink, and application is started successfully.
* During uninstall: ensure application is stopped successfully, files and symlink cleaned up successfully.

src/main/scala/com/typesafe/sbt/packager/rpm/RpmMetadata.scala
```diff
@@ -59,6 +58,47 @@ case class RpmScripts(
...
+    posttrans.fold("")("\n%posttrans\n" + _ + "\n\n")
+
+  def verifyscriptContent(): String =
+    verifyscript.fold("")("\n%verifyscript\n" + _ + "\n\n")
+
+  def preunContent(): String =
+    preun.fold("")("\n%preun\n" + _ + "\n\n")
+
+  def postunContent(tearDownSymlinkScript: Option[String]): String = {
+    val scripts = Seq(postun, tearDownSymlinkScript).collect {
```
muuki88: this can be a simple `flatten`, or?

src/main/scala/com/typesafe/sbt/packager/rpm/RpmMetadata.scala
```diff
@@ -59,6 +58,47 @@ case class RpmScripts(
...
   postun: Option[String] = None) {
+
+  def pretransContent(): String =
+    pretrans.fold("")("\n%pretrans\n" + _ + "\n\n")
+
+  def preContent(): String =
+    pre.fold("")("\n%pre\n" + _ + "\n\n")
+
+  def postContent(buildSymlinkScript: Option[String]): String = {
+    val scripts = Seq(post, buildSymlinkScript).collect {
```
muuki88: `flatten` ?

src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
muuki88: Not implemented? what's missing here?

muuki88:
I'm still confused why the `relocate` command in rpm is necessary as this adds so much complexity. The `defaultLinuxInstallationPath` can't fix this?

Nevertheless code looks okay. And if the relocate feature isn't used, nothing of these bash-magic will make it into the `scriptlets`.

src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
```diff
@@ -284,6 +284,8 @@ object JavaServerAppPackaging extends AutoPlugin {
     Some(script)
   }
 
+  protected def makePostInstallScript(): Option[String] = ???
```
muuki88: Codacy Issue found: [Usage of ??? operator](https://www.codacy.com/app/nepomukseiler/sbt-native-packager/file/1241567461/issues/source?bid=2040925&fileBranchId=2935951#l287)

fsat:
@muuki88, unfortunately the `defaultLinuxInstallationPath` does not fix this. And unfortunately the bash magic will need to be present because the relocation is specified during package install time.

When rpm install is called with relocate option (i.e. `rpm -i <path to my package file> --relocate /usr/share=/opt`), the `defaultLinuxInstallationPath` which is `/usr/share` is now mapped to the new location `/opt`. Internally RPM will look at the `%files` list and move all the files listed under `/usr/share` to `/opt`.

Before this change, the symlink is created during the build time and registered as part of `%files` list. Because of this, specifying the `--relocate` during rpm install will break the symlink.

For example, a symlink from `/usr/bin/myapp` to `/usr/share/myapp/bin/myapp` are created when the package is built. If I were to relocate the `/usr/share` to `/opt`, the `/usr/bin/myapp` symlink will still point to `/usr/share/myapp/bin/myapp` which is no longer valid because all the files are now under `/opt`.

This change will manage the symlink creation as part of post-install and post-uninstall steps using `%post` and `%postun` respectively. The symlink now will point to the correct relocated path. If the package is not relocated (i.e. using the `defaultLinuxInstallationPath`), the `relocateLink` function will leave the link as-is.

The additional complexity introduced in the `%post` and `%postun` scriptlet is unavoidable because rpm does not provide means to handle symlink relocation out of the box.

Hopefully it explains why this change is done.

src/sbt-test/jar/classpath-jar/src/main/scala/test/Test.scala
```diff
@@ -3,17 +3,27 @@ package test
...
+
   val system = ActorSystem("testSystem")
+  import system.dispatcher
+
   val router = system.actorOf(Props[PrintActor])
-  router ! "SUCCESS!!"
-  system.shutdown
+  (router ? "SUCCESS!!").foreach { _ =>
+    system.shutdown
+  }
```
fsat: @muuki88 - this is not related to the RPM symlink issue, but I have intermittent failure on this test today, and hence the fix.

The failure is originally caused by the race condition where the actor system may be shutdown before the actor under test has the chance to print the expected message.

fsat:
@muuki88 - if you don't have any further feedback, can you merge this and release another milestone version please?

muuki88:
Released in `1.0.5-RC1`

huntc:
Thanks!

fsat:
Thank you @muuki88!

PR #689, opened 2015-11-04
add line break after installSection and scriptlets

shijinkui:
when install rpm, have below error.  Maybe there are no line break after `%install`

error: %pre(agent-1.0.30-1.noarch) scriptlet failed, exit status 1
error:   install: %pre scriptlet failed (2), skipping agent-1.0.30-1

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaAppBashScript.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaAppBashScript.scala
@@ -49,7 +49,7 @@ object JavaAppBashScript {
     val jarPrefixed = """^\-jar (.*)""".r
     val args = mainClass match {
       case jarPrefixed(jarName) => Seq("-jar", jarName)
-      case className           => Seq(className)
+      case className            => Seq(className)
     }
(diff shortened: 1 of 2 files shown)


muuki88:
I wait for #685 to be merged as this is a rather small change.
The build works with your fix?

shijinkui:
@muuki88 after custom spec, the rpm can be installed without error.

Another problem, startService show be split into register service and start service.
When I install the rpm, i need't start it right now, especially on some vm mirror, such as aws or aliyun ecs.

```
startService() {
    app_name=$1
    if hash update-rc.d >/dev/null 2>&1; then
		echo "Adding $app_name to autostart using update-rc.d"
		update-rc.d $app_name defaults
		service $app_name start
    elif hash chkconfig >/dev/null 2>&1; then
		echo "Adding $app_name to autostart using chkconfig"
		chkconfig --add emr-agent
		chkconfig $app_name on
		service $app_name start
    else
		echo "WARNING: Could not add $app_name to autostart: neither update-rc nor chkconfig found!"
    fi
}
```

muuki88:
@shijinkui you can override the different implementations in the `linuxScriptReplacements` AFAIK

shijinkui:
OK

Issue #633, opened 2015-07-30
use case: single project, multiple apps


wookietreiber:
As I understand it, from a cursory reading of your documentation, sbt-native-packager considers only a single main class and will create only a single script invoking that main class, is that correct?

The use case I am currently pondering on is having a single project with multiple main classes, i.e. multiple distinct apps, each of which should get its own script, e.g.:

``` scala
package foobar

object foo extends App with Library with Util
object bar extends App with Library with Util

trait Library
trait Util
```

Supposed to be deployed result:

```
bin/foo
bin/bar
lib/project.jar
```

Each of the scripts in `bin` uses the same `lib/project.jar` only with a different main class.

Is this possible to achieve right now? If not: Is sbt-native-packager able to be extended to support this use case?

metasim:
The way I handle this is via a `packaging` submodule, which in turn has separate submodules for each package I want to create. [Here's a rough idea](https://gist.github.com/metasim/b89ac245f0316d53dd9e) of what I'm talking about.

muuki88:
Submodules are one way to handle this. If you just want different startup scripts you could

- create a script for each main in `src/universal/bin`
- content of each script ( not tested, only to the get the idea )
```bash
#!/bin/bash
you-app-name -main path.to.Main
```

See #319

muuki88:
@metasim can you put your snippet [here](https://github.com/sbt/sbt-native-packager/wiki/Helpful-Snippets)? At some point we can move parts from there to the docs.

metasim:
@muuki88 How's this?: https://github.com/sbt/sbt-native-packager/wiki/Helpful-Snippets

muuki88:
@metasim sweet :)

muuki88:
@wookietreiber did one of the solutions solve your problem?

wookietreiber:
Sorry for not replying sooner, I had to postpone the project I had this feature in mind for a while.

Today I dove a bit into sbt tasks and the way I handled it for now is this:

```scala
val scripts = taskKey[Unit]("Creates the scripts.")

scripts := {
  val scriptDir = target.value / "scripts"
  if (!scriptDir.exists) scriptDir.mkdir()

  val prefix = sys.env.getOrElse("PREFIX", "/usr/local")

  def script(clazz: String) =
    s"""|#!/bin/sh
        |java -cp ${prefix}/share/grid-engine-tools/grid-engine-tools.jar $clazz
        |""".stripMargin

[...]

PR #691, opened 2015-11-05
Allow to override systemd start script name

rozky:
This allows to add "@" to the script name which is useful when using systemd. Any unit name containing "@" in name is used as template unit which allows to create multiple units from single start script. More information: https://wiki.archlinux.org/index.php/Systemd (in section "Using units" -> "Note")


muuki88:
Nice, pretty neat.

muuki88:
@rozky Could make another pull request adding a small test and a bit of documentation how to use this?

muuki88:
This could also be a nice way to define multiple `instances` pointing each to a different main class. #633

PR #694, opened 2015-11-11
Fix systemv restartService function

finlob:


muuki88:
Thanks a lot! I'm really missing integration tests :(

PR #696, opened 2015-11-12
do not overwrite the PIDFILE is set if env_config

yanns:


yanns:
this change is not sufficient. We have to remove the `--make-pidfile` if the application manages the PID file itself (like play does)

PR #697, opened 2015-11-13
fix SystemV init script template for debian packaging

yanns:
For processes that manage their own pid file, they should define the `PIDFILE` variable (sourced either through `/etc/default/…` or `/etc/sysconfig/…`). Otherwise, if undefined, this script will capture the pid and write it to `/var/run/${{app_name}}/running.pid`.

This commit is related to https://github.com/sbt/sbt-native-packager/commit/45bc5dbaffdca8eaef4b498c6b93e3e03350428d that implements the same logic for rpm packaging

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/etc-default-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/etc-default-template
@@ -2,7 +2,7 @@
 # ##### Environment Configuration #####
 # #####################################
 
-# This file gets sourced before the actual bashscript
+# This file gets sourced before the actual startscript
 # gets executed. You can use this file to provide
 # environment variables
 
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/systemv/start-debian-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/java_server/systemloader/systemv/start-debian-template
@@ -18,7 +18,12 @@ source /lib/lsb/init-functions
 # $JAVA_OPTS used in $RUN_CMD wrapper
 export JAVA_OPTS
 
-PIDFILE=/var/run/${{app_name}}/running.pid
+# If program manages its own PID file then it
+# should declare its location in PIDFILE
+if [ -z "$PIDFILE" ]; then
+    create_pidfile=true
+    PIDFILE=/var/run/${{app_name}}/running.pid
+fi
 
 if [ -z "$DAEMON_USER" ]; then
     DAEMON_USER=${{daemon_user}}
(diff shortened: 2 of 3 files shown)


kardapoltsev:
One note about stopping daemon. We probably shouldn't remove `PIDFILE` in case when it managed by app itselt.

yanns:
@kardapoltsev good point.
But for a  play application for example, the stop script must know the pid to send a kill signal to the application. I think we need the PIDFILE for the stop.

kardapoltsev:
We need `PIDFILE` for stopping service but I think that it shouldn't be removed by start script. Play application removes its own pid file as I remember.

muuki88:
@yanns code looks good to me. Things I would like to see before merging this:

- Adding documentation in the [play section](http://www.scala-sbt.org/sbt-native-packager/topics/play.html) and the [server configuration](http://www.scala-sbt.org/sbt-native-packager/archetypes/java_server/customize.html#linux-configuration)
- a `log-daemon` msg which path is used ( _pid file managed by app_ or _pid file managed by start-stop daemon_ )

Thanks for making the SystemV script more robust :)

yanns:
@kardapoltsev I now understood what you mean. You're right. I'll change that.

yanns:
@muuki88 the play section and the server configuration already contain the needed information.
It was implemented in the rpm start script but not in the debian one (see the linked commit)

OK for the msg. I'll add that.

muuki88:
@yanns yes, it's added to the docs, but it states that this is for `rpm / systemv`, where now you would have to add it for `debian` as well. Which will again create some confusion for play users.

yanns:
@muuki88 yes you're right, I oversaw that in the play doc.
But in the server doc, it's only mentioning `systemv`. There is here nothing to change I guess.

yanns:
I tested writing the msg and I am not sure we should do it:
```bash
$  /etc/init.d/my-app start
 * Starting my-app
using PID file '21747PIDFILE' managed by application
                                                                                               [ OK ]
```
It breaks the output.

muuki88:
ewy. Okay, we leave it as is.

yanns:
@kardapoltsev @muuki88 please review the new version that contain changes according to your feedbacks.

muuki88:
LGTM

kardapoltsev:
LGTM too except that @muuki88 mentioned log message about pid file management.
> a log-daemon msg which path is used ( pid file managed by app or pid file managed by start-stop daemon )

yanns:
@kardapoltsev please have a look at the previous discussions.
I tried to add the msg and it breaks the output. With @muuki88 we decided not to make it.

kardapoltsev:
Excuse me, I've missed it. Thanks for your contributions!

yanns:
cool thanks!

PR #700, opened 2015-11-23
Removed default-jre | java6-runtime dependency in jdeb packaging

kardapoltsev:


muuki88:
LGTM 
This should improve packages for beginners :)

PR #701, opened 2015-11-25
Different start script replacements for different system loaders

kardapoltsev:
The reason is that somebody may want to build rpm with customized systemV script and deb with default upstart script.


muuki88:
Pretty nice! IIRC there is a hint on the `cheatsheet` page (docs) for this templating as well. Could you check if this is still correct? If yes feel free to merge

kardapoltsev:
Yes, I forget to fix this page. Will do it soon.

muuki88:
Awesome :)

PR #711, opened 2015-12-03
Ensure systemd works with RPM relocatable packages

fsat:
Ensure correct installation path is being used in the service file required by systemd.

This is done by replacing the default installation path with the actual installation path
within the service file during install time of the RPM package.


fsat:
@muuki88 and @huntc - can I get a PR review on this please?

This fix will ensure systemd service file has the correct install path when the RPM package is being relocated during package install time.

muuki88:
LGTM. Could you have a look at #698 for me in return :)

fsat:
@muuki88 - thanks for looking at this, and sorry for not getting to #698 sooner.

muuki88:
@fsat no problem, really appreciate your help :)

PR #714, opened 2015-12-13
Travis: build on OS X

larsrh:
This only runs `jar/*` scripted tests, because all the others have dependencies which are not available on the Travis workers: `universal` requires `xz`, the others Linux packaging tools.


larsrh:
I believe the build on 8e15f97 failed because I already pushed a new revision before the Linux tests even started, leading to an out-of-sync `.travis.yml`. Anyway, the build for db60e99 seems to be chugging along just fine.

muuki88:
Nice. I will take a closer look when I am home

larsrh:
Don't merge just yet. There might be a better approach: Since `brew` is available, it might be possible to install `xz` (larsrh@b697bc1).

larsrh:
Superseded by #715.

PR #715, opened 2015-12-13
Travis: build on OS X (take 2)

larsrh:
Supersedes #714.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,14 @@
-language: scala
+language: java
 os:
-  - linux
+  - osx
+env:
+  - SCALA_VERSION=2.10.5
 script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,13 +1,19 @@
-language: scala
+language: java
```
muuki88: why is this change necessary?

larsrh: The `sbt` launcher script is unavailable on OS X anyway.

larsrh: (The other option would be to not include it in the repository but fetch it in `before_install`.)

muuki88:
Having a dedicated launcher script is worth it, when we have OS X builds now. Thanks a lot for your effort :)

lunaryorn:
@larsrh Psst, wise man use force push, young padawan :sunglasses:  :stuck_out_tongue_winking_eye:

PR #719, opened 2015-12-25
Mark docker and jdeb as provided dependencies

muuki88:
Follow up to the discussion in #693

## Goals

- java-based packaging alternatives dependencies are marked as `provided`
- less dependencies per default, faster builds, less dependency conflict potential

PR #720, opened 2016-01-01
Adding ant dependency for jdkpackager builds. Follow up to #719

muuki88:
Changing `jdeb` to a provided dependency removed the `ant` dependency from the runtime  classpath, which is needed for the `jdkpackager` plugin.

Adding travis tests and dependency.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,9 @@ script:
   - if [[ "$TRAVIS_OS_NAME" = "linux" ]]; then
       ./sbt ++$SCALA_VERSION "scripted rpm/* debian/*";
     fi
+  - if [[ "$TRAVIS_JDK_VERSION" = "oraclejdk8" ]]; then
+      ./sbt ++$SCALA_VERSION "scripted jdkpackager/*";
+    fi
 notifications:
   email:
     - qbranch@typesafe.com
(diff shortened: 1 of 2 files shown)


muuki88:
@metasim you have any idea why this test breaks on `travis` ?

> test-package-image/target/jdkpackager/build.xml:25: fx:info doesn't support the nested "association" element.

On my machine (Ubuntu 15.10, java version "1.8.0_66") things work.

metasim:
@muuki88  I need to check the logs more, but my guess is that the travis config had JDK version less than 1.8.0_40, which is when the file association capability was added. Any idea if travis allows for finer grained Java version specifications?

metasim:
Indeed, the failed build uses an older version of Java 1.8.0 (why the hell Oracle couldn't bump the minor revision number instead of this build number crap is beyond me).

    java version "1.8.0_31"

muuki88:
I'm afraid travis has no other way to specify the minor/patch versions in oracle-jdk. For now I ignore this test and add it when travis has upgraded their java version.

Thanks for looking into :)

muuki88:
There is an issue (over a year old) which states that travis will upgrade soon (note in november) [#3259](https://github.com/travis-ci/travis-ci/issues/3259)

Issue #518, opened 2015-03-07
proguard support


joprice:
Any tips on integration with sbt-proguard?

muuki88:
Haven't used it yet, but if you have a working example, I am interested.

muuki88:
Any progress @joprice ?

mikebridge:
@muuki88 I was able to create a Raspbian .deb file for a scala service with sbt-native-packager but it's huge.  Is there any way to get proguard to process the jar before it is packaged?  Would it require a separate Proguard AutoPlugin implementation inside sbt-native-packager?

muuki88:
To work with native-packager proguard needs to support multiple jars as input and stripping out the class files in each of them. IMHO the best way would be to use native packager [together with sbt-assembly](http://www.scala-sbt.org/sbt-native-packager/topics/custom.html#sbt-assembly) and use proguard on the assembly jar. 

Also: https://github.com/sbt/sbt-proguard

mikebridge:
@muuki88 Thanks---I am doing this in a multiple-project setup, and my `build.sbt` now contains something like this:

```scala
lazy val  mysubproject = project.in(file("mysubproject")).
  enablePlugins(JavaAppPackaging).
  settings(proguardSettings: _*).
  settings(assemblySettings: _*).
  settings(
    name := "mysubproject",

    version := "0.0.1",

    serverLoading in Debian := ServerLoader.SystemV,

    debianPackageDependencies in Debian := Seq("mosquitto"),

    jarName in assembly := "mysubproject.jar",

    mappings in Universal := {
      val universalMappings = (mappings in Universal).value
      val fatJar = (assembly in Compile).value
      val filtered = universalMappings filter {
        case (file, name) =>  ! name.endsWith(".jar")
      }
      filtered :+ (fatJar -> ("lib/" + fatJar.getName))
    },

    ProguardKeys.options in Proguard ++= Seq("-dontnote", "-dontwarn", "-ignorewarnings"),

    ProguardKeys.options in Proguard += ProguardOptions.keepMain("com.example.Main"),

    ProguardKeys.proguardVersion in Proguard := "5.2",

    scriptClasspath := Seq( (jarName in assembly).value ),
    
    // ... etc.
))
```

Running `myproject/universal:package-zip-tarball` will now create and package the fat jar (nice!), and `myproject/proguard:proguard` will create a jar and run proguard on it, but I'm at a loss at this point about what to do next to integrate them together....

[...]

PR #721, opened 2016-01-06
Add Proguard example to Custom Package Format documentation

mikebridge:
As per #518


muuki88:
Awesome! Thanks for the comprehensive guidw

PR #733, opened 2016-01-26
Docker Plugin qualification

matthughes:
Since sbt-native already provides a DockerPlugin, many people get confused as to the difference between the one included in sbt-native and the `sbt-docker` one linked on the homepage.  The question has come up several times in gitter.


muuki88:
thanks :)

PR #745, opened 2016-02-21
SystemD services now source /etc/default/{{app_name}} (resolves #737)

### LINKED ISSUE #737: systemd scripts completely ignoring `/etc/default/{{package-name}}`
The package is including the file, and SystemV scripts properly reference it, but the `SystemD` template is completely ignoring it.

Perhaps it should set the EnvironmentFile [EnvironmentFile](http://0pointer.de/public/systemd-man/systemd.exec.html) directive? 

> Similar to Environment= but reads the environment variables from a text file. The text file should contain new-line-separated variable assignments. Empty lines, lines without an "=" separator, or lines starting with ; or # will be ignored, which may be used for commenting. A line ending with a backslash will be concatenated with the following one, allowing multiline variable definitions. The parser strips leading and trailing whitespace from the values of assignments, unless you use double quotes (").
> 
> The argument passed should be an absolute filename or wildcard expression, optionally prefixed with "-", which indicates that if the file does not exist, it will not be read and no error or warning message is logged. This option may be specified more than once in which case all specified files are read. If the empty string is assigned to this option, the list of file to read is reset, all prior assignments have no effect.
> 
> The files listed with this directive will be read shortly before the process is executed (more specifically, after all processes from a previous unit state terminated. This means you can generate these files in one unit state, and read it with this option in the next).
> 
> Settings from these files override settings made with Environment=. If the same variable is set twice from these files, the files will be read in the order they are specified and the later setting will override the earlier setting.

I'm not sure what the drawbacks are. Seems like some people prefer `/etc/sysconfig`, and [others prefer just adding service.d files](http://serverfault.com/questions/413397/how-to-set-environment-variable-in-systemd-service).

¯\_(ツ)_/¯

timcharper:
This change modifies the SystemD template to include the packaged
/etc/default environment settings file. Since a sourced SystemD
environment file is not the same thing as a bourne shell source script,
a different template is used.

Those deploying their package to multiple platforms can also specify
different /etc/default templates by suffixing `-systemd` to their name,
ie: src/templates/etc-default-systemd

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ project/project
 log/
 target/
 .cache
+.ensime*
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
@@ -39,8 +39,6 @@ object JavaServerAppPackaging extends AutoPlugin {
   /** These settings will be provided by this archetype*/
   def javaServerSettings: Seq[Setting[_]] = linuxSettings ++ debianSettings ++ rpmSettings
 
-  protected def etcDefaultTemplateSource: java.net.URL = getClass.getResource(ETC_DEFAULT + "-template")
-
   /**
    * general settings which apply to all linux server archetypes
    *
@@ -59,27 +57,30 @@ object JavaServerAppPackaging extends AutoPlugin {
     },
     // === etc config mapping ===
     bashScriptEnvConfigLocation := Some("/etc/default/" + (packageName in Linux).value),
-    linuxEtcDefaultTemplate <<= sourceDirectory map { dir =>
-      val overrideScript = dir / "templates" / ETC_DEFAULT
-      if (overrideScript.exists) overrideScript.toURI.toURL
-      else etcDefaultTemplateSource
-    },
-    linuxStartScriptName := None,
-    makeEtcDefault <<= (packageName in Linux, target in Universal, linuxEtcDefaultTemplate, linuxScriptReplacements)
-      map makeEtcDefaultScript,
-    linuxPackageMappings <++= (makeEtcDefault, bashScriptEnvConfigLocation) map { (conf, envLocation) =>
-      val mapping = for (
-        path <- envLocation;
-        c <- conf
-      ) yield LinuxPackageMapping(Seq(c -> path), LinuxFileMetaData(Users.Root, Users.Root, "644")).withConfig()
 
-      mapping.toSeq
-    }
+    linuxStartScriptName := None
+  )
 
+  /* etcDefaultConfig is dependent on serverLoading (systemd, systemv, etc.),
+   * and is therefore distro specific. As such, these settings cannot be defined
+   * in the global config scope. */
+  private[this] val etcDefaultConfig: Seq[Setting[_]] = Seq(
+    linuxEtcDefaultTemplate := getEtcTemplateSource(
+      sourceDirectory.value,
+      serverLoading.value),
+    makeEtcDefault := makeEtcDefaultScript(
+      (packageName in Linux).value,
+      (target in Universal).value,
+      linuxEtcDefaultTemplate.value,
+      linuxScriptReplacements.value),
+    linuxPackageMappings ++= etcDefaultMapping(
+      makeEtcDefault.value,
+      bashScriptEnvConfigLocation.value)
   )
 
   def debianSettings: Seq[Setting[_]] = {
     import DebianPlugin.Names.{ Preinst, Postinst, Prerm, Postrm }
+    inConfig(Debian)(etcDefaultConfig) ++
     inConfig(Debian)(Seq(
       serverLoading := Upstart,
       startRunlevels <<= (serverLoading) apply defaultStartRunlevels,
@@ -131,6 +132,7 @@ object JavaServerAppPackaging extends AutoPlugin {
 
   def rpmSettings: Seq[Setting[_]] = {
     import RpmPlugin.Names.{ Pre, Post, Preun, Postun }
+    inConfig(Rpm)(etcDefaultConfig) ++
     inConfig(Rpm)(Seq(
       serverLoading := SystemV,
       startRunlevels <<= (serverLoading) apply defaultStartRunlevels,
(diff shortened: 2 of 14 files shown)


timcharper:
@muuki88, this change is done and tested (lamentably, only under Debian Linux). What are your thoughts on my approach? I'd like to get some feedback on it before recommending it for merge.

.gitignore
```diff
@@ -9,3 +9,4 @@ project/project
 log/
 target/
 .cache
+.ensime*
```
muuki88: pure personal interest: what editor do you use with ensime?

kardapoltsev: I'd like to recommend emacs + evil (:

src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
```diff
@@ -59,27 +57,26 @@ object JavaServerAppPackaging extends AutoPlugin {
     },
     // === etc config mapping ===
     bashScriptEnvConfigLocation := Some("/etc/default/" + (packageName in Linux).value),
-    linuxEtcDefaultTemplate <<= sourceDirectory map { dir =>
```
muuki88: These were redundant, right? As we only use `debian:packageBin` or `rpm:packageBin`

src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
```diff
@@ -59,27 +57,26 @@ object JavaServerAppPackaging extends AutoPlugin {
...
-    }
+    linuxStartScriptName := None
+  )
 
+  /* etcDefaultConfig is dependent on serverLoading (systemd, systemv, etc.),
+   * and is therefore distro specific. As such, these settings cannot be defined
+   * in the global config scope. */
+  private val etcDefaultConfig: Seq[Setting[_]] = Seq(
+    linuxEtcDefaultTemplate <<= (sourceDirectory, serverLoading)
+      map getEtcTemplateSource,
```
muuki88: We try to avoid the sbt symbolic operators. Instead of
```
linuxEtcDefaultTemplate <<= (sourceDirectory, serverLoading) map getEtcTemplatSource
```

we try to clean up the code base using

```
linuxEtcDefaultTemplate := getEtcTemplatSource(sourceDirectory.value, serverLoading.value)
```

In some cases this adds a few extra stuff, but makes it easier to understand for newcomers.

src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
```diff
@@ -59,27 +57,26 @@ object JavaServerAppPackaging extends AutoPlugin {
...
+    makeEtcDefault <<= (packageName in Linux, target in Universal, linuxEtcDefaultTemplate, linuxScriptReplacements)
+      map makeEtcDefaultScript,
+    linuxPackageMappings <++= (makeEtcDefault, bashScriptEnvConfigLocation) map
+      etcDefaultMapping
   )
 
   def debianSettings: Seq[Setting[_]] = {
     import DebianPlugin.Names.{ Preinst, Postinst, Prerm, Postrm }
+    inConfig(Debian)(etcDefaultConfig) ++
```
muuki88: +1

src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
```diff
@@ -242,6 +240,33 @@ object JavaServerAppPackaging extends AutoPlugin {
...
   }
 
+  private[this] def getEtcTemplateSource(sourceDirectory: File, loader: ServerLoader): java.net.URL = {
+    val suffix = loader match {
+      case Upstart => "-upstart"
+      case SystemV => "-systemv"
+      case Systemd => "-systemd"
+    }
+
+    val overrides = List[File](sourceDirectory / "templates" / (ETC_DEFAULT + suffix), sourceDirectory / "templates" / ETC_DEFAULT)
```
muuki88: Slick implementation. Keeping things backwards compatible ( and easy to use ), but add customized behavior.

muuki88:
I like the implementation a lot. We add further customization without touching the existing solution. If you add one test for the customization and some documentation this PR is from my perspective ready to merge

.gitignore
timcharper: That's it! Emacs and evil :)

src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
timcharper: No, they didn't work anymore in the global config scope since I added the dependency for `serverLoading`, and that config key is not defined globally.

src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala
```diff
@@ -59,27 +57,30 @@ object JavaServerAppPackaging extends AutoPlugin {
...
-      mapping.toSeq
-    }
 
+    linuxStartScriptName := None
+  )
+
+  /* etcDefaultConfig is dependent on serverLoading (systemd, systemv, etc.),
+   * and is therefore distro specific. As such, these settings cannot be defined
+   * in the global config scope. */
+  private val etcDefaultConfig: Seq[Setting[_]] = Seq(
```
muuki88: Codacy Issue found: [List(etcDefaultConfig) should have a smaller scope by using private[this]](https://www.codacy.com/app/nepomukseiler/sbt-native-packager/file/2094166179/issues/source?bid=2040925&fileBranchId=3140198#l67)

timcharper: I don't understand this; `private[this]` is not smaller than `private`.

timcharper:
```
scala> trait A { private val a = 1 }
defined trait A

scala> object B extends A { println(a) }
<console>:11: error: not found: value a
       object B extends A { println(a) }
                                    ^
```

timcharper: What ? Codacy is clearly wrong here.

timcharper:
@muuki88 Quick ping to say I've done as you've requested and everything's green again.

muuki88:
Awesome. Thanks a lot, I will try to release an RC

timcharper:
\o/

PR #748, opened 2016-03-01
Fix errors when .bat file path contains paranthesis.

szdmr:
Without double quotes the path with paranthesis gets evaluated causing an error on Windows systems. Specifically the error text is "\\\.. was unexpected at this time."

I got the error when I moved my Play distro inside a Dropbox folder with name "Dropbox (COMPANYNAME)". The path (COMPANYNAME) part gets evaluated as the if condition clause and makes the batch throw the error above and close. 

Double quoting the %CFG_FILE% variable seems to be fixing the error.

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/bat-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/bat-template
@@ -20,7 +20,7 @@ for %%x in (!cmdcmdline!) do if %%~x==/c set DOUBLECLICKED=1
 rem FIRST we load the config file of extra options.
 set "CFG_FILE=%@@APP_ENV_NAME@@_HOME%\@@APP_ENV_NAME@@_config.txt"
 set CFG_OPTS=
-if exist %CFG_FILE% (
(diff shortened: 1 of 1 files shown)


typesafehub-validator:
Hi @szdmr,

szdmr:
FYI I signed the Typesafe Contributors License Agreement.

muuki88:
Thanks for your PR. I'm not too familiar with the windows stuff, so
@nazoking @bjuric do you have any objections?

muuki88:
@szdmr on which systems did you test the change? I really would like to merge this change, but we had some rough experiences with the windows bat script :(

szdmr:
@muuki88 tried it on Windows 8.1 and Windows 10 only, both 64 bits. I don't know if it breaks anything for Windows 8 or below, but do not expect it to do.

It would fail if %CFG_FILE% contained a double quote, but that's impossible as Windows does not let a file or folder to contain double quote in names.

muuki88:
thanks for the clarification :+1:

PR #757, opened 2016-03-20
FIX #676 adding configuration scopes to project configurations

### LINKED ISSUE #676: Auto plugins are required to declare their project configurations
Auto plugins are required to register any configurations that they may introduce in order to allow a user to declare any addition configurations of their own. For example, given some other configuration named `Backend`:

```scala
lazy val root = (project in file("."))
  .enablePlugins(JavaAppPackaging)
  .configs(Backend)
```

...a `bundle:dist` will subsequently fail as the plugin's configurations are effectively zapped. 

Each new auto plugin that introduces a configuration should declare the configurations that it provides so that they may be prepended to declarations such as the one above. The method to override for each auto plugin is `projectConfigurations`. Here's an example from sbt-bundle (pending a PR):

```scala
  override def projectConfigurations: Seq[Configuration] =
    Seq(
      Bundle,
      BundleConfiguration
    )
```

muuki88:
Adding format plugin configurations to `projectConfigurations` as described in #676 

cc @huntc


huntc:
It isn't clear to me what we are trying to do here at a glance. Could you please enhance the description of the PR?

muuki88:
Sorry @huntc I just referenced the issue #676 , which was a request to add the AutoPlugin scopes to the `projectConfigurations`

huntc:
Ah sorry I didn't see that ref, and I'd forgotten about the issue too!

LGTM

Issue #754, opened 2016-03-14
Upstart script does not use killTimeout


elm-:
We are using upstart on ubuntu and the default kill timeout of 5 seconds is not overwritten by setting of LinuxKeys.killTimeout in the settings. I checked the template and it does not contain the relevant placeholder. is there a reason or can this be added?

muuki88:
Thanks for your feature request @elm- 

There is no reason for that, it wasn't simply implemented. If you could create a small PR adding this, would be awesome :)

elm-:
@muuki88 thanks a lot, sorry I actually missed your comment to provide a pull request, I would have done that otherwise.

muuki88:
No problem. We made some bigger changes to the systemloader feature, which makes a a lot easier now to integrate such specific changes.

There will always be a next time for a pull request :) For instance #770 ;)

Issue #728, opened 2016-01-18
systemd service customization (open files)


francisdb:
Would be nice to have a way to customise the systemd service file the same way as we can do with the bat/bash files. More specifically we would need to add the open files limit option:

http://www.freedesktop.org/software/systemd/man/systemd.exec.html

Eg: 
When using systemd you might need to set LimitNOFILE=10500 in the service file.
Manual way to do this:

```
sudo pico /usr/lib/systemd/system/myapp.service
# in the service section add:
# LimitNOFILE=10500
sudo systemctl daemon-reload
sudo systemctl restart myapp.service
```

For the bash file this can be done like this (when not using systemd)

```
bashScriptExtraDefines += """ulimit -n 10500"""
```

Might even be interesting to have a unified way to set up open file limits...

In case you want to see the applied limits to your service

``` bash
cat /proc/$(cat /var/run/myapp/play.pid)/limits
```

muuki88:
At the moment you have the following options

### Override the complete script

This is documented [here](http://www.scala-sbt.org/sbt-native-packager/archetypes/java_server/customize.html#override-start-script). You have to configure everything your self.

We should add that to the [archetype cheatsheet](http://www.scala-sbt.org/sbt-native-packager/archetypes/cheatsheet.html)

### Hack it into predefined settings

There are currently four settings

- `startRunlevels`
- `stopRunlevels`
- `requiredStartFacilities`
- `requiredStopFacilities`

which are essentially strings. You could add your settings in there, but this is rather bad.

## Solution

I like the idea of a generic data structure where you can append things which will be added to the start script. Most of the heavy lifting is done in [JavaServerBashScript](https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerBashScript.scala), which reads the templates and substitutes the placeholders.

However the system-loaders are so different, it's hard to provide a generic setting that fits for all. For `SystemD` and `Upstart` the scripts are actually  that simple, you can copy n paste them and mark your adjustments there.

@francisdb if you have a good idea how to make things work for all systemloaders in an easy way?

francisdb:
Overriding the script is always a solution but not preferable as you might forget to update your copy when the source is updated (bug fixes, refactorings, ...)

Even appending might not fit all as some systems. I fear it will be quite hard to come up with a solution for all loaders/supported packagings.

I guess this will have to be a case-by-case solution or the override solution

[...]

Issue #679, opened 2015-10-12
[bug] wrong systemd script file path in debian


ptonelli:
The systemd path for the service script for debian in JavaServerApp is 
`/usr/lib/systemd/system`

Following the Debian packaging [documentation](https://wiki.debian.org/Teams/pkg-systemd/Packaging)

It should be:
`/lib/systemd/system`

The first path seems to be correct for RedHat based distributions ([doc](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/System_Administrators_Guide/chap-Managing_Services_with_systemd.html))

muuki88:
Which system fails for you? On Ubuntu `15.10` this seems to work. 
A fix would be rather easy ( if you like to open a pull request):

- [Debian startscript mapping](https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala#L100)
- [Get Startscript Location](https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala#L230-L236)

ptonelli:
I am under debian jessie (8.2)

the official path in ubuntu seems to be the same as in debian ([doc](https://wiki.ubuntu.com/systemd)).

You are right, it works even with the `/usr/lib/systemd/system` path even in debian.

muuki88:
No Problem :) Thanks for the detailed reporting anyway.

PR #785, opened 2016-04-25
Extract Systemloaders into AutoPlugins

muuki88:
# Systemloader Plugin - Proposal

Goal of this PR is to refactor the `JavaServerAppPackaging` autoplugin and make system-loaders more flexible and configurable.

The main idea is to split the systemloaders in separate AutoPlugins. Usage would look like this

```scala
enablePlugins(JavaServerAppPackaging, SystemdPlugin)
```

This would trigger all the special server-folders and the configuration of a systemd start script.

Pro | Con
------------ | -------------
Systemloaders are truly isolated and can evolve independetly | No easy way to built a package with multiple systemloaders
Systemloaders can  be used with other archetypes | Possible breaking changes

Squeezing every customization into the `JavaServerAppPackaging` doesn't scale well codewise. The main pain point is, that things like this won't work anymore

```scala
serverLoading in Rpm = SystemV,
serverLoading in Debian = Systemd
```

We probably could allow something like

```scala
enablePlugins(Systemd, SystemV)
```

when break our API and really isolate the systemloaders. However I'm not sure if this is a common need to

- specify different loaders
- package with multiple loaders

The solution would, again, be submodules with different configurations.

# Related issues

- #754
- #728
- #679
- #770 

cc: @eed3si9n @dwhjames @fiadliel @metasim @kardapoltsev


src/main/scala/com/typesafe/sbt/packager/archetypes/SystemD.scala
```diff
@@ -0,0 +1,24 @@
...
+import com.typesafe.sbt.packager.linux.LinuxPlugin.autoImport.serverLoading
+
+object SystemD extends AutoPlugin {
+
+  object autoImport {
+    // all systemd specific settings/tasks here
+    val systemdExitSuccessStatus = settingKey[Int]("defines the ExitSuccessStatus for systemd")
+  }
+
+  import autoImport._
```
muuki88: Codacy Issue found: [Put imports in the beginning of the package](https://www.codacy.com/app/nepomukseiler/sbt-native-packager/file/2460298881/issues/source?bid=2040925&fileBranchId=3254858#l16)

kardapoltsev:
It's a good idea IMHO.

dpennell:
What would it even mean to have multiple system loaders?
- build one RPM that tries to use multiple system loaders?  
 - Does it use all available (yikes)?
 - One is default, alternates can be selected at install time?
Unless somebody steps forward with a concrete use-case, I think you should proceed.

muuki88:
Thanks for your feedback. I will create an implementation see how everything plays out.

muuki88:
I will merge this and release as `1.2.0-M1`. From there we can tackle all issues related to systemloader configuration issues.

PR #787, opened 2016-04-26
Don't compress debian packages

pauldraper:
Representative example:

A Play project takes 30 seconds for `dpkg-deb` to build a 70MB deb with compression.
The same project takes 1 second to build a 79MB deb without compression.

Since JARs are already compressed archives, a sensible default is to not compress them again. (In fact, I'd be surprised if the small gains in client downloads weren't almost entirely overcome by the time to decompress.)

Users can use `debianNativeBuildOptions` to override options for `dpkg-deb` if they would like a different compression type, level, etc.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/debian/NativePackaging.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/debian/NativePackaging.scala
@@ -43,6 +43,7 @@ trait DebianNativePackaging extends DebianPluginLike {
    * package.
    */
   private[debian] def debianNativeSettings: Seq[Setting[_]] = inConfig(Debian)(Seq(
+    debianNativeBuildOptions += "-Znone", // packages are largely JARs, which are already compressed
     genChanges <<= (packageBin, target, debianChangelog, name, version, debianPackageMetadata) map {
       (pkg, tdir, changelog, name, version, data) =>
         changelog match {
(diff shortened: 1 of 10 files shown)


muuki88:
Thanks @pauldraper for your contribution. Disabling compression sounds like a reasonable default. Also you could achieve a similar effect by using `JDebPackaging`.

Your code looks good. Could you implement a new test that sets the `debianNativeBuildOptions := Nil` and shows that it still works. Also this new setting should be added to the [Debian Plugin docs](http://sbt-native-packager.readthedocs.org/en/latest/formats/debian.html)

pauldraper:
Tests and documentation added.

src/main/scala/com/typesafe/sbt/packager/debian/NativePackaging.scala
```diff
@@ -43,6 +43,7 @@ trait DebianNativePackaging extends DebianPluginLike {
    * package.
    */
   private[debian] def debianNativeSettings: Seq[Setting[_]] = inConfig(Debian)(Seq(
+    debianNativeBuildOptions += "-Znone", // packages are largely JARs, which are already compressed
```
muuki88: We shouldn't scope this. The setting is already prefixed with _debian_. I just realised it after reading the tests.

muuki88: If this is a bit too much refactoring  for you, I can adjust that after merging.

pauldraper: It seems that `debian...` keys are scoped to the `Debian` config whenever possible.

https://github.com/pauldraper/sbt-native-packager/blob/pauldraper-deb-compress/src/main/scala/com/typesafe/sbt/packager/debian/DebianPlugin.scala#L170

muuki88: This more a bug then a feature :( I think we clean this up in another PR. Thanks a lot for your work :)

PR #766, opened 2016-04-06
AshScriptPlugin - Restore argument behavior as it was before a037519

dhoepelman:

PR #792, opened 2016-05-14
fix #766

giabao:
I don't fully understand this: `set -- $(loadConfigFile "$script_conf_file") "$opts"`
But I tested in alpipe:3.3  docker image:
```bash
docker run --rm -ti alpine sh
```

inside alpipe:
```bash
cat <<EOF > application.ini
-Dfoo="-J-Xms512M"
-J-Xms512M
-Xmx512M
EOF

touch test && chmod +x test

cat <<'EOF' > test
#!/bin/sh
loadConfigFile() {
  cat "$1" | sed '/^\#/d'
}
set -- $(loadConfigFile "application.ini") "$opts"
echo "[$opts]"
EOF

# ./test 
[]

cat <<'EOF' > test
#!/bin/sh
loadConfigFile() {
  cat "$1" | sed '/^\#/d'
}
set -- $(loadConfigFile "application.ini") "$@"
echo "[$@]"
EOF

# ./test 
[-Dfoo="-J-Xms512M" -J-Xms512M -Xmx512M]

cat <<'EOF' > test
#!/bin/sh
loadConfigFile() {
  cat "$1" | sed '/^\#/d'
}
set -- $(loadConfigFile "application.ini")
echo "[$@]"
EOF

# ./test 
[-Dfoo="-J-Xms512M" -J-Xms512M -Xmx512M]

cat <<'EOF' > test
#!/bin/sh
loadConfigFile() {
  cat "$1" | sed '/^\#/d' | sed 's/^-J-X/-X/' | tr '\r\n' ' '
}
opts=$(loadConfigFile "application.ini")
echo "[$opts]"
EOF

# ./test 
[-Dfoo="-J-Xms512M" -Xms512M -Xmx512M ]
```

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/ash-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/ash-template
@@ -29,7 +29,7 @@ realpath () {
 
 # Loads a configuration file full of default command line options for this script.
 loadConfigFile() {
-  cat "$1" | sed '/^\#/d'
+  cat "$1" | sed '/^\#/d' | sed 's/^-J-X/-X/' | tr '\r\n' ' '
 }
 
 
@@ -40,7 +40,6 @@ lib_dir="$(realpath "${app_home}/../lib")"
 ${{template_declares}}
 
 # If a configuration file exist, read the contents to $opts
-[ -f "$script_conf_file" ] && set -- $(loadConfigFile "$script_conf_file") "$opts"
+[ -f "$script_conf_file" ] && opts=$(loadConfigFile "$script_conf_file")
 
 java -classpath $app_classpath $opts $app_mainclass $@
-


muuki88:
Thanks for your pull request. What is the problem you'll trying to fix with this PR?

giabao:
1. The current code `set -- $(loadConfigFile "$script_conf_file") "$opts"` **do NOT load configs** into $opts
(the 1st `./test` output `[]`)

2. In the `bash-template` (not `ash-template`), the corresponding statement is `set -- $(loadConfigFile "application.ini") "$@"` - which correctly load configs into `$@`
(the 2nd `./test` output `[-Dfoo="-J-Xms512M" -J-Xms512M -Xmx512M]`)

3. When ommit `$@` from the statement `set -- $(loadConfigFile "application.ini") "$@"` - as in the 3rd `./test` then `$@` is still correctly set. So I said _"I don't fully understand.. "_

4. The last `./test` is to confirm that the changed code of this PR is correctly.
Note that in the `bash-template`, `-J-Xms512M` is passed to the bash script in which the `-J` is removed before passing the java options to `java`.
So here, in `ash-template` we should also remove `-J`. Otherwise the following sbt settings will **caused error** in `ash` but not in `bash`:
`javaOptions in Universal += "-J-Xmx512M"`

muuki88:
Thanks a lot for the clarification. 

Regarding point *4.*. Would you like to provide another pull request fixing this?

PR #798, opened 2016-05-26
Fix typo in debian systemv script.

jan0sch:
An `if` construct was not properly closed by `fi` which led to errors
when trying to start the system service.


kardapoltsev:
Good job, thanks!

Issue #780, opened 2016-04-19
Incorrect namespace with WiX 4.x


smoes:
When trying to create an msi with the generated .wix file (with candle, newest WiX toolset, v4.0.3922.0) I get the following error:

    error CNDL0199 : The Wix element has an incorrect namespace of   'http://schemas.microsoft.com/wix/2006/wi'. 
    Please make the Wix element look like the following: <Wix xmlns="http://wixtoolset.org 

Works well with v3.X Versions of the WiX toolset, so I assume it breaks with v4.X.

muuki88:
Are you using `jdkpackager:packageBin` or `windows:packageBin`?

smoes:
`windows:packageBin` 

Cmake seems to face similar problems when switching to WiX toolset 4.x: https://cmake.org/Bug/view.php?id=16033

Issue #781, opened 2016-04-19
Adding shortcut via wixFeatures results in duplicated primary key


smoes:
When trying to add a shortcut using `wixFeatures`I get the error from light:

    The primary key 'ApplicationProgramsFolderRemove' is duplicated in table 'RemoveFile'

Here is the important snippet from build.scala:

    // Adding the bat file explicitly as mentioned in the documentation
    mappings in Windows <<= (sourceDirectory in Windows) map { (dir) =>
      Seq((dir / "my.bat") -> "bin/my.bat")
    },

    // Adding the wanted wixFeature:
    wixFeatures += WindowsFeature(
      id = "shortcuts",
      title = "Shortcuts in start menu",
      desc = "Add shortcuts for execution and uninstall in start menu",
      components = Seq(
        AddShortCuts(Seq("bin/my.bat"))

      )
    )

Am I doing something wrong or is that a bug?
Using WiX toolchain v3.5 at the moment.

muuki88:
Sorry for the issues with windows. We have very few windows user, so this part of native-packager isn't as shiny as the other stuff. We already had some small fixes around these duplicated key issues, but looks like there still more.

Are you familiar with wix? If you are willing to make a pull request, we can surely help you to get things up and running. Starting point would be the [WixHelper.scala](https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/windows/WixHelper.scala) and the [WindowsPlugin](https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/windows/WindowsPlugin.scala)

smoes:
Thank you @muuki88 for the fast response.

I am not too familiar with WiX. However, it seems like this is a bug in the plugin itself. I suppose that there is always one `AddShortCuts` processed in `makeComponentInfo`, no matter if there is one explicitly added or not. This always appears in the corresponding .wix file as follows:

    <DirectoryRef Id="ApplicationProgramsFolder">
      <Component Id="shortcut_d074a9b9_045a_4f82_9363_7cc797820d6b177694507" Guid="aeb3fd7b-e20f-421d-ab49-0f82c48de7c0">
           
        <RemoveFolder Id="ApplicationProgramsFolderRemove" Directory="ApplicationProgramsFolder" On="uninstall"/>
[...]

PR #802, opened 2016-06-02
Introduce new namespace and solve duplicate key issue

smoes:
…4 by default but can be overridden by user.
Fixes issue #780 and allows using the newest Wix toolset.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/windows/WindowsPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/windows/WindowsPlugin.scala
@@ -57,8 +57,7 @@ object WindowsPlugin extends AutoPlugin {
       "-cultures:en-us"),
     wixProductId := WixHelper.makeGUID,
     wixProductUpgradeId := WixHelper.makeGUID,
-    wixMajorVersion := 4,
-
+    wixMajorVersion := 3,
     maintainer in Windows <<= maintainer,
     packageSummary in Windows <<= packageSummary,
     packageDescription in Windows <<= packageDescription,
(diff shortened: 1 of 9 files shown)


smoes:
The error is due to the standard major version being `4`, but it is using `3`. Please add `wixMajorVersion := 3` to settings.

src/main/scala/com/typesafe/sbt/packager/windows/WindowsPlugin.scala
```diff
@@ -57,6 +57,8 @@ object WindowsPlugin extends AutoPlugin {
       "-cultures:en-us"),
     wixProductId := WixHelper.makeGUID,
     wixProductUpgradeId := WixHelper.makeGUID,
+    wixMajorVersion := 4,
```
muuki88: Can you change this to `3`, so we remain backwards compatible and the tests run.

muuki88:
Code looks good. Please add some [documentation](http://sbt-native-packager.readthedocs.io/en/latest/formats/windows.html). You can build the docs [with this guide](https://github.com/sbt/sbt-native-packager/wiki/Developer-Guide#documentation)

Also an additional windows test would be nice, it seems appveyor only [supports Wix 3.10](https://www.appveyor.com/docs/installed-software#languages-libraries-frameworks).

smoes:
Now also resolves issue #781. 
I removed the auto start menu shortcut generation for config files, since that should be left to the user anyways. That resolves the conflict of two coexisting AddShortCut WindowsFeatures. Config files still can be added to the start menu manually.

smoes:
So the final solution for #781 was different than I expected:

The RemoveFolder directive has nothing to do with shortcuts. I re-inserted the previous standard shortcuts for the menu files but removed the RemoveFolder directive for ApplicationFolder from the AddShortcuts xml. It was moved to ApplicationFolder itself so that it appears only once in the very right spot, independent from shortcuts.

muuki88:
LGTM. I will try t o test it this week on my windows machine.

@nazoking do you have time to test this as well?

PR #806, opened 2016-06-11
Add: dockerAdditionalVersions

dgilling:
Useful if you want other version tags added to Docker Image in addition
to the project version. Examples include tagging by the git hash or git
branch using the sbt-git plugin

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -74,6 +74,7 @@ object DockerPlugin extends AutoPlugin {
     dockerExposedVolumes := Seq(),
     dockerRepository := None,
     dockerUpdateLatest := false,
+    dockerAdditionalVersions := Seq(),        
     dockerEntrypoint := Seq("bin/%s" format executableScriptName.value),
     dockerCmd := Seq(),
     dockerCommands := {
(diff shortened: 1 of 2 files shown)


typesafehub-validator:
Hi @dgilling,

muuki88:
Thanks for your pull request. Can you add documentation and tests for this feature?

If I understand this correctly, this will publish multiple docker images with different versions in the form of

`packageName:tag:version`

dgilling:
Yes that is correct.

For my workflow, I use it in combination with sbt-git to add a docker tag that is the branch name.
This allows me to reference an image by the normal version tag and also by :develop if for example the git branch was named “develop"

I will add some docs soon and get the paper signed.

Derric

muuki88:
Hi. I had some time to read through the docker stuff and native-packagers implementation.
IMHO a setting `dockerAdditionalVersions` is the wrong name and definition here. Currently  the tag under which name the image gets published is defined [here](https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala#L133-L135)

```scala
dockerTarget <<= (dockerRepository, packageName, version) map {
   (repo, name, version) => repo.map(_ + "/").getOrElse("") + name + ":" + version
}
```

The `publishLocal` method uses this `dockerTarget` ( incorrectly scoped to `in Docker` ) as the tag
```scala
publishLocal <<= (stage, dockerTarget, dockerUpdateLatest, streams) { ... }
```

This results in a docker build command like this
```
docker build --force-rm -t ${dockerTarget}
# dockerTarget expanded
docker build --force-rm -t [${dockerRepository}/]${packageName}:${version}
# actual command (if dockerRepository is none)
docker build --force-rm -t my-app:0.1.0
```

The thing `dockerAdditionalVersions` actually should do is add _more_ `-t` flags ( actually I would prefer using `--tag=`, which is easier to read.

### Proposal

Instead of introducing a very specific setting `dockerAdditionalVersions`, which runs the build multiple times. I would rather implement a generic `dockerBuildOptions: TaskSetting[Seq[String]]` and a more specific `dockerTags: TaskSetting[Seq[DockerTag]]` setting.

```scala
dockerBuildOptions := Seq("--force-rm"),
dockerBuildOptions += dockerTags.value.map { tag => s"--tag=${tag.asBuildOption}" }
dockerBuildOptions ++= if(dockerLatest.value) { /* add latest tag */ } else Seq.empty
...

def publishLocalDocker(context: File, options: Seq[String], log: Logger) {
   val cmd = Seq("docker", "build") ++ options ++ Seq(".")
   // ...
}
```

which would create a general abstraction for the build options. 

The `DockerTag` could look like this

```scala
case class DockerTag(name: String, version: Option[String]) {
   def asBuildOption: String = name + version.map(v => s":$v"}.getOrElse("")
}
```

@fiadliel WDYT?

dgilling:
Sorry about the delay.

Document is signed.
With reduced file size to avoid Email Reject.

muuki88:
Hi @dgilling . There is going to be another contribution, implementing the suggested solution above. I would close this pull request in favor of the more generic solution.

muuki88:
closing in favor of #854

Issue #812, opened 2016-06-15
RPM scriptles from file produce duplicate entries.


zoosky:
Here is a test project https://github.com/zoosky/scriptlets-override-rpm which produces the error:
https://github.com/zoosky/scriptlets-override-rpm

This is the SPEC file with duplicates, see %post and %postun
https://github.com/zoosky/scriptlets-override-rpm/blob/master/rpm-test.spec#L157

Testet with 
* Scala 2.11.8, 2.10.6
* sbt 0.13.11
* sbt-native-packager 1.1.1

muuki88:
Thanks for the sample project reproducing the error. Would you like to make a pull request with a failing test for this: https://github.com/sbt/sbt-native-packager/tree/master/src/sbt-test/rpm/scriptlets-override-rpm

The test only checks for existence, but not for uniqueness.

zoosky:
I'd like to. Have to find a time slot..

zoosky:
@muuki88 Unfortunately this fix not only removes the duplicates but adds also mor standard scripts into the spec file linke

```
startService() {
    app_name=$1
    if hash update-rc.d >/dev/null 2>&1; then
		echo "Adding $app_name to autostart using update-rc.d"
		update-rc.d $app_name defaults
		service $app_name start
    elif hash chkconfig >/dev/null 2>&1; then
		echo "Adding $app_name to autostart using chkconfig"
		chkconfig --add rpmpackage
		chkconfig $app_name on
		service $app_name start
    else
		echo "WARNING: Could not add $app_name to autostart: neither update-rc nor chkconfig found!"
    fi
}

#
# Removing service from autostart
# $1 = service name
#
stopService() {
    app_name=$1
    if hash update-rc.d >/dev/null 2>&1; then
	echo "Removing $app_name from autostart using update-rc.d"
	update-rc.d -f $app_name remove
	service $app_name stop
    elif hash chkconfig >/dev/null 2>&1; then
	echo "Removing $app_name from autostart using chkconfig"
	chkconfig $app_name off
	chkconfig --del $app_name
	service $app_name stop
    else
	echo "WARNING: Could not remove $app_name from autostart: neither update-rc nor chkconfig found!"
    fi

}

#
# Restarting the service after package upgrade
# $1 = service name
#
restartService() {
	app_name=$1
	service $app_name restart
}

# Scriptlet syntax: http://fedoraproject.org/wiki/Packaging:ScriptletSnippets#Syntax
[...]

PR #819, opened 2016-06-21
failing test for rpm mainainer scripts

zoosky:
Test:
> scripted rpm/scriptlets-override-rpm

as asked for in https://github.com/sbt/sbt-native-packager/issues/812


muuki88:
Brilliant! Generally a failing test is the best way to get a bug fixed in OSS :D

muuki88:
cheery-picked in #825

Issue #789, opened 2016-05-11
brp-java-repack-jars still running despite setting rpmBrpJavaRepackJars to false


jjunqueira:
Hi,

This is related to this http://stackoverflow.com/questions/36940733/sbt-native-packager-rpm-brp-java-repack-jars-still-running-despite-setting-rpmbr question.

I was asked to open an issue referencing this https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/rpm/RpmPlugin.scala#L112.

Let me know if I can provide any more information.

muuki88:
Awesome, thanks :) 

> Let me know if I can provide any more information

A pull request would be the only thing left to do ;)

YuvalItzchakov:
I've tried using both the mentioned workaround (http://stackoverflow.com/a/36966909/1870803) and using `rpmBrpJavaRepackJars := false` with sbt-native-packager 1.1.0-RC and 1.1.1. Both are being ignored doing rpm packaging and I still see `brp-java-repack-jars` being ran.

Am I missing something?

    lazy val rpmBuildSettings = Seq(
      rpmPrefix := Some(defaultLinuxInstallLocation.value),

      linuxPackageSymlinks := Seq.empty,
      defaultLinuxLogsLocation := defaultLinuxInstallLocation.value + "/" + name.value,

      rpmBrpJavaRepackJars in Rpm := false,
      rpmBrpJavaRepackJars := false,

      //64bit architecture
      packageArchitecture in Rpm := "x86_64",

      //name of the package will be as the project name (defined in project's build.sbt)
      packageName in Rpm := s"${name.value}",

      //Start of FAT jar configuration

      // we specify the name for our fat jar
      assemblyJarName := s"${name.value}_${version.value}.jar",
      // the bash scripts classpath only needs the fat jar
      scriptClasspath := Seq(assemblyJarName.value),
      
      // removes all jar mappings in universal and appends the fat jar
      mappings in Universal := {
        // universalMappings: Seq[(File,String)]
        val universalMappings = (mappings in Universal).value
        val fatJar = (assembly in Compile).value

        // removing means filtering
        val filtered = universalMappings filter {
          case (file, fileName) => !fileName.endsWith(".jar")
        }

        filtered :+ fatJar -> ("lib/" + fatJar.getName)
      },
      //End of FAT jar configuration

      //Takes the application.conf file from resources and puts it in conf folder during installation
      mappings in Universal += {

[...]

PR #826, opened 2016-06-26
Fix scoping in RPM plugin for #789

thetristan:
This change updates the RPM plugin settings so that keys starting with `rpm` are no longer scoped to the `Rpm` configuration. This should address #789.

I didn't see a formal testing process documented (though I may have missed it in the docs), so far I've only verified that `sbt test` and `sbt scripted rpm/*` are passing (both ran on a VM running centos7).

TODO 
- [x] Investigate if additional tests are needed, and if so add them

### DIFF
--- a/src/sbt-test/rpm/simple-rpm/build.sbt
+++ b/src/sbt-test/rpm/simple-rpm/build.sbt
@@ -1,3 +1,5 @@
+import com.typesafe.sbt.packager.linux._
+
 enablePlugins(RpmPlugin)
 
 name := "rpm-test"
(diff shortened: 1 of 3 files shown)


muuki88:
Thanks for your pull request. Code LGTM. The is a minimal [Contributing.md](https://github.com/sbt/sbt-native-packager/blob/master/CONTRIBUTING.md), which describes how to run the tests. We don't have a"formal test process", as the test-matrix for this project is almost infinite :(

There is a simple test `sbt-test/rpm/simple-rpm`, which only checks if the rpm has been built. It would be nice to extend this test to check if the `spec` file contains the correct contents. The `sbt-test/rpm/systemd-rpm` build.sbt has an example on how to check the spec file.

muuki88:
@thetristan do you need any help for extending the tests?

thetristan:
@muuki88 I should be OK. I've been busy with real life kind of things but I'll try to get this change wrapped up this weekend.

thetristan:
I updated the base test for `simple-rpm` to check the values for the existing keys in that project. I'll extend the test at some point later today to add coverage for the additional keys that aren't prefixed with `rpm*`, too.

src/sbt-test/rpm/simple-rpm/build.sbt
```diff
@@ -19,4 +19,29 @@ rpmUrl := Some("http://github.com/sbt/sbt-native-packager")
...
 
-
+TaskKey[Unit]("unzip") <<= (packageBin in Rpm, streams) map { (rpmFile, streams) =>
+  val rpmPath = Seq(rpmFile.getAbsolutePath)
+  Process("rpm2cpio" , rpmPath) #| Process("cpio -i --make-directories") ! streams.log
+}
+
+TaskKey[Unit]("checkSpecFile") <<= (target, streams) map { (target, out) =>
+  val spec = IO.read(target / "rpm" / "SPECS" / "rpm-test.spec")
+  println(spec)
```
muuki88: I see traces of debugging sbt :D

muuki88:
looks good so far :)

thetristan:
OK, I was finally able to add those additional checks. Let me know if you see anything else that should be improved.

thetristan:
At first glance it looks like the build failure from my last push is unrelated. I rebased my commits to re-trigger the build and will investigate further if it fails again.

muuki88:
Thanks a lot. Will release this ASAP.

thetristan:
Sure, you're welcome!

PR #839, opened 2016-07-09
WIP 633 Add prototype for multiple apps in single project

muuki88:
## Prototype

This is the first attempt to provide _multiple start scripts_ within a single project. The main idea is to leverage the already provided script by `JavaAppPackaging` and just forward calls to it. This results in
- the default start script (using the `mainClass in Compile` defined class`=
- scripts for all `discoveredMainClasses in Compile)
## Usage

Assuming two main classes `com.example.DefaultMain` and `com.example.AnotherMain`.

``` scala
enablePlugins(JavaAppPackaging, ScriptPerMainClass)

mainClass in Compile := Some("com.example.DefaultMain")
```

`sbt stage` will create the following output

```
target/universal/stage
├── bin
│   ├── another-main
│   ├── default-main
│   ├── test-project-simple
│   └── test-project-simple.bat
├── conf
│   └── application.ini
└── lib
    ├── com.typesafe.config-1.2.1.jar
    ├── default.test-project-simple-0.2.0.jar
    └── org.scala-lang.scala-library-2.10.4.jar
```
## TODOs
- [x] Tests need to be fixed and/or extended
- [x] Documentation is missing
- [x] Bat script plugin is missing
- [x] Application.ini refactoring still needs to be done


muuki88:
comments welcome: @metasim , @wookietreiber , @kardapoltsev

wookietreiber:
First of all thank you for considering this use case and providing an implementation for it. I have a few questions regarding your code samples:

1.  What is the purpose of `bin/test-project-simple[.bat]`?
2.  Why do you set `mainClass in Compile`? When scripts are created for all detected main classes, setting this should be superfluous.

muuki88:
> What is the purpose of bin/test-project-simple[.bat]?

This is generated by the `JavaAppPackaging`. This initial version won't support windows, so there is only one bat file.

> Why do you set mainClass in Compile? When scripts are created for all detected main classes, setting this should be superfluous.

I though of two options. 

1. Generate the `bash-template` script for each main class, replicating the template and just alter the main class
2. Generate small scripts that call the default start script and only override the `-main` parameter.

I went for the second option, which I thought reduces duplicated bash scripts. On the other hand this generates the confusing default script. Hm.

wookietreiber:
> 1.  Generate the bash-template script for each main class, replicating the template and just alter the main class
> 2.  Generate small scripts that call the default start script and only override the -main parameter.

I would prefer the first option.

metasim:
@muuki88 No particular thoughts. I'm glad it's implemented via a plugin, and that separate launch scripts are created. `javapackager` has the ability to deploy two launchers, so long term would be interesting to see if it could automatically leverage this.

One question: how do you control the number of launcher scripts that are generated? Is there a task you can override to filter the found  `main` classes?

muuki88:
Thanks a lot for the input. 

@wookietreiber you are right. These are both valid use cases. I think I would implement them both in one plugin and switch between them via a setting. 

> how do you control the number of launcher scripts that are generated? Is there a task you can override to filter the found main classes?

@metasim not yet. I was not sure if this should be added in the first version as "too many" start scripts won't hurt and you can always filter them via the `mappings in Universal`. If I add this, I thought of something like `mainClassToScriptName = settingKey[String => Option[String]]`.

metasim:
>  I was not sure if this should be added in the first version as "too many" start scripts won't hurt and you can always filter them via the mappings in Universal. If I add this, I thought of something like `mainClassToScriptName = settingKey[String => Option[String]]`.

I like that, because it allows a) name transformation, **and** b) filtering.

Are `Class` names really "stringly typed" in sbt?

muuki88:
@metasim . Yes, `Class`es are "stringly typed".

muuki88:
@kardapoltsev / @metasim could you take a look? If you think the behaviour is okay, I'll merge this and release. You can use the github review feature to do so :)

src/sphinx/archetypes/java_app/index.rst
```diff
@@ -147,6 +147,67 @@ application arguments with ``--``. For example
...
 
+Multiple Applications
+=====================
+
+If you have multiple main classes then the ``JavaAppPackaging`` archetype provides you with two different ways of
+generating start scripts.
+
+1. A start script for each entry point. This is the default behaviour, when no ``mainClass in Compile`` is set
+2. One start script for the defined ``mainClass in Compile`` and forwarding scripts for all other main classes.
+
```
metasim: What is meant by "forwarding scripts"? Maybe flesh out the semantic difference, and/or indicate details follow.

src/sphinx/archetypes/java_app/index.rst
```diff
@@ -147,6 +147,67 @@ application arguments with ``--``. For example
...
+   bin/
+     bar-main
+     bar-main.bat
+     my-project
+     my-project.bat
+
+
+Now you can package your application as usual, but with multiple start scripts.
+
```
metasim: Here I might go into a bit more detail on the differences between the main start script and what it means to have a "forwarder". Also, how do you disable the generation of forwarders?

src/sphinx/archetypes/java_app/index.rst
muuki88: True. I'm still not happy with the naming anyway. _Forwarder_, _Delegate_,..
However I will define the semantics more clearly.

Issue #838, opened 2016-07-08
docker:publishLocal results in error with Docker 1.12.0-rc3


maciej:
Calling `docker:publishLocal` on a project from SBT results in:

```
[info] Built image [REDACTED]/[REDACTED]:prod
unknown shorthand flag: 'f' in -f
See 'docker tag --help'.
```

I'm using sbt-native-packager 1.1.2, latest OSX / MacOS and _Docker for Mac_.

```
✗ docker --version
Docker version 1.12.0-rc3, build 91e29e8, experimental
```

nlhkh:
I am seeing the same issue. Must be that the packager did not accommodate for update in Docker command.

nlhkh:
[818](https://github.com/sbt/sbt-native-packager/issues/818)

maciej:
I've created a backport of the PR to the `1.1.x` branch: #845.

@muuki88 would it be possible to get it accepted in the 1.1.x branch and released? :-)

muuki88:
@maciej perfect. I will take a look and get a release out ASAP.

maciej:
@muuki88 awesome, thank you!

PR #845, opened 2016-07-12
docker: tag -f is deprecated (backport to 1.1.x)

### LINKED ISSUE #818: docker: tag -f is deprecated
As of docker 1.10.1 `-f flag on docker tag` is deprecated and removed in 1.2.0:

```scala
val latestCmd = Seq("docker", "tag", "-f", tag, name)
```

https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala#L284

They explain the motivation for removing it as follows:
>To make tagging consistent across the various docker commands, the -f flag on the docker tag command is deprecated. It is not longer necessary to specify -f to move a tag from one image to another. Nor will docker generate an error if the -f flag is missing and the specified tag is already in use.

So, can we please remove that usage?

maciej:
fixes #818, #838 in the 1.1.x branch


muuki88:
Released in `v1.1.3`. Thanks for the backport :)

PR #854, opened 2016-07-26
Adding 'dockerBuild{Command, Options}' and renaming 'dockerTag'

makubi:
* dockerBuildCommand
The 'dockerBuildCommand' let users customize the command for building
Docker images.
The current default is 'docker build [dockerBuildOptions] .'.

* dockerBuildOptions
Furthermore 'dockerBuildOptions' defines which build options are used.
The current default is '--force-rm -t [dockerAlias]' (which is expanded
if 'dockerUpdateLatest' is set to 'true'.

This results in an execution of 'docker build --force-rm -t
[dockerAlias] .' (if 'dockerUpdateLatest' is 'false').

* dockerTag being renamed to dockerAlias
The 'dockerTag' setting is renamed to 'dockerAlias' as this better
reflects the setting. Docker aliases consist of a registry path,
username, name and tag ([REGISTRY_HOST/][USERNAME/]NAME[:TAG]).
To not confuse the Docker image tag with this setting (as it takes the
full alias name), the setting is renamed.


src/main/scala/com/typesafe/sbt/packager/docker/DockerAlias.scala
```diff
@@ -0,0 +1,8 @@
+package com.typesafe.sbt.packager.docker
+
+case class DockerAlias(registryHost: Option[String], username: Option[String], name: String, tag: Option[String]) {
```
muuki88: A little documentation for this class and the public fields would be helpful.

muuki88:
LGTM. @fiadliel do you have any objections?

src/main/scala/com/typesafe/sbt/packager/docker/DockerAlias.scala
makubi: Done

nafg:
BTW github releases page, release notes for I think 1.2.0-M5, says "*dockerTarget* renamed"

PR #866, opened 2016-08-18
added addApp and addJava to ash template like in bash template 

### LINKED ISSUE #862: ash-template doesn't have the addApp function like in bash template
This creates some pain cause the configuration flags can't be passed to the application when using an ash template.

botany:
Fixes #862 
I have only tested manually to generate tested in BusyBox v1.22.1.
If any tests needed for this functionality, please guide me how to do them.

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/ash-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/ash-template
@@ -1,4 +1,6 @@
 #!/bin/sh
+java_args=""
+app_commands=""
 
 realpath () {
 (
@@ -27,11 +29,20 @@ realpath () {
 )
 }
 
+addJava() {
+  echo "[addJava] arg = '$1'"
+  java_args="$java_args $1"
+}
+
+addApp() {
+  echo "[addApp] arg = '$1'"
+  app_commands="$app_commands $1"
+}
+
 # Loads a configuration file full of default command line options for this script.
 loadConfigFile() {
   cat "$1" | sed '/^\#/d' | sed 's/^-J-X/-X/' | tr '\r\n' ' '
 }
-
 
 real_script_path="$(realpath "$0")"
 app_home="$(realpath "$(dirname "$real_script_path")")"
@@ -42,4 +53,4 @@ ${{template_declares}}
 # If a configuration file exist, read the contents to $opts
 [ -f "$script_conf_file" ] && opts=$(loadConfigFile "$script_conf_file")
 
-exec java -classpath $app_classpath $opts $app_mainclass $@
+exec java -classpath $app_classpath $opts $java_args $app_mainclass $app_commands $@


typesafehub-validator:
Hi @botany,

muuki88:
Hi Botany,

Thanks a lot for trying to implement this :)

How do you use the newly introduced functions? I guess you are using the `bashScriptExtraDefines`?

The `$java_args` configure the same thing as the `$opts` do, right? Not sure if we need two ways of configuring java options.

|  javaOptions in Universal | bashScriptExtraDefines |
| --------------------------------- | --------------------------------- |
| `javaOptions in Universal += "-Xmx2g"` | `bashScriptExtraDefines += "addJava -Xmx2g"` |

I would rather prefer the first option rather then supporting two if there's no good reason.

The `$app_commands` ( maybe `app_args` are a better name? ) may be different. There is the `$@` rest parameter for the start command, but if you don't have any control on how your image is started this doesn't help. And again, would you add them via `bashScriptExtraDefines` ?


Regarding the tests, we currently don't have any on our CI running, but please add one that shows things are working in the [sbt-test/ash](https://github.com/sbt/sbt-native-packager/tree/master/src/sbt-test/ash) directory.

We may also start documenting the features supported by the AshScript [here](https://github.com/sbt/sbt-native-packager/blob/master/src/sphinx/archetypes/misc_archetypes.rst).

The [Developer Guide](https://github.com/sbt/sbt-native-packager/wiki/Developer-Guide) describes how to build the documentation and run the tests.

botany:
Hi Nepomuk,

Yep, I'm adding them via `bashScriptExtraDefines`. 
When I first faced the problem I went through the docs here: 
[](http://www.scala-sbt.org/sbt-native-packager/archetypes/java_app/customize.html?highlight=configuration)

which explicitly support adding configuration properties via different options. 
There're actually not 2 but 3 ways to configure this: 
- `application.ini` file
- `javaOptions in Universal`
- `bashScriptExtraDefines `

Unfortunately the way it works for the bash script isn't the way it works for the ash script. 
The current version of the ash script supports only `application.ini ` and only for java arguments. 
If you look at the way it's configured here(including naming and args passing approach): 
[](
https://github.com/sbt/sbt-native-packager/blob/97d315e613caa87668c503cdf9248264949612a1/src/main/resources/com/typesafe/sbt/packager/archetypes/bash-template#L249)

You will see that all I did is moved some parts into the ash script and fixed some minor syntax things as sh and bash aren't the same. 

To my humble opinion it's too late to change this confusing behavior for the bash template cause breaking things for existing users is not good. On the other hand the ash template behavior should probably be similar to the bash template cause essentially they are the same. 

Regarding tests, I will take a look, tnx.

muuki88:
> Unfortunately the way it works for the bash script isn't the way it works for the ash script. 
The current version of the ash script supports only `application.ini` and only for java arguments.

My hope was to keep only the relevant stuff in the _ash_ script and not build another big ball of bash. However I see your point. Switching the start script template should not change the behavior. A feature may not be available, but it never should behave differently.

> To my humble opinion it's too late to change this confusing behavior for the bash template cause breaking things for existing users is not good

That's a valid point. If a clean up happens, we should perform it for every usage.

In short: Code looks okay. Test and docs should be added :)

muuki88:
Can you take a look at my comment in #878 . Maybe we can fix this along with your commit.

muuki88:
Close in favor of #942

PR #881, opened 2016-09-19
Add ability to expose UDP ports

### LINKED ISSUE #843: Support UDP ports for dockerExposedPorts
`dockerExposedPorts` takes a Seq of Int, which makes it impossible to specify transport (Docker supports TCP and UDP).

I wish to add `EXPOSE 12345/udp` to my Dockerfile

NeQuissimus:
Add `dockerExposedUdpPorts` similar to the existing `dockerExposedPorts`.

Note that this depends on #880 for the tests to pass.

Fixed #843

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -74,6 +74,7 @@ object DockerPlugin extends AutoPlugin {
   override lazy val projectSettings = Seq(
     dockerBaseImage := "openjdk:latest",
     dockerExposedPorts := Seq(),
+    dockerExposedUdpPorts := Seq(),
     dockerExposedVolumes := Seq(),
     dockerRepository := None,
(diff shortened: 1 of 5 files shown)


src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -207,8 +208,8 @@ object DockerPlugin extends AutoPlugin {
    * @param exposedPorts
    * @return if ports are exposed the EXPOSE command
    */
-  private final def makeExposePorts(exposedPorts: Seq[Int]): Option[CmdLike] = {
-    if (exposedPorts.isEmpty) None else Some(Cmd("EXPOSE", exposedPorts mkString " "))
+  private final def makeExposePorts(exposedPorts: Seq[Int], exposedUdpPorts: Seq[Int]): Option[CmdLike] = {
```
NeQuissimus: Do we care? I think most of the methods in this class could have a more limited scope.

muuki88: Nope. I always forget to configure this :/

PR #888, opened 2016-10-06
Generate EXPOSE when only UDP ports are exposed

lustefaniak:
Right now setting
```
dockerExposedPorts := Seq()
dockerExposedUdpPorts := Seq(10000, 10001)
```

would not generate EXPOSE section


muuki88:
Perfect! Thanks for taking care. I try to release this ASAP.

PR #889, opened 2016-10-06
Update documentation with warning about overriding default tar options.

mackler:
Includes an example (that seems to be working for me).

I'm completely unfamiliar with Sphinx neither an Sbt expert, so please forgive me if this is wrong in either substance or format.

### DIFF
--- a/src/sphinx/formats/universal.rst
+++ b/src/sphinx/formats/universal.rst
@@ -158,6 +158,12 @@ If you want to force local for the `tgz` output add this line:
   universalArchiveOptions in (Universal, packageZipTarball) := Seq("--force-local", "-pcvf")
 
 This will set the cli options for the `packageZipTarball` task in the `Universal` plugin to use the options ``--force-local`` and ``pcvf``.
+Be aware that the above line will overwrite the default options.  You may want to prepend your options, doing something like:
+
+.. code-block:: scala
+
+  universalArchiveOptions in (Universal, packageZipTarball) :=
+    (Seq("--exclude", "*~") ++ (universalArchiveOptions in (Universal, packageZipTarball)).value)
 
 Currently, these task can be customized:
 


src/sphinx/formats/universal.rst
```diff
@@ -158,6 +158,12 @@ If you want to force local for the `tgz` output add this line:
   universalArchiveOptions in (Universal, packageZipTarball) := Seq("--force-local", "-pcvf")
 
 This will set the cli options for the `packageZipTarball` task in the `Universal` plugin to use the options ``--force-local`` and ``pcvf``.
+Be aware that the above line will overwrite the default options.  You may want to prepend your options, doing something like:
+
+.. code-block:: scala
+
+  universalArchiveOptions in (Universal, packageZipTarball) :=
```
muuki88: Thanks for clarifying this. It would be nicer add this as a second example as this is more a sbt operator confusion then a native-packager documentation issue. The `:=` operator overrides an existing setting whereas `+=` / `++=` appends something to an existing setting that is a `Seq`. So you change the example to

```scala
// override settings
universalArchiveOptions in (Universal, packageZipTarball) := Seq("--force-local", "-pvcf")
// addditional settings
universalArchiveOptions in (Universal, packageZipTarball) ++= Seq("--exclude", "*~"
```

mackler: Would the second one work?  I am under the impression that extra options to `tar` need to be _prepended_ rather than appended, because the default switches include `-c` which must be followed by the tar-file name.  But I believe the appended options will be placed _before_ the tar-file name.  Am I wrong about that?

muuki88: Hm. Not sure. In your description you overrode the complete settings
```
universalArchiveOptions in (Universal, packageZipTarball) := Seq("--exclude", "*~")
```

which caused an error. Can you try if appending works as well? I guess since these are all options the order shouldn't matter.

mackler: I tried the append as

    universalArchiveOptions in (Universal, packageZipTarball) ++= Seq("--exclude", "*~")

and the resulting output included:

    Running with tar -pcvf --exclude *~ /tmp/sbt_72102b69/myApp-0.1.tar myApp-0.1
    tar: *~: Cannot stat: No such file or directory
    tar: /tmp/sbt_72102b69/myApp-0.1.tar: Cannot stat: No such file or directory

and then later:

    gzip: can't stat: /tmp/sbt_72102b69/myApp-0.1.tar: No such file or directory
    java.lang.RuntimeException: Error gziping /tmp/sbt_72102b69/myApp-0.1.tar. Exit code: 1
    [error] (server/universal:packageZipTarball) Error gziping /tmp/sbt_72102b69/myApp-0.1.tar. Exit code: 1

Correct me if I'm wrong, but it looks like the generated command told `tar` to name the tar-file `--exclude`, and that files named `*~` and `/tmp/sbt_72102b69/myApp-0.1.tar` are to be included in the archive.

Again, the default options seem to depend on nothing coming between the `-f` switch and the tar-file name, which is what appending--but not prepending--custom options to the defaults does.

muuki88: Ah, I see. Thanks for checking all this, currently I don't have much time, even for the small things.

PR #890, opened 2016-10-06
Updated cheatsheet with the correct file path for overriding service manager scripts

mitch-seymour:


muuki88:
Thanks for the fast pr :)

PR #895, opened 2016-10-12
[rpm] Fixes #894. Symlinks only removed during uninstall, not during update of RPM.

### LINKED ISSUE #894: During RPM (yum) upgrade, config gets removed (1.2.0-M6)
When updating a RPM with a newer version, the config files (e.g. `/etc/$app/`) get removed. Looking at the generated spec file, there are several `rm -rf` commands in the `%postun` scriptlet. I would expect them to reside within the preceeding `if [ $1 -ge 1 ] ` block, where it is checked, whether an upgrade or an uninstall is performed.

kodemaniak:
The removal of the symlinks in `%preun` now is only triggered during an uninstall. Before an update of an package removed the symlinks and left the package in a non-functioning state.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/rpm/RpmMetadata.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/rpm/RpmMetadata.scala
@@ -312,13 +312,14 @@ case class RpmSpec(meta: RpmMetadata,
     if (symlinks.isEmpty)
       None
     else {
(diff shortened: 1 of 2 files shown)


muuki88:
I'm out of town ATM so I can't release this. I will test and merge when I'm back in three weeks.

kodemaniak:
Great, I'll use my own build until then.

muuki88:
@kardapoltsev if you approve, feel free to merge.

PR #896, opened 2016-10-13
RpmNoReplaceplugin and LinuxMappingDSL for "noreplace" configs

### LINKED ISSUE #572: Marking default config files as noreplace in RPMs
Currently, the config files get replaced during an rpm update. However, at least for production deployments it is very likely that the default configs have been modified on disk, so that replacing them with the defaults during an update will actually make the services not run correctly (which might even result in serious problems obviously). For rpms there is the option to mark files as "config(noreplace)" in order to tell rpm not to update the files on disk automatically, but to copy the default config (if it has changed) to a new file called <config>.rpmnew. The RPM plugin already supports this, only the default mappings need to be changed AFAICT.

see also: https://groups.google.com/forum/#!topic/sbt-native-packager/SjdgGI1deZk

kodemaniak:
Added a RpmConfigNoReplacePlugin that depends on the Rpm and ServerAppPackaging plugins. It marks all config files as "noreplace" in the RPM, thus preventing them from being replaced during updates. The main method was added to the LinuxMappingDSL and can also be used in build.sbt for non server projects.

Fixes #572.


src/main/scala/com/typesafe/sbt/packager/linux/RpmNoReplaceConfigPlugin.scala
```diff
@@ -0,0 +1,18 @@
...
+import com.typesafe.sbt.packager.archetypes.JavaServerAppPackaging
+import com.typesafe.sbt.packager.rpm.RpmPlugin
+import com.typesafe.sbt.packager.rpm.RpmPlugin.autoImport
+import sbt._
+
+/**
+  * This plugin automatically updates all config files in a RPM to "noreplace". Those files will not be overwritten
+  * during an update if they have been changed on disk. This is useful for server apps.
+  */
+object RpmNoReplaceConfigPlugin extends AutoPlugin with LinuxKeys with LinuxMappingDSL {
```
muuki88: I'm still not convinced of my initial idea implementing this as an autoplugin. It only makes sense when enabled by default and for debian as well. 

This will break backwards compatibility, which I usually prefer to avoid. If you can make a good case ( maybe some best practices, blog posts on this issue? ) I'm fine with adding this.

From a code perspective:

- instead of extending the `LinuxKeys` use the "global" [Keys](https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/Keys.scala) object

src/test/scala/com/typesafe/sbt/packager/linux/LinuxMappingDSLSpec.scala
```diff
@@ -0,0 +1,26 @@
+package com.typesafe.sbt.packager.linux
+
+import java.io.File
+
+import org.scalatest.{Matchers, WordSpec}
+
+/**
+  * Created by carsten on 13.10.16.
+  */
+class LinuxMappingDSLSpec
```
muuki88: Awesome. Have you run `sbt test:scalafmt` on this?

src/main/scala/com/typesafe/sbt/packager/rpm/RpmMetadata.scala
```diff
@@ -312,13 +312,14 @@ case class RpmSpec(meta: RpmMetadata,
     if (symlinks.isEmpty)
       None
     else {
+      val checkUninstall = "if [ $1 -eq 0 ] ;\nthen"
```
muuki88: This change is from #894 . Can you remove this from the PR when #894 is merged?

src/test/scala/com/typesafe/sbt/packager/linux/LinuxMappingDSLSpec.scala
kodemaniak: Now, yes :)

src/main/scala/com/typesafe/sbt/packager/rpm/RpmMetadata.scala
kodemaniak: I removed the commit from the pull request.

src/test/scala/com/typesafe/sbt/packager/linux/LinuxMappingDSLSpec.scala
muuki88: Nice :)

src/main/scala/com/typesafe/sbt/packager/linux/RpmNoReplaceConfigPlugin.scala
kodemaniak: So, I remove the plugin for now. And I'll add some docs then.

muuki88: Yep. When you have added some docs this is ready to be merged. Thanks for your time and patience :)

kodemaniak:
I removed the plugin and added some docs. Please let me know if any further changes are required.

kodemaniak:
I just realized that the `/etc/default/<package-name>` mapping is still not set to noreplace by using the DSL. Is that one somehow added in the end? My assumption was that the auto plugins are executed before any modifications are made from the `build.sbt`.

muuki88:
Ah. This is as scoping issue. The [server archetype](https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaServerApplication.scala) scopes the etc default template to `Rpm` and `Debian`. To fix this you need to use the correct scoped `linuxPackageMappings`.

```scala
linuxPackageMappings in Rpm := ...
```

muuki88:
Ah... hmm.. updating adds a merge commit. Let's see how that plays out when using the "squash & merge" button.

kodemaniak:
I finally had the time to try it out, but when I scope the linuxPackageMappings to Rpm, then `/etc/default/<package-name>` is completely gone.

muuki88:
I can confirm this ( for `rpm` ATM ). It seems a lot of things are missing when `linuxPackageMappings in Rpm := configWithNoReplace(linuxPackageMappings.value)` is added.

The `spec` file without the configuration` in the `test-project-simple`.

```
%files
%attr(0755,root,root) /usr/share/test-project-simple/bin/test-project-simple
%config %attr(0644,root,root) /usr/share/test-project-simple/conf/application.ini
%attr(0644,root,root) /usr/share/test-project-simple/lib/default.test-project-simple-0.2.0.jar
%attr(0644,root,root) /usr/share/test-project-simple/lib/org.scala-lang.scala-library-2.10.6.jar
%attr(0644,root,root) /usr/share/test-project-simple/lib/com.typesafe.config-1.2.1.jar
%dir %attr(755,test-project-simple,test-project-simple) /var/log/test-project-simple
%config %attr(644,root,root) /etc/default/test-project-simple
%dir %attr(755,test-project-simple,test-project-simple) /var/run/test-project-simple
%config %attr(0644,root,root) /usr/lib/systemd/system/test-project-simple.service
```

And this is with the configWithNoReplace

```
%attr(0755,root,root) /usr/share/test-project-simple/bin/test-project-simple
%config(noreplace) %attr(0644,root,root) /usr/share/test-project-simple/conf/application.ini
%attr(0644,root,root) /usr/share/test-project-simple/lib/default.test-project-simple-0.2.0.jar
%attr(0644,root,root) /usr/share/test-project-simple/lib/org.scala-lang.scala-library-2.10.6.jar
%attr(0644,root,root) /usr/share/test-project-simple/lib/com.typesafe.config-1.2.1.jar
%dir %attr(755,test-project-simple,test-project-simple) /var/log/test-project-simple
```

muuki88:
Sorry. This is again, a configuration issue.

If you replace the `linuxPackageMappings in Rpm` you must use same as an input.

```scala
linuxPackageMappings in Rpm := configWithNoReplace((linuxPackageMappings in Rpm).value)
```

Would you mind making a small PR and add this to the docs?

kodemaniak:
Just fixed the docs:

https://github.com/sbt/sbt-native-packager/pull/920

PR #897, opened 2016-10-17
fix a bug in the ash-template

stew:
`[ "foo" == "bar" ]` is not valid posix. It should be a single =, or else you'll get "unexpected operator/operand" on a posix compliant shell:

    > posh
    $ [ "foo" = "foo" ] && echo yes 
    yes
    $ [ "foo" == "foo" ] && echo yes
    posh: [: ==: unexpected operator/operand

    > dash                                                                                                               ⏎
    $ [ "foo" = "foo" ] && echo yes 
    yes
    $ [ "foo" == "foo" ] && echo yes
    dash: 2: [: foo: unexpected operator


typesafehub-validator:
Hi @stew,

PR #898, opened 2016-10-19
add RUN_OPTS support for systemd

kardapoltsev:
RUN_OPTS could be specified in `/etc/defalt/app_name` like in a `SystemV` setup


src/main/resources/com/typesafe/sbt/packager/archetypes/systemloader/systemd/start-template
```diff
@@ -6,7 +6,7 @@ Requires=${{start_facilities}}
 Type=simple
 WorkingDirectory=${{chdir}}
 EnvironmentFile=${{env_config}}
-ExecStart=${{chdir}}/bin/${{exec}}
+ExecStart=${{chdir}}/bin/${{exec}} $RUN_OPTS
```
muuki88: This should already work. The `EnvironmentFile` above provides the ability to add environment variables.

In systemv this always felt like a hack.

muuki88: I took a look at the code. Currently only the debian script supports this. Using global variables for application startup seems insecure and magic. 

What is the problem you are trying to solve?

kardapoltsev: I'm trying to pass command line arguments to my app.

kardapoltsev: > I'd rather remove the RUN_OPTS support from debian and improve the application configuration documentation
Let's go this way (:

muuki88:
Okay cool :) Can I help you with your specific problem?

Issue #836, opened 2016-07-06
Disable symbolic link removal on update


jneko33:
I wish to retain symbolic links made during install, on update.

During a package upgrade process, because the new package gets installed before the old package gets uninstalled, I've found that symlinks made by the post script of the new package are then deleted by the postuninstall script of the old package (as the symlinks are the same in both packages). 
This behaviour could not be overwritten because in `com.typesafe.sbt.packager.rpm.RpmMetadata.writeSpec(rpmRoot: File, tmpRoot: File) [line 261]`, teardown of symlinks is always appended to the end of the postuninstall script as long as the `linuxPackageSymlinks` key is populated.

It would be useful if tearing down of symlinks to somehow be conditional on update, preferably to have update behaviour configurable.

dpennell:
I have the same issue.  While it looks like it is fairly common to use %post to add symbolic links, there seems to be pretty consistent advice to create them in the %install section.  Here are some references:
- http://lists.rpm.org/pipermail/rpm-list/2014-October/001646.html
- https://www.redhat.com/archives/rpm-list/2005-July/msg00127.html
- http://stackoverflow.com/questions/7521980/packaging-symlinks-via-rpmbuild

I assume that if you list the links in %files, then upgrade can "do the right thing".

Another option that I considered is to use the argument that RPM passes to %postun as described in http://www.rpm.org/max-rpm/s1-rpm-inside-scripts.html.  This would avoid removing all symbolic links, so links that are not present in newer versions would not get deleted.

muuki88:
Thanks @jneko33 for the detailed bug report and thanks @dpennell for the research. I tried a naive approach and putting the symlinks into the install section. I haven't tested this yet, just want to get something out quick to get feedback.

PR #901, opened 2016-10-28
Treat symlinks as normal files in spec file

dpennell:
Symlinks are now treated like normal files in the spec.  All of the special handling in the scriptlets has been removed.  Symlinks are now handled properly by upgrades.

### DIFF
--- a/src/sbt-test/rpm/symlink-rpm/build.sbt
+++ b/src/sbt-test/rpm/symlink-rpm/build.sbt
@@ -1,0 +1,32 @@
+import com.typesafe.sbt.packager.linux.{LinuxPackageMapping, LinuxSymlink}
+
+enablePlugins(JavaAppPackaging)
+
+name := "rpm-test"
+version := "0.1.0"
+maintainer := "David Pennell <dpennell@good-cloud.com>"
+
+packageSummary := "Test rpm package"
+packageName in Linux := "rpm-package"
+packageDescription := """A fun package description of our software,
+  with multiple lines."""
+
+rpmRelease := "1"
+rpmVendor := "typesafe"
+rpmUrl := Some("http://github.com/sbt/sbt-native-packager")
+rpmLicense := Some("BSD")
+
+linuxPackageSymlinks := {
+  val helloSymlink = LinuxSymlink(((file(defaultLinuxInstallLocation.value) / (packageName in Linux).value / "lib") / "hello.link").toString, "/fake/hello.txt")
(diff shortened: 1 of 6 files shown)


muuki88:
Thanks 😊 I'll take a closer when I come home tomorrow

muuki88:
cc @fsat can you take a look? Would be nice 😊

muuki88:
@dpennell I would first merge #895 and release this. Sorry for the short answer above, I'm not back from vacation and can look into this in more detail.

@fsat implemented a great part of the initial relocation implementation, so it would be nice if he could validate your changes. Maybe he also tried this approach first, but didn't worked out for him.

From my perspective I like the implementation. For two reasons
1. Removing bash code ( always good )
2. Staging directory looks more like the final system installation

dpennell:
I am certainly not an RPM expert.  I based my solution on what appeared to be well informed advice (see comments in #836) to just treat symlinks as regular files (which they are).  I definitely think this needs a serious review.  I've tested on my centos VM and it is being used in our product (currently in QA).

@muuki88 I *think* that this solution would remove the need for #895.

muuki88:
Yes, I hope this will remove the need for #895.

I will test this my self, but I also would like to wait for @fsat review on this.
@huntc Conductr has used some of the relocation features as well, right?

src/sbt-test/rpm/symlink-rpm/build.sbt
```diff
@@ -0,0 +1,32 @@
...
+linuxPackageSymlinks := {
+  val helloSymlink = LinuxSymlink(((file(defaultLinuxInstallLocation.value) / (packageName in Linux).value / "lib") / "hello.link").toString, "/fake/hello.txt")
+  Seq(helloSymlink)
+}
+
+TaskKey[Unit]("check-spec-file") <<= (target, streams) map { (target, out) =>
+  val spec = IO.read(target / "rpm" / "SPECS" / "rpm-package.spec")
+  out.log.success(spec)
+  assert(spec contains "%attr(0644,root,root) /usr/share/rpm-package/lib/rpm-test.rpm-test-0.1.0.jar", "Wrong installation path\n" + spec)
+  assert(spec contains "/usr/share/rpm-package/lib/hello.link", "Missing or incorrect symbolic link\n" + spec)
```
muuki88: Wrong links appear in the spec like this?

muuki88:
Closing this for #916 as I can't update this PR.

PR #924, opened 2016-12-02
Stop JavaAppPackaging from auto enabling the DockerPlugin

fchaillou:
When enabling the JavaAppPackaging, the DockerPlugin is automatically enabled.
It causes issue in our case and I don't think it makes generally sense.
It feels the user should knowlingly enable it if he needs it.
What do you think ?

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaApp.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaApp.scala
@@ -39,7 +39,7 @@ object JavaAppPackaging extends AutoPlugin {
   import JavaAppPackaging.autoImport._
 
   override def requires =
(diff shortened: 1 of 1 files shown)


typesafehub-validator:
Hi @fchaillou,

fchaillou:
FYI, i signed the lightbend CLA

muuki88:
We try to give the user a "batteries included" experience. I know there are at least two other sbt docker plugins out there, but I would like to keep the experience as is.

What I rather hope for are pull request to add the missing features that the other plugins have 😉

I know that one of the biggest issues with native-packager docker images is the size. There have been at least two attempts to split these. One was a proof of concept by me and another one was a "Library Plugin" that extracts all dependencies ( however there was never a pull request (

PR #925, opened 2016-12-04
Use stage  task instead of debianExplodedPackage

muuki88:
The `DebianPlugin` now uses the `stage` task to generate a staged folder where the `dpkg` build process is executed.

- Deprecated `debianExplodePackage`
- Replaced deprecated symbolic operators
- Fixed IntelliJ warnings


muuki88:
AppVeyor seems to have forgotten about sbt.

PR #931, opened 2017-01-16
Wip/remove symbolic operators

muuki88:


muuki88:
@eed3si9n / @dwijnand  could one of you have a short look if the migration looks fine so far?

The only thing I stumbled upon was  the `??` operator. You have an insights on this one?

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/BashStartScriptPlugin.scala
```diff
@@ -55,8 +55,8 @@ object BashStartScriptPlugin extends AutoPlugin {
     bashScriptDefines ++= bashScriptExtraDefines.value,
     bashScriptReplacements := generateScriptReplacements(bashScriptDefines.value),
     // Create a bashConfigLocation if options are set in build.sbt
-    bashScriptConfigLocation <<= bashScriptConfigLocation ?? Some(appIniLocation),
-    bashScriptEnvConfigLocation <<= bashScriptEnvConfigLocation ?? None,
+    bashScriptConfigLocation := (bashScriptConfigLocation ?? Some(appIniLocation)).value,
```
muuki88: I can't replace the `??` by simply using

```scala
bashScriptConfigLocation := Some(appIniLocation)
```

eed3si9n: What's the warning on that?

eed3si9n:
I only glanced at the changes, but they look ok. The tricky bits that the compiler might not catch is input tasks with `Unit`, but you don't have those keys except for in the scripted tests.

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/BashStartScriptPlugin.scala
muuki88: I haven't seen any and haven't found anything in the migration guide.
However I haven't seen this operator anywhere else yet, so I was a bit curious :)

muuki88:
Thanks @eed3si9n for the fast reaction.

PR #932, opened 2017-01-18
Fixes issue #789 where brpJavaRepack was negated

### LINKED ISSUE #789: brp-java-repack-jars still running despite setting rpmBrpJavaRepackJars to false

YuvalItzchakov:
Fixes #789 where the value of `rpmBrpJavaRepackJars` was required to be set to `false` in order for repacking *not* to occur.

I've done three things:

1. Created a shorter pre script which utilizes `sed` in a single liner, since there are versions which don't support multi line macros (when running the test I kept getting errors about macro having an empty body, see [this](https://www.redhat.com/archives/fedora-packaging/2005-July/msg00113.html) for more). I also changed the macro prefix from %define to %global, as that's the recommended approach by Fedora.

2. Negated the default flag from `false` to `true`. My assumption was people might not want to disable repackaging out of the box, but feel free to tell me otherwise

3. Added a test to validate that the removal of brpRepack works when the flag is set to `false`.

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/rpm/brpJavaRepackJar
+++ b/src/main/resources/com/typesafe/sbt/packager/rpm/brpJavaRepackJar
@@ -1,6 +1,1 @@
-%define __os_install_post \
-%{_rpmconfigdir}/brp-compress \
-%{!?__debug_package:%{_rpmconfigdir}/brp-strip %{__strip}} \
-%{_rpmconfigdir}/brp-strip-static-archive %{__strip} \
-%{_rpmconfigdir}/brp-strip-comment-note %{__strip} %{__objdump} \
-%{nil}
+%global __os_install_post %(echo '%{__os_install_post}' | sed -e 's!/usr/lib[^[:space:]]*/brp-java-repack-jars[[:space:]].*$!!g')
--- a/src/main/resources/com/typesafe/sbt/packager/rpm/brpJavaRepackJar
+++ b/src/main/resources/com/typesafe/sbt/packager/rpm/brpJavaRepackJar
@@ -1,1 +1,1 @@
-%global __os_install_post %(echo '%{__os_install_post}' | sed -e 's!/usr/lib[^[:space:]]*/brp-java-repack-jars[[:space:]].*$!!g')
+%define __jar_repack %nil
--- a/src/main/scala/com/typesafe/sbt/packager/rpm/RpmPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/rpm/RpmPlugin.scala
@@ -81,7 +81,7 @@ object RpmPlugin extends AutoPlugin {
     rpmConflicts := Seq.empty,
     rpmSetarch := None,
     rpmChangelogFile := None,
-    rpmBrpJavaRepackJars := true,
+    rpmBrpJavaRepackJars := false,
     rpmPretrans := None,
     rpmPre := None,
     rpmPost := None,
--- a/src/main/scala/com/typesafe/sbt/packager/rpm/RpmPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/rpm/RpmPlugin.scala
@@ -81,7 +81,7 @@ object RpmPlugin extends AutoPlugin {
     rpmConflicts := Seq.empty,
     rpmSetarch := None,
     rpmChangelogFile := None,
-    rpmBrpJavaRepackJars := false,
+    rpmBrpJavaRepackJars := true,
     rpmPretrans := None,
     rpmPre := None,
     rpmPost := None,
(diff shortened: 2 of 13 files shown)


src/main/scala/com/typesafe/sbt/packager/rpm/RpmPlugin.scala
```diff
@@ -121,7 +121,7 @@ object RpmPlugin extends AutoPlugin {
       (rpmProvides, rpmRequirements, rpmPrerequisites, rpmObsoletes, rpmConflicts) apply RpmDependencies,
     maintainerScripts in Rpm := {
       val scripts = (maintainerScripts in Rpm).value
-      if (rpmBrpJavaRepackJars.value) {
+      if (!rpmBrpJavaRepackJars.value) {
```
muuki88: Yep. This is the actual bug fix. Thanks :)

src/main/scala/com/typesafe/sbt/packager/rpm/RpmPlugin.scala
```diff
@@ -81,7 +81,7 @@ object RpmPlugin extends AutoPlugin {
     rpmConflicts := Seq.empty,
     rpmSetarch := None,
     rpmChangelogFile := None,
-    rpmBrpJavaRepackJars := false,
+    rpmBrpJavaRepackJars := true,
```
muuki88: Together with the fix  this doesn't change the default behaviour, right?

I'm not building rpm's on a daily basis, so I'm not sure what impact this has. When we introduced
this it was because `rpm:packageBin` took a long time to extract and repack all the jar files, which
is IMHO wasted time. I would prefer `false` as a default.

src/main/resources/com/typesafe/sbt/packager/rpm/brpJavaRepackJar
```diff
@@ -1,6 +1 @@
-%define __os_install_post \
-%{_rpmconfigdir}/brp-compress \
-%{!?__debug_package:%{_rpmconfigdir}/brp-strip %{__strip}} \
-%{_rpmconfigdir}/brp-strip-static-archive %{__strip} \
-%{_rpmconfigdir}/brp-strip-comment-note %{__strip} %{__objdump} \
-%{nil}
+%global __os_install_post %(echo '%{__os_install_post}' | sed -e 's!/usr/lib[^[:space:]]*/brp-java-repack-jars[[:space:]].*$!!g')
```
muuki88: I found [this bug entry](https://bugzilla.redhat.com/show_bug.cgi?id=219731). Did you try the line

```
%define __jar_repack 0
```
?

src/main/scala/com/typesafe/sbt/packager/rpm/RpmPlugin.scala
YuvalItzchakov: @muuki88 Right, I negated the default because this is basically the behavior people had without actually knowing it so far. The time impact is pretty major if you're doing daily builds. I was waiting for one of you guys to set the tone on this, i'll revert the default to `false`.

src/main/resources/com/typesafe/sbt/packager/rpm/brpJavaRepackJar
YuvalItzchakov: Going to try it now, I saw:

`%define __jar_repack %nil`

Yestarday. Will try.

YuvalItzchakov: It works :)

src/main/scala/com/typesafe/sbt/packager/rpm/RpmPlugin.scala
```diff
@@ -81,7 +81,7 @@ object RpmPlugin extends AutoPlugin {
     rpmConflicts := Seq.empty,
     rpmSetarch := None,
     rpmChangelogFile := None,
-    rpmBrpJavaRepackJars := true,
+    rpmBrpJavaRepackJars := false,
```
YuvalItzchakov: The documentation regarding [jar repackaging](http://www.scala-sbt.org/sbt-native-packager/formats/rpm.html#jar-repackaging) is now going to be outdated (and generally it's pointing to a blog post that no longer exists). Where is the documentation maintained?

muuki88:
Sorry for the delay @YuvalItzchakov I'm still getting used the new review tools by Github :)
I missed your comments.

Changes look pretty good! Thanks for the effort :)

Issue #937, opened 2017-02-03
Ability to independently override different systemloader templates


ANorwell:
sbt-native-packager provides the ability to include a single override template in (e.g. for systemd) `templates/systemloader/systemd`. Additionally, there are multiple default templates provided, which are used as fallbacks,  e.g. for systemd `loader_functions` and `start_template`.  

However, it seems that providing the override template will cause it to override all templates. It doesn't seem like there's a way to override `start_template` without overriding `loader_functions` as well, with the same file. This rarely makes sense.

To give a concrete example, for systemd I want to modify the unit definition beyond what's in `start_template`. So, I provide `templates/systemloader/systemd` with my modified unit template. But, this results in this template also being used for loader_functions and being inserted into the start script, which fails to run.

To me, the solution here is to allow users to provide `loader_functions` and `start_template` files, not a single `systemd` file. This would be somewhat of a breaking change, though.

ANorwell:
The functionality to override partial templates seems to already exist, via `linuxScriptReplacements`. However, it remains true that at least for systemd, the behaviour when `templates/systemloader/systemd` is overridden does not seem to make any sense.

ANorwell:
I've found a very hacky workaround that allows me to insert arbitrary entries into the sytemd unit, as follows:

```
linuxScriptReplacements += "SuccessExitStatus" -> "\nLimitNOFILE=16384"
```

Obviously this is a complete hack, but in some ways it's better than redefining the entire unit file template. I wonder if it would make sense for the unit file template to have something like `${{additionalSystemdOptions}}`, that would allow for this.

muuki88:
Thanks @ANorwell for the detailed explanation and concrete use cases.

From your good research and the one example you are giving, I assume you use native-packager `1.1.x` or lower. With `1.2.0-M8` the systemd plugin is extracted in its own autoplugin and you can set the `SuccessExitStatus` via an sbt setting.

```scala
enablePlugin(SystemdPlugin)
systemdSuccessExitStatus := Seq("0", "1")
```

If you like to have additional settings, you can submit a pull request. The _hacky workaround_ is actually how things work in native-packager. The `linuxScriptReplacements` are a bunch of string-replacements. It's not very elegant, but straightforward and easy to debug.

> However, it seems that providing the override template will cause it to override all templates

[...]

PR #938, opened 2017-02-03
Allow custom templates to be provided separately for each template

ANorwell:
This is a proposed fix for to https://github.com/sbt/sbt-native-packager/issues/937.

Unfortunately, this is a breaking change to the templates directory layout, but it seems to be necessary to allow the different templates to be set independently.

Also, I'm not familiar with the sbt-native-packager codebase. This seems to be the right change, but it's possible this will have unintended consequences.


src/main/scala/com/typesafe/sbt/packager/archetypes/systemloader/package.scala
```diff
@@ -71,5 +71,5 @@ package object systemloader {
     loader.toString + "/" + name
 
   private def overrideFromFile(sourceDirectory: File, loader: ServerLoader, name: String): Option[URL] =
-    Option(sourceDirectory / "templates" / "systemloader" / loader.toString).filter(_.exists).map(_.toURI.toURL)
+    Option(sourceDirectory / "templates" / "systemloader" / loader.toString / name).filter(_.exists).map(_.toURI.toURL)
```
muuki88: Thanks for fixing this. This quite some bug. Along with 6 tests that tested the wrong implementation:

```
[error] (*:scripted) Failed tests:
[error] 	rpm / override-start-script-systemd
[error] 	rpm / override-start-script-systemv
[error] 	rpm / override-start-script-upstart
[error] 	debian / override-start-script-systemd
[error] 	debian / override-start-script-systemv
[error] 	debian / override-start-script-upstart
```

typesafehub-validator:
Hi @ANorwell,

ANorwell:
I fixed the tests, and also added two new tests for overriding loader-scripts, one for deb and one for rpm.

src/sbt-test/debian/override-loader-functions/build.sbt
```diff
@@ -0,0 +1,26 @@
...
+  with multiple lines."""
+
+TaskKey[Unit]("check-loader-functions") <<= (target, streams) map { (target, out) =>
+  val extracted = target / "tmp" / "extracted-package"
+  extracted.mkdirs()
+  Seq("dpkg-deb", "-e", (target / "debian-test_0.1.0_all.deb").absolutePath, extracted.absolutePath).!
+
+  val script = IO.read(extracted / "postinst")
+  
+  assert(script.contains("# right systemd template"), s"override script wasn't picked, script is\n$script")
```
muuki88: This checks that the `loader-functions` has be correctly inserted. I think this check is enough, even though I wouldn't my checking that the `wrong systemd template` isn't contained in the script.

ANorwell: added such an assertion

muuki88:
Thanks again for your time and work 😊
I'll release this ASAP

PR #954, opened 2017-03-29
Add new system loader setting for file descriptor limit

levinson:
There currently is not a good way to increase the file descriptor limit for system loaders. I have added a new system loader key `fileDescriptorLimit` with support for upstart, systemd and systemv loaders, which can be overridden to change the default limit of 1024.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/systemloader/SystemloaderPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/systemloader/SystemloaderPlugin.scala
@@ -5,7 +5,6 @@ import sbt.Keys.{sourceDirectory, target}
 import com.typesafe.sbt.SbtNativePackager.{Debian, Rpm}
 import com.typesafe.sbt.packager.Keys.{
   defaultLinuxStartScriptLocation,
-  fileDescriptorLimit,
   killTimeout,
   linuxMakeStartScript,
   linuxPackageMappings,
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/systemloader/SystemloaderPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/systemloader/SystemloaderPlugin.scala
@@ -5,6 +5,7 @@ import sbt.Keys.{sourceDirectory, target}
 import com.typesafe.sbt.SbtNativePackager.{Debian, Rpm}
 import com.typesafe.sbt.packager.Keys.{
   defaultLinuxStartScriptLocation,
+  fileDescriptorLimit,
   killTimeout,
   linuxMakeStartScript,
   linuxPackageMappings,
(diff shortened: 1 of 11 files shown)


typesafehub-validator:
Hi @levinson,

src/main/scala/com/typesafe/sbt/packager/archetypes/systemloader/SystemloaderPlugin.scala
```diff
@@ -158,6 +162,7 @@ object SystemloaderPlugin extends AutoPlugin {
       "stop_runlevels" -> stopRunlevels.getOrElse(""),
       "start_facilities" -> startOn.getOrElse(""),
       "stop_facilities" -> stopOn.getOrElse(""),
+      "file_descriptor_limit" -> fileDescriptorLimit.getOrElse(""),
```
muuki88: What happens if `ulimit` is called with no arguments? Will this result in `unlimited`?

src/main/scala/com/typesafe/sbt/packager/archetypes/systemloader/SystemloaderPlugin.scala
```diff
@@ -50,6 +51,7 @@ object SystemloaderPlugin extends AutoPlugin {
   def systemloaderSettings: Seq[Setting[_]] = Seq(
     serverLoading := None,
     serviceAutostart := true,
+    fileDescriptorLimit := Some("1024"),
```
muuki88: I think this should go into the `LinuxPlugin`. At somepoint I will refactor this into the `SystemloaderPlugin` where it belongs, but for this PR it's fine to put it into the `LinuxPlugin`

src/sphinx/archetypes/systemloaders.rst
```diff
@@ -49,6 +49,9 @@ General Settings
   ``retryTimeout``
     Timeout between retries in seconds
 
+  ``fileDescriptorLimit``
+    Maximum number of open file descriptors for the spawned application. The default value is 1024.
```
muuki88: :heart:

src/main/scala/com/typesafe/sbt/packager/archetypes/systemloader/SystemloaderPlugin.scala
zakcrawford: If ulimit is called with no arguments it assumes `-f` and will not set limits, instead it will return the value of  `The maximum size of files created by the shell`.

From the man pages: http://linuxcommand.org/man_pages/ulimit1.html
```
 If limit is given, it is the new value of the specified resource
              (the -a option is display only).  If no option is given, then -f
              is  assumed.  Values are in 1024-byte increments, except for -t,
              which is in seconds, -p, which is in units of  512-byte  blocks,
              and  -n and -u, which are unscaled values.  The return status is
              0 unless an invalid option or argument is supplied, or an  error
              occurs while setting a new limit.
```

muuki88:
Awesome work. Thanks

levinson:
Thanks for the quick review and merge!

PR #962, opened 2017-04-12
Add Docker labels easily

NeQuissimus:
This change allows for an easy way to add labels to Docker images.
It was previously possible to add this as custom commands but given the importance of labels, it seems reasonable to make adding them more easily.

I believe there is not much documentation necessary. I added a reference to the settings key.

As an example, we will be using this to forward Jenkins job variables into our Docker images.

```
dockerLabels := Map(
  "BUILD" -> sys.env.getOrElse("BUILD_NUMBER", "local"),
  "GIT_COMMIT" -> sys.env.getOrElse("GIT_COMMIT", "local")
)
```


src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -90,22 +91,23 @@ object DockerPlugin extends AutoPlugin {
...
+          Seq(
+            makeWorkdir(dockerBaseDirectory),
+            makeAdd(dockerBaseDirectory),
+            makeChown(user, group, "." :: Nil)
+          ) ++
+          dockerLabels.value.map(makeLabel) ++
+          makeExposePorts(dockerExposedPorts.value, dockerExposedUdpPorts.value) ++
+          makeVolumes(dockerExposedVolumes.value, user, group) ++
+          Seq(makeUser(user), makeEntrypoint(dockerEntrypoint.value), makeCmd(dockerCmd.value))
+      }
```
muuki88: These look like formatting changes. When you run `sbt scalafmt` it should format correctly.
Or if you use intellij along with the `scalafmt` plugin you may use a newer or older version.

NeQuissimus:
I had forgotten about a test :)

I ran `scalafmt` against everything, seems there were a few files it was unhappy with...

muuki88:
Thanks a lot :)

PR #965, opened 2017-04-17
Add docker:clean task

### LINKED ISSUE #501: Clean image after publishing
We are using a CI system to push our images to the registry, after every build.
After pushing an image to the registry, the image is left in the local registry.
It would be great if we could have an extra target to push an image and remove the image locally afterwards.
Something like:
 `activator clean docker:publishClean`

NeQuissimus:
This task will call `docker rmi` by default.
The command can be controlled using the dockerRmiCommand key.

Fixes #501 

- Added a test that tests for the images' existence before and after
- Reused `clean` as it is familiar
- Command override provided as key
- Documentation

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -7,6 +7,7 @@ import java.util.concurrent.atomic.AtomicBoolean
 import sbt._
 import sbt.Keys.{
   cacheDirectory,
+  clean,
   mappings,
   name,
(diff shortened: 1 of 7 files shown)


NeQuissimus:
@muuki88 Can you kick off the CIs again? The Typesafe Bintray was down...

muuki88:
Ah, sure.

NeQuissimus:
@muuki88 Looks like the Windows guy needs yet another poke :D

muuki88:
Sorry for the late reply. The appveyor stuff has some other issue :P 

Thanks a lot for your contribution :)

PR #989, opened 2017-05-27
Implement stage command for rpm

### LINKED ISSUE #848: add rpmExplodedPackage command
The debian plugin has `debianExplodedPackage` which creates the package in a format that mirrors that of the installed package.  `debianExplodedPackage` is a dependency of `debian:packageBin`.   Neither `debianExplodedPackage` nor `debian:packageBin` deletes the exploded package.  Having the exploded package is useful for inspection.  Running `debianExplodedPackage` is considerably faster than running `debian:packageBin` and enables bypassing the expensive package build/install steps for developer testing (you can just copy the files into your install directory).

The rpm plugin combines both steps into `rpm:packageBin` and deletes the exploded package after completion.  This change would restructure `rpm:packageBin` and introduce a separate `rpmExplodedPackage` task.  Like the debian plugin, `rpm:packageBin` will not delete the exploded package.

muuki88:
Similar to `debian:stage`, `docker:stage` and `universal:stage` the rpm workspace can now be stage as well.

This should fix #848

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/rpm/RpmHelper.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/rpm/RpmHelper.scala
@@ -11,16 +11,35 @@ object RpmHelper {
   def hostVendor =
     Process(Seq("rpm", "-E", "%{_host_vendor}")) !!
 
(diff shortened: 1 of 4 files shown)


muuki88:
This may be of interest for you guys: @dpennell @fsat

PR #992, opened 2017-05-31
Allow specifying classifier for default package in deployment settings

Krever:
Hi there

This PR introduce the ability to provide an artifact classifier for default packages when calling `makeDeploymentSettings`. 

We need this feature in our build process and I see no other workaround that to produce second artifacts with `addPackage` call.


muuki88:
I restarted the travis build as it seemed to have timed out.

PR #1000, opened 2017-06-19
Cross build with sbt 1.0

muuki88:
Initial attempt to cross build sbt-native-packager for sbt `1.0.0-M6` and `0.13.15`.

Inspired by and copied from

- https://github.com/sbt/sbt/issues/3245
- https://github.com/sbt/sbt-bintray/commit/3ec91ad8ec8e68192a19e705038f1362846548cd
- https://github.com/sbt/sbt-pgp/commit/d1bfc03931a873c27bea79f4b1b43f044210e688
- https://github.com/ktoso/sbt-jmh/commit/db26a9c260d551537d89681f8c6ef2bffc7edb3c


muuki88:
First of all: Even this is the first major version bump of SBT and native-packager is a pretty old plugin, the necessary changes are really small and comprehensible.

I mostly struggled with the the `PathFinder` changes as some operators are gone, the dependency has to be explicitly added and I had to figure out were the new implicit conversions are located :D But nothing awkward or confusing.

What puzzles me at the moment is the changed behavior with special artifacts and artifact scopes. We have two provided dependencies:

```
// used in the DockerSpotifyClientPlugin
"com.spotify" % "docker-client" % "3.5.13" % "provided" 

// used in the JDebPackaging plugin
"org.vafer" % "jdeb" % "1.3" % "provided" artifacts Artifact("jdeb", "jar", "jar")
```

The `provided` scope used to work for all scripted tests, which means
- I need to add the dependency explicitly if I want to test the plugin
- If the plugin is not being activated, then the classes aren't loaded

Apparently this lazy loading behavior changed as I now get `ClassNotFound` exceptions in every scripted test. Removing the `provided` scope fixed the issue.

The second thing that puzzled me was that the `jdeb` dependency still wasn't working with `ClassNotFound` exceptions. `"org.vafer" % "jdeb" % "1.3"` is a maven plugin, thus a special artifact is added ( as described in https://github.com/sbt/sbt/issues/499 ). However this doesn't seem to be sufficient enough. I applied the same fix we have in place for coursier ( 2e5b9ec7e59e4f4ffcefdf0758a2cb9bb2a80849 ) and added 

```scala
classpathTypes += "maven-plugin"
```

to the native-packager `build.sbt`. I think this change makes totally sense, but I guess this could be a problem for some other plugin authors.

src/sbt-test/universal/multiproject-classifiers/build.sbt
```diff
@@ -0,0 +1,42 @@
...
+
+lazy val assets = config("assets")
+
+lazy val sub = project
+  .in(file("sub"))
+  .enablePlugins(JavaAppPackaging)
+  .settings(mySettings)
+  .settings(
+    ivyConfigurations += assets,
+    artifact in assets := artifact.value.withClassifier(classifier = Some("assets")),
```
muuki88: @eed3si9n Where can I find the latest `Artifact` API definition? I wasn't able to find the publish API documentation, or the sources files in the `sbt-librarymanagment` project. I actually guessed this method call :stuck_out_tongue: 
Before it was `copy(classifier = Some("assets"))`

muuki88: The [sbt docs on Aritfacts](http://www.scala-sbt.org/1.0/docs/Artifacts.html) still state there is a `copy` method

eed3si9n: It's generated from https://github.com/sbt/librarymanagement/blob/1.0/librarymanagement/src/main/contraband/librarymanagement.json. I should make sure the generated code goes into GitHub like other sbt modules.

bpholt:
Is there anything else that needs to be done to get this merged? I'm working on converting my company's plugins to sbt 1.0 and like @eed3si9n said, this is an important one. If there's something I can help with, please let me know.

muuki88:
Thanks a lot for your offer 😍
Are you using one of the "Deployment" plugins like `UniversalDeploymentPlugin`? If so, could you test it with this branch?

I tried to release a first RC1 with this, but messed up at some point. The release now takes almost 1 1/2 hours ony machine due to the scripted tests in two versions.

muuki88:
The first release candidate is out `1.2.2-RC2`

bpholt:
@muuki88 We are using `DockerPlugin` with `JavaServerAppPackaging` and as far as I can tell, it's working fine. I built and published it locally, and pulled it into our [sbt-docker-containers](https://github.com/Dwolla/sbt-docker-containers) plugin as well as the private plugin we maintain with default settings for our apps, and built and ran an app locally.

I'll test again with `1.2.2-RC2` but I think we're on the right track! 😃

I did notice a few eviction warnings during the build, which may or may not be a problem. We try to treat warnings as errors so I'm conditioned to worry about them, idk.

bpholt:
Following up on earlier—RC2 looks good AFAICT, other than the previously mentioned eviction warnings.

PR #1003, opened 2017-06-22
Fix a bug which cause upgrade failing because of no restartService

buster84:
I got the following error when upgrading:
```
/var/tmp/rpm-tmp.zpTtMD: line 4: restartService: command not found
```


muuki88:
Thanks a lot! Changes are looking good.

PR #1013, opened 2017-08-17
Wip/build with sbt 1.0

muuki88:


ennru:
Go, @muuki88 go!

muuki88:
Two things left

[ ] Fix test projects with https://github.com/tkroman/src-dep-auto-plugin-example
[ ] Fix windows test command

PR #1014, opened 2017-08-21
Wrap docker labels with " character

kimxogus:
## Issue
For now, `dockerLabels` doesn't generate labels with wrapping `"` character.
- build.sbt
```sbt
dockerLabels := Map("foo" -> "bar")
```
- Generated Dockerfile
```Dockerfile
LABEL foo=bar
```

This may have issue with string values with spaces
- build.sbt
```sbt
dockerLabels := Map("fooBar" -> "foo bar")
```
- Generated Dockerfile
```Dockerfile
LABEL fooBar=foo bar
```

## Fix
Fixed that to wrap with `"` character.
- build.sbt
```sbt
dockerLabels := Map("fooBar" -> "foo bar", "number" -> 123)
```
- Generated Dockerfile
```Dockerfile
LABEL fooBar="foo bar"
LABEL number=123
```
Also, `dockerLabels` will accept map containing any type of value and make them string automatically(for convenience).
Only `Number` typed values won't be wrapped with `"` character

- [Docker doc reference](https://docs.docker.com/engine/reference/builder/#label)

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -169,7 +169,7 @@ object DockerPlugin extends AutoPlugin {
     */
   private final def makeLabel(label: (String, String)): CmdLike = {
     val (variable, value) = label
-    Cmd("LABEL", s"$variable=\"${value.toString}\"")
+    Cmd("LABEL", variable + "=\"" + value.toString + "\"")
   }
 
   /**
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -167,13 +167,9 @@ object DockerPlugin extends AutoPlugin {
     * @param label
     * @return LABEL command
     */
-  private final def makeLabel(label: (String, Any)): CmdLike = {
+  private final def makeLabel(label: (String, String)): CmdLike = {
     val (variable, value) = label
-    val valueString = value match {
-      case n: Number => n.toString
-      case _ => "\"" + value.toString + "\""
-    }
-    Cmd("LABEL", s"$variable=$valueString")
+    Cmd("LABEL", s"$variable=\"${value.toString}\"")
   }
 
   /**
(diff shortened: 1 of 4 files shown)


lightbend-cla-validator:
Hi @kimxogus,

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -167,9 +167,13 @@ object DockerPlugin extends AutoPlugin {
     * @param label
     * @return LABEL command
     */
-  private final def makeLabel(label: Tuple2[String, String]): CmdLike = {
+  private final def makeLabel(label: (String, Any)): CmdLike = {
     val (variable, value) = label
-    Cmd("LABEL", s"${variable}=${value}")
+    val valueString = value match {
+      case n: Number => n.toString
```
muuki88: Why the special case for numbers? I haven't found an example that highlights
a special behavior?

src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala
```diff
@@ -31,7 +31,7 @@ trait DockerKeys {
   val dockerExecCommand = SettingKey[Seq[String]]("dockerExecCommand", "The shell command used to exec Docker")
   val dockerBuildOptions = SettingKey[Seq[String]]("dockerBuildOptions", "Options used for the Docker build")
   val dockerBuildCommand = SettingKey[Seq[String]]("dockerBuildCommand", "Command for building the Docker image")
-  val dockerLabels = SettingKey[Map[String, String]]("dockerLabels", "Labels applied to the Docker image")
+  val dockerLabels = SettingKey[Map[String, Any]]("dockerLabels", "Labels applied to the Docker image")
```
muuki88: `Any` is not a good type here. If the number/string separation is really necessary I would rather map this into a separate ADT. We then need to introduce implicit conversions to not break existing builds.

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
kimxogus: Oh, I scanned Dockerfile docs again and it seems not to be needed. I'll change it to wrap everything in `"`.

src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala
kimxogus: ok

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
muuki88: Awesome :)

PR #1020, opened 2017-08-28
Change startup script name generation

### LINKED ISSUE #1016: Duplicated startup script names in universal zip
When the project contains several main classes with the same name in different packages, universal zip contains multiple startup scripts with the same name inside the `bin/` directory.

## How to reproduce

1. Unzip the attached example: [native-packager-bug.zip](https://github.com/sbt/sbt-native-packager/files/1245294/native-packager-bug.zip)
2. Run `sbt 'universal:packageBin'`

## Expected behaviour

Create differently named startup scripts (for example, prepend their name with package name when clashes occur).

## Actual behaviour

The created file contains duplicated file names:
```
$ unzip -l target/universal/test-1.0.zip 
Archive:  target/universal/test-1.0.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
     5403  2017-08-23 15:00   test-1.0/lib/test.test-1.0.jar
  5246851  2017-07-25 13:47   test-1.0/lib/org.scala-lang.scala-library-2.12.3.jar
     9159  2017-08-23 15:00   test-1.0/bin/test
     9159  2017-08-23 15:00   test-1.0/bin/test
     4064  2017-08-23 15:00   test-1.0/bin/test.bat
     4064  2017-08-23 15:00   test-1.0/bin/test.bat
---------                     -------
  5278700                     6 files
$ unzip target/universal/test-1.0.zip 
Archive:  target/universal/test-1.0.zip
  inflating: test-1.0/lib/test.test-1.0.jar  
  inflating: test-1.0/lib/org.scala-lang.scala-library-2.12.3.jar  
  inflating: test-1.0/bin/test       
replace test-1.0/bin/test? [y]es, [n]o, [A]ll, [N]one, [r]ename: 
```

## Information

Version information:
sbt: 1.0.0
sbt-native-packager: 1.2.2
build system: Ubuntu 16.04.3 LTS

atrosinenko:
This PR tries to lower the chance of generation of multiple scripts
with equal names in one archive. Fixes #1016.

Additionally, an attempt was made to improve the word splitting
for the script names, so, for example `UITest` becomes `ui-test`
and not `u-i-test`.

I had to edit `ForwarderScripts.apply(...)` as well but I'm not sure I did it the right way.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/CommonStartScriptGenerator.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/CommonStartScriptGenerator.scala
@@ -33,7 +33,11 @@ trait CommonStartScriptGenerator {
     */
   protected[this] val keySurround: String => String
 
-  protected[this] val makeScriptsExecutable: Boolean
+  /**
+    * Set executable bit of the generated scripts to this value
+    * @todo Does it work when building archives on hosts that do not support such permission?
+    */
+  protected[this] val executableBitValue: Boolean
 
   protected[this] def createReplacementsForMainScript(mainClass: String,
                                                       mainClasses: Seq[String],
@@ -48,6 +52,11 @@ trait CommonStartScriptGenerator {
     def withScriptName(scriptName: String): SpecializedScriptConfig
   }
 
+  /**
+    * The type of specialized ScriptConfig.
+    * This enables callback methods of the concrete plugin implementations
+    * to use fields of config that only exist in their ScriptConfig specialization.
+    */
   protected[this] type SpecializedScriptConfig <: ScriptConfig
 
   protected[this] def generateStartScripts(config: SpecializedScriptConfig,
@@ -100,10 +109,10 @@ trait CommonStartScriptGenerator {
     val scriptContent = TemplateWriter.generateScript(template, replacements, eol, keySurround)
     val scriptNameWithSuffix = mainScriptName(config)
     val script = targetDir / scriptTargetFolder / scriptNameWithSuffix
+
     IO.write(script, scriptContent)
     // TODO - Better control over this!
-    if (makeScriptsExecutable)
-      script.setExecutable(true)
+    script.setExecutable(executableBitValue)
     script -> s"$scriptTargetFolder/$scriptNameWithSuffix"
   }
 
@@ -129,8 +138,7 @@ trait CommonStartScriptGenerator {
         val scriptContent = TemplateWriter.generateScript(forwarderTemplate, replacements, eol, keySurround)
 
         IO.write(file, scriptContent)
-        if (makeScriptsExecutable)
-          file.setExecutable(true)
+        file.setExecutable(executableBitValue)
         file -> s"$scriptTargetFolder/$scriptName"
     }
   }
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/CommonStartScriptGenerator.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/CommonStartScriptGenerator.scala
@@ -1,0 +1,126 @@
+package com.typesafe.sbt.packager.archetypes.scripts
+
+import java.io.File
+import java.net.URL
+
+import com.typesafe.sbt.packager.archetypes.TemplateWriter
+import sbt._
+
+trait CommonStartScriptGenerator {
+
+  /**
+    * Script destination in final package
+    */
+  protected[this] val scriptTargetFolder: String = "bin"
+
+  /**
+    * Suffix to append to the generated script name (such as ".bat")
+    */
+  protected[this] val scriptSuffix: String
+
+  /**
+    * Name of the forwarder template if user wants to provide custom one
+    */
+  protected[this] val forwarderTemplateName: String
+
+  /**
+    * Line separator for generated scripts
+    */
+  protected[this] val eol: String
+
+  /**
+    * keySurround for TemplateWriter.generateScript()
+    */
+  protected[this] val keySurround: String => String
+
+  protected[this] val makeScriptsExecutable: Boolean
+
+  protected[this] def createReplacementsForMainScript(mainClass: String,
+                                                      mainClasses: Seq[String],
+                                                      config: SpecializedScriptConfig): Seq[(String, String)]
+
+  protected[this] trait ScriptConfig {
+    val executableScriptName: String
+    val scriptClasspath: Seq[String]
+    val replacements: Seq[(String, String)]
+    val templateLocation: File
+
+    def withScriptName(scriptName: String): SpecializedScriptConfig
+  }
+
+  protected[this] type SpecializedScriptConfig <: ScriptConfig
+
+  protected[this] def generateStartScripts(config: SpecializedScriptConfig,
+                                           mainClass: Option[String],
+                                           discoveredMainClasses: Seq[String],
+                                           targetDir: File,
+                                           log: sbt.Logger): Seq[(File, String)] =
+    StartScriptMainClassConfig.from(mainClass, discoveredMainClasses) match {
+      case NoMain =>
+        log.warn("You have no main class in your project. No start script will be generated.")
+        Seq.empty
+      case SingleMain(main) =>
+        Seq(createMainScript(main, config, targetDir, Seq(main)))
+      case MultipleMains(mains) =>
+        generateMainScripts(mains, config, targetDir)
+      case ExplicitMainWithAdditional(main, additional) =>
+        createMainScript(main, config, targetDir, discoveredMainClasses) +:
+          createForwarderScripts(config.executableScriptName, additional, targetDir)
+    }
+
+  private[this] def generateMainScripts(discoveredMainClasses: Seq[String],
+                                        config: SpecializedScriptConfig,
+                                        targetDir: File): Seq[(File, String)] =
+    ScriptUtils.createScriptNames(discoveredMainClasses).map {
+      case (qualifiedClassName, scriptName) =>
+        val newConfig = config.withScriptName(scriptName)
+        createMainScript(qualifiedClassName, newConfig, targetDir, discoveredMainClasses)
+    }
+
+  /**
+    *
+    * @param mainClass - Main class added to the java command
+    * @param config - Config data for this script
+    * @param targetDir - Target directory for this script
+    * @return File pointing to the created main script
+    */
+  private[this] def createMainScript(mainClass: String,
+                                     config: SpecializedScriptConfig,
+                                     targetDir: File,
+                                     mainClasses: Seq[String]): (File, String) = {
+    val template = resolveTemplate(config.templateLocation)
+    val replacements = createReplacementsForMainScript(mainClass, mainClasses, config)
+    val scriptContent = TemplateWriter.generateScript(template, replacements, eol, keySurround)
+    val scriptNameWithSuffix = config.executableScriptName + scriptSuffix
+    val script = targetDir / scriptTargetFolder / scriptNameWithSuffix
+    IO.write(script, scriptContent)
+    // TODO - Better control over this!
+    if (makeScriptsExecutable)
+      script.setExecutable(true)
+    script -> s"$scriptTargetFolder/$scriptNameWithSuffix"
+  }
+
+  private[this] def resolveTemplate(defaultTemplateLocation: File): URL =
+    if (defaultTemplateLocation.exists) defaultTemplateLocation.toURI.toURL
+    else getClass.getResource(defaultTemplateLocation.getName)
+
+  private[this] def createForwarderScripts(executableScriptName: String,
+                                           discoveredMainClasses: Seq[String],
+                                           targetDir: File): Seq[(File, String)] = {
+    val tmp = targetDir / scriptTargetFolder
+    val forwarderTemplate = getClass.getResource(forwarderTemplateName)
+    ScriptUtils.createScriptNames(discoveredMainClasses).map {
+      case (qualifiedClassName, scriptNameWithoutSuffix) =>
+        val scriptName = scriptNameWithoutSuffix + scriptSuffix
+        val file = tmp / scriptName
+
+        val replacements = Seq("startScript" -> executableScriptName, "qualifiedClassName" -> qualifiedClassName)
+        val scriptContent = TemplateWriter.generateScript(forwarderTemplate, replacements, eol, keySurround)
+
+        IO.write(file, scriptContent)
+        if (makeScriptsExecutable)
+          file.setExecutable(true)
+        file -> s"$scriptTargetFolder/$scriptName"
+    }
+  }
+}
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/CommonStartScriptGenerator.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/CommonStartScriptGenerator.scala
@@ -62,20 +62,27 @@ trait CommonStartScriptGenerator {
       case SingleMain(main) =>
         Seq(createMainScript(main, config, targetDir, Seq(main)))
       case MultipleMains(mains) =>
-        generateMainScripts(mains, config, targetDir)
+        generateMainScripts(mains, config, targetDir, log)
       case ExplicitMainWithAdditional(main, additional) =>
         createMainScript(main, config, targetDir, discoveredMainClasses) +:
-          createForwarderScripts(config.executableScriptName, additional, targetDir)
+          createForwarderScripts(config.executableScriptName, additional, targetDir, config, log)
     }
 
   private[this] def generateMainScripts(discoveredMainClasses: Seq[String],
                                         config: SpecializedScriptConfig,
-                                        targetDir: File): Seq[(File, String)] =
-    ScriptUtils.createScriptNames(discoveredMainClasses).map {
+                                        targetDir: File,
+                                        log: sbt.Logger): Seq[(File, String)] = {
+    val classAndScriptNames = ScriptUtils.createScriptNames(discoveredMainClasses)
+    ScriptUtils.warnOnScriptNameCollision(classAndScriptNames, log)
+    classAndScriptNames.map {
       case (qualifiedClassName, scriptName) =>
         val newConfig = config.withScriptName(scriptName)
         createMainScript(qualifiedClassName, newConfig, targetDir, discoveredMainClasses)
     }
+  }
+
+  private[this] def mainScriptName(config: ScriptConfig): String =
+    config.executableScriptName + scriptSuffix
 
   /**
     *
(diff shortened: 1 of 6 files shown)


lightbend-cla-validator:
Hi @atrosinenko,

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/ScriptUtils.scala
```diff
@@ -0,0 +1,68 @@
...
+    * @param discoveredMainClasses discovered qualified main class names
+    * @return sequence of tuples: (passed in class name) -> (generated script name)
+    */
+  def createScriptNames(discoveredMainClasses: Seq[String]): Seq[(String, String)] = {
+    val names = discoveredMainClasses.map { qualifiedClassName =>
+      val lowerCased = toLowerCase(qualifiedClassName)
+      (qualifiedClassName, lowerCased.replace('.', '-'), lowerCased.split("\\.").last)
+    }
+
+    names.groupBy(_._3).toSeq.flatMap {
```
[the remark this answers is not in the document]
atrosinenko: What is the preferred way to rephrase this?

atrosinenko:
Strange, at least one of Travis failures looks like some misconfiguration.

atrosinenko:
If you propose to extend the `simpleName`s just for the classes that generate conflicting script names without this, then it may introduce additional conflicts, something like
```
pkg1.Pkg1Test -> pkg-1-test
pkg1.Test     -> test -> pkg-1-test
pkg2.Test     -> test -> pkg-2-test
```
So it may be not so simple. On the other hand, when these heuristics fail, it does not render resulting zip completely unusable (though, it may became unusable in fully automatic setups). So there may be another option to drop the most nested numbering logic completely (not very good, does the other means lower the chance of failure to some negligible level?). Or apply some iterative algorithm.

Another problem, if we decide to iteratively extend the `simpleName` to not make it very long: consider the following classes:
```
com.company.feature1.core.MainClass
com.company.feature2.ui.MainClass
```
If we just extend it from right to left, then we probably end up with meaningless `core-main-class` and `ui-main-class`. In fact, we would prefer `feature-1-main-class` and `feature-2-main-class`. Another option may be to strip equal prefix from the left, then extend it as `main-class` -> `feature-1-main-class` -> `feature-1-core-main-class`.

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/ScriptUtils.scala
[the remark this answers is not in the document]
atrosinenko: Oh, I didn't thought about it. It really looks more readable.

muuki88:
> If you propose to extend the `simpleNames` just for the classes that generate conflicting script names without this, then it may introduce additional conflicts.

This is true if we change the class name. In your example the camel cased name `Pkg1Test`  is transformed to `pkg1-test`, which we shouldn't do. We should only transform `.`  (separating packages) to `-` in the script name. The compiler then ensures that if we expand from right to left in the end every bash script is unique.

> Another problem, if we decide to iteratively extend the `simpleName ` to not make it very long

I'm willing to choose this trade off. I consider this rather an edge case as this could be fixed on an application level by renaming classes or moving it to another package. I know naming is one of the harder problems in software engineering, but users of your scripts will be happy to have meaningful script names instead of weird package names ;)

I first thought of adding a "resolve conflict strategy" as sbt-assembly does it, but I think it's an overkill as we need to document and maintain it.

Long story short: A recursive algorithm extending only the conflicting classes until all conflicts are resolved seems the most stable and simplest solution :)

atrosinenko:
So should the lower casing algorithm be completely dropped? Previously, class names were splitted into words separated with `-` (but this failed with abbreviations), if I got it right.

muuki88:
Either this or we replace the dots with underscored. This would keep backwards compatibility and wouldn't generate toooo ugly script names.

lightbend-cla-validator:
Hi @atrosinenko,

atrosinenko:
Please excuse me for overcomplicating everything once again, but here is another try: now heuristics are even worse :( but they should give more meaningful results (see tests). Based on your review I have to either try to simplify it to some informally provable form or drop it in favor of simpler prefix-based strategy.

muuki88:
Thanks for your time and dedication @atrosinenko  :)
The results make sense (stripping common packages and keep script names short).

Still I think this in an edge case and we should rather add the ability to configure script names then a complicated algorithm that can't be explain in two sentences. Maintaining and explaining this will cost more time I fear ;)

What  do you thing of this implementation?

```scala
  private case class MainClass(scriptName: String, fullyQualifiedClassName: String) {
    val parts: Array[String] = toLowerCase(fullyQualifiedClassName).split("\\.")
    def asTuple: (String, String) = (fullyQualifiedClassName, scriptName)
  }

  def getMainClasses(discoveredMainClasses: Seq[String]): Seq[(String, String)] = {
    val mainClasses = discoveredMainClasses.map { qualifiedClassName =>
      val lowerCased = toLowerCase(qualifiedClassName)
      val parts = lowerCased.split("\\.")
      MainClass(parts.last, qualifiedClassName)
    }
    disambiguateNames(mainClasses, 1).map(_.asTuple)
  }

  private def disambiguateNames(mainClasses: Seq[MainClass], expansionIndex: Int): Seq[MainClass] =
    if (mainClasses.isEmpty) {
      // end of recursion
      mainClasses
    } else {
      // find duplicates
      val (duplicates, uniques) = mainClasses
        .groupBy(_.scriptName)
        .partition {
          case (_, classes) => classes.length > 1
        }

      // expand names for duplicated script names
      val expandedScriptNames = for {
        (_, classes) <- duplicates
        extendedClasses <- classes.map(expandClassName(expansionIndex))
      } yield extendedClasses

      uniques.toSeq.flatMap(_._2) ++ disambiguateNames(expandedScriptNames.toSeq, expansionIndex + 1)
    }

  private def expandClassName(expansionIndex: Int)(mainClass: MainClass): MainClass = {
    val scriptName = mainClass.parts.takeRight(expansionIndex + 1).mkString("_")
    mainClass.copy(scriptName = scriptName)
  }

```

atrosinenko:
Please excuse me for the delay and thank you for code advice. I have rewritten disambiguation logic once again, now heavily based on your proposal. Now it looks much simpler than from the second try. I decided to expand from the left after stripping common prefix instead of expanding from the right -- looks like it can give more meaningful distinction by features instead of some technical innermost packages (but I don't insist it is the most correct approach).

I had to manually rebase previous commits as well, so it may be worth to minimally glance on them.

atrosinenko:
Oops, fixed the corner case of single main class (the common prefix included the class name itself) -- that didn't caught by the tests on Linux.
PS: The CLA signing bot suggested me to notify the maintainer that CLA is signed for already opened PRs. :)

muuki88:
Thanks @atrosinenko I'll try to take a look this week. I'll be away for 4 weeks on vacation then and will merge it after that :)

muuki88:
I'll be on vacation un 20.10. Thanks for your understanding if this can't be merged until then

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/ScriptUtils.scala
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+    * Converts class name to lower case, applying some heuristics
+    * to guess the word splitting.
+    * @param qualifiedClassName a class name
+    * @return lower cased name with '-' between words. Dots ('.') are left as is.
+    */
+  def toLowerCase(qualifiedClassName: String): String = {
+    // suppose list is not very huge, so no need in tail recursion
+    def split(chars: List[Char]): List[Char] = chars match {
```
muuki88: I still think this is too much for an edge case. Prettifying should be done by the user.
We could do that in a second pull request where we provide a setting that let's the user
configure executable script names for each main class.

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/ScriptUtils.scala
```diff
@@ -0,0 +1,103 @@
...
+
+    def scriptName(expansionIndex: Int): String =
+      (packages.take(expansionIndex) :+ className).mkString("_")
+    def asTuple(expansionIndex: Int): (String, String) =
+      (fullyQualifiedClassName, scriptName(expansionIndex))
+  }
+
+  private[this] def disambiguateNames(
+                                       mainClasses: Seq[MainClass],
+                                       expansionIndex: Int
```
muuki88: I like that parameter name :+1:

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/ScriptUtils.scala
```diff
@@ -0,0 +1,103 @@
...
+    * @param discoveredMainClasses discovered qualified main class names
+    * @return sequence of tuples: (passed in class name) -> (generated script name)
+    */
+  def createScriptNames(discoveredMainClasses: Seq[String]): Seq[(String, String)] = {
+    val mainClasses = discoveredMainClasses.map { qualifiedClassName =>
+      val lowerCased = toLowerCase(qualifiedClassName)
+      val parts = lowerCased.split("\\.")
+      MainClass(qualifiedClassName, parts)
+    }
+    val commonPrefixLength = mainClasses.map(_.parts.init.toList).reduce(commonPrefix).size
```
muuki88: I would leave out every prettifying here. As mentioned below the way forward would be
a user facing setting where a mapping between mainClass and executable script name
can be configured.

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/ScriptUtils.scala
```diff
@@ -0,0 +1,103 @@
...
+    private val packages: Seq[String] = parts.init
+    private val className: String = parts.last
+
+    def scriptName(expansionIndex: Int): String =
+      (packages.take(expansionIndex) :+ className).mkString("_")
+    def asTuple(expansionIndex: Int): (String, String) =
+      (fullyQualifiedClassName, scriptName(expansionIndex))
+  }
+
+  private[this] def disambiguateNames(
```
muuki88: Well done implementation. This the bare minimum for these feature to work

atrosinenko:
Please excuse me for huge delay. I have stripped some of complicated logic.

For now, I have left lowercasing logic as-is , but changed other beautifying to simply the following:
* split lower-cased names by uniqueness of their last parts (simple names)
* a) unique names replace by lower-cased simple names
* b) non-unique names replace by lower cased fully qualified names with `.` replaced by `-`

I cannot prove right now that lower-casing of fully qualified class names is reversible (that is, cannot introduce name clashes), but names that differ just in letter case are already clashed on Windows. Supposing all lower-cased qualified names are distinct, this trivial disambiguation seems to not introduce clashes:
* the part of resulting name after the last `_` (if any) is generated from the simple class name, so no clashes for names from a) among them and with names from b) (even from the default package)
* names from b) are all distinct because they are just lower-cased qualified names and we supposed they are all distinct :)

muuki88:
No need to apologise :smile: . I have to thank you for keeping up with all my change requests :heart: 

The changes are looking maintainable (great docs, thank you :+1: ) and reasonable.
I have only a last request :) :) :)
Could you add a small section to the [java app multiple apps documentation](https://github.com/sbt/sbt-native-packager/blob/master/src/sphinx/archetypes/java_app/index.rst#multiple-applications) that the start scripts are disambiguated. Ideally with a couple of examples?

atrosinenko:
I will try to make requested fixes soon.

I spotted that name lower-casing logic can introduce clashes, at least in some corner cases like `Test1Class` and `Test1class` will both end up in `test-1-class` (for this particular case there is already clash for case-insensitive file system). I want to show warnings in such cases. I also want to slightly factor out common logic from `BatStartScriptPlugin` and `BashStartScriptPlugin`, so it will be easier to implement this warning.

atrosinenko:
But since this logic just inserts some `-`s and changes characters to lower case, it seems to introduce new clashes only if there are already some "case-insensitive clashes" like `pkg1.A` and `pkg1.a` (or class names contain a `-` :) ). Meanwhile, subsequent disambiguation logic can introduce new clashes, if class names contain a `_`, but it is too many coincidences.

muuki88:
> But since this logic just inserts some -s and changes characters to lower case, it seems to introduce new clashes only if there are already some "case-insensitive clashes" like pkg1.A and pkg1.a (or class names contain a - :) ).

True, and you are right that these are too many coincidences ( or rather obscure class namings ).
So keeping things simple is a better option :)

atrosinenko:
I have added a note to the docs about name disambiguation and added a warning on script name conflicts (although, I did not test the warning itself at all, only added a test on duplication detection). When adding a warning, I refactored `BashStartScriptPlugin` and `BatStartScriptPlugin` and I hope they still work :) -- `^validate` and `^validateOSX` pass on my Ubuntu, but I did not tested that these scripts works. I can test it in real life setting next week.

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/CommonStartScriptGenerator.scala
```diff
@@ -0,0 +1,137 @@
...
+                                     mainClasses: Seq[String]): (File, String) = {
+    val template = resolveTemplate(config.templateLocation)
+    val replacements = createReplacementsForMainScript(mainClass, mainClasses, config)
+    val scriptContent = TemplateWriter.generateScript(template, replacements, eol, keySurround)
+    val scriptNameWithSuffix = mainScriptName(config)
+    val script = targetDir / scriptTargetFolder / scriptNameWithSuffix
+    IO.write(script, scriptContent)
+    // TODO - Better control over this!
+    if (makeScriptsExecutable)
+      script.setExecutable(true)
```
muuki88: You you simply do
```scala
script.setExecutable(makeScriptsExecutable)
```

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/CommonStartScriptGenerator.scala
```diff
@@ -0,0 +1,137 @@
...
+      case (qualifiedClassName, scriptNameWithoutSuffix) =>
+        val scriptName = scriptNameWithoutSuffix + scriptSuffix
+        val file = tmp / scriptName
+
+        val replacements = Seq("startScript" -> executableScriptName, "qualifiedClassName" -> qualifiedClassName)
+        val scriptContent = TemplateWriter.generateScript(forwarderTemplate, replacements, eol, keySurround)
+
+        IO.write(file, scriptContent)
+        if (makeScriptsExecutable)
+          file.setExecutable(true)
```
muuki88: You you simply do
```scala
script.setExecutable(makeScriptsExecutable)
```

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/CommonStartScriptGenerator.scala
```diff
@@ -0,0 +1,137 @@
...
+  protected[this] trait ScriptConfig {
+    val executableScriptName: String
+    val scriptClasspath: Seq[String]
+    val replacements: Seq[(String, String)]
+    val templateLocation: File
+
+    def withScriptName(scriptName: String): SpecializedScriptConfig
+  }
+
+  protected[this] type SpecializedScriptConfig <: ScriptConfig
```
muuki88: What's the purpose of this type?

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/CommonStartScriptGenerator.scala
```diff
@@ -0,0 +1,137 @@
...
+    */
+  protected[this] val keySurround: String => String
+
+  protected[this] val makeScriptsExecutable: Boolean
+
+  protected[this] def createReplacementsForMainScript(mainClass: String,
+                                                      mainClasses: Seq[String],
+                                                      config: SpecializedScriptConfig): Seq[(String, String)]
+
+  protected[this] trait ScriptConfig {
```
muuki88: I like binding the script config type to the actual instance.

src/sphinx/archetypes/java_app/index.rst
```diff
@@ -223,6 +223,48 @@ For two main classes ``com.example.FooMain`` and ``com.example.BarMain`` ``sbt s
 
 Now you can package your application as usual, but with multiple start scripts.
 
+A note on script names
```
muuki88: well done :+1:

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/CommonStartScriptGenerator.scala
atrosinenko: Yes, assuming `script.setExecutable(false)` is no-op and works well on Windows :) Meanwhile, does building zips with executable bits set on Windows works at all?

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/CommonStartScriptGenerator.scala
atrosinenko: Some template methods (or how they call it?) can depend on fields from the specialized type. Particularly, it was the way I allow `Replacements.appDefines()` to access `configLocation` and `extraDefines` from specialized config.

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/CommonStartScriptGenerator.scala
atrosinenko: Of course, I don't hesitate to simplify this line, if it is possible.

atrosinenko:
Forgot to say, there were some hardcoded `bin/`s and `scripts/`s in the paths for main (and, probably, forwarder) scripts, sometimes script was saved to `scripts/` and the second element of tuple referenced `bin/`. I'm not sure I got it right, but for now I simply changed it to `scriptTargetFolder` (`== "bin/"`) everywhere.

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/CommonStartScriptGenerator.scala
muuki88: Good point. No idea what happens on windows. We can leave this as is and should put a comment in there that `setExecutable` may behave differently on different platforms.

muuki88:
All tests seem to pass, so the `bin` refactoring worked ( and is quite nice ). 
As mentioned in the comment above we can leave the `setExecutable` code as is, but should put an inline comment why.

Then this is good to merge :heart:

atrosinenko:
I have added an explanation for the `SpecializedScriptConfig` type and changed executable bit logic to unconditionally setting some particular value, as you proposed first, because it is simpler and, after all, it is worth to not only set an executable bit on shell script, but also to **explicitly unset** it on .bat script. I don't know how this behaves on Windows, but AFAIK the shell scripts were previously generated when building on Windows too, weren't they? If so, this should not error out, but will fail to set that bit if I get the documentation on `setExecutable(...)` right (it can throw an exception, but it is when `SecurityManager` disallows an operation).

muuki88:
Nice nice nice :sunglasses: 

I retriggered the one build that timed-out. If this is green, then we can merge this.
I'm very happy about the output. Thanks for all your time and patience :relaxed:

atrosinenko:
Thanks for all your time and assistance! :)

PR #1024, opened 2017-08-28
Load variables in /etc/default/{{env_file}}; pass on to process

### LINKED ISSUE #1023: Upstart init script does not source environment variables, nor does it pass environment variables on
## Expected behaviour

When I generate an upstart template, I would expect that the environment variables specified in `/etc/default/{my-service-name}` would be sources and used by Upstart.

## Actual behaviour

`/etc/default/{my-service-name}` is ignored entirely. Further, the `sudo` command, by default, resets all environment variables, so even if any were specified in the init file, they would be ignored.

## Information

- What sbt-native-packager are you using: `1.2.2`
- What sbt version: `0.13.x`
- What is your build system (e.g. Ubuntu, MacOS, Windows, Debian ): `Ubuntu`
- What package are you building (e.g. docker, rpm, ...): `debian`
- What version has your build tool (find out with e.g. `rpm --version`)
- What is your target system (e.g. Ubuntu 16.04, CentOS 7): `Ubuntu 14.04`

timcharper:
Fixes #1023

Note that auto-export is not enabled, so vars must be exported as they
are with SystemV.


timcharper:
Thinking about this, we should probably enable auto-export (`set -a`). This is least likely to result in confusion.

muuki88:
Thanks for fixing this :heart_eyes: 

> Thinking about this, we should probably enable auto-export (set -a). This is least likely to result in confusion.

Would this replace your change in this pull request?

timcharper:
The default Upstart template for /etc/default/{{env_file}} actually says to use export... so maybe it's fine.

If `set -a` were sensible, then it wouldn't replace the patch, it'd be an addition.

muuki88:
I'll be on vacation un 20.10. Thanks for your understanding if this can't be merged until then

kardapoltsev:
I agree that we need to fix behaviour of `/etc/default/app_name` across all loaders. 
@timcharper are you going to fix a test?

muuki88:
If I get this right then `set -a` would be applied in the start script and has the effect that all variables exported there (also those being source from `/etc/default/*`) are available in the called commands, which is currently not the case.

> -a Mark variables and function which are modified or created for export to the environment of subsequent commands

If that's thte case then we should add this.

@kardapoltsev It would be awesome if you could take this pull request and add the necessary changes in a new pull request.
Thanks @timcharper for your initial work and all the input. I'm sorry for not being responsive all the time. I hope this isn't the reason you stopped working on this :)

PR #1038, opened 2017-09-27
Add recipe for setting umask

### LINKED ISSUE #1032: Provide means to set umask of user?
It would be great if there was some means of setting the umask of a user created by an RPM build via native-packager. Obviously this can be done with some other provisioning tool but this strikes me as a relatively common requirement that might be worth incorporating directly in native-packager?

keirlawson:
Closes #1032


src/sphinx/recipes/umask.rst
```diff
@@ -0,0 +1,20 @@
+.. _umask:
+
+Setting the umask for your process
+==================================
+
+When using a linux package such as a RPM or Deb, the umask, associated with the user running the process, 
+will determine what permissions files generated by the process, such as log file, will have.  Usually the system default umask
+for daemons is fine and can be left unchanged, however sometime the need arises to oerride it.
```
muuki88: IMHO we should add a hint that this requires the `JavaAppServerPackaging`

src/sphinx/recipes/umask.rst
```diff
@@ -0,0 +1,20 @@
+.. _umask:
+
+Setting the umask for your process
+==================================
+
+When using the :ref:`java-server-plugin`, the umask, associated with the user running the process, 
+will determine what permissions files generated by the process, such as log file, will have.  Usually the system default umask
+for daemons is fine and can be left unchanged, however sometime the need arises to oerride it.
```
erickpintor: I think there is a typo in "oerride".

muuki88: Good catch! 😘

mkurz:
Added docs for a more general way how to set umask, not just when using System V/systemd: #1274

PR #1042, opened 2017-10-05
Windows batch script improvements

### LINKED ISSUE #688: Universal java options ignored on Windows - cannot set javaOptions
I have added an option for max heap space:

```
javaOptions in Universal += "-J-Xmx1024m"
```

This goes into `conf/application.ini` when packaging and is properly parsed on Linux and Mac from the bash script. On Windows, the `.bat` file is broken, though. Instead of trying to read this file, it looks for `./<APP-NAME>_config.txt` which does not exist, resulting in an application launch with wrong heap settings.

It also seems that the format of the `.ini` file is not compatible. If I copy the `application.ini` file to `<APP-NAME>_config.txt`, the `.bat` file has this error:

```
Unrecognized option: -J-Xmx1024m
```

dwickern:
I added some functions to the batch script to mirror the bash script. This means you can set extra java args using similar syntax:
```sbt
batScriptExtraDefines += """call :add_java "-Dconfig.file=%APP_HOME%\conf\production.conf""""
bashScriptExtraDefines += """addJava "-Dconfig.file=${app_home}/../conf/production.conf""""
```

Previously the batch script used a dynamic environment variable name `<APP_NAME>_HOME`, which is difficult to use from the sbt build. Now it's called `APP_HOME` always.

The bat script will still read `<APP_NAME>_HOME` from the environment for backwards compatibility.

### application.ini

The bat script now supports `javaOptions in Universal` via a generated `application.ini` file. Fixes #688

The ini file uses the `-J` syntax, so you can use the same ini file for both bat and bash, or you can have different files by specifying `batScriptConfigLocation` and/or `bashScriptConfigLocation`.

The bat script will still read `<APP_NAME>_config.txt` for backwards compatibility.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/ApplicationIniGenerator.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/ApplicationIniGenerator.scala
@@ -5,6 +5,10 @@ import java.io.File
 import sbt._
 
 trait ApplicationIniGenerator {
+  /**
+    * @return the existing mappings plus a generated application.ini
+    *         if custom javaOptions are specified
+    */
   def generateApplicationIni(universalMappings: Seq[(File, String)],
                              javaOptions: Seq[String],
                              bashScriptConfigLocation: Option[String],
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/ApplicationIniGenerator.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/ApplicationIniGenerator.scala
@@ -1,0 +1,40 @@
+package com.typesafe.sbt.packager.archetypes.scripts
+
+import java.io.File
+
+import sbt._
+
+trait ApplicationIniGenerator {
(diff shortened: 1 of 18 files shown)


muuki88:
Thanks a lot for your work 😍

I'll review this in depth, when I'm back from my vacation ( ~22.10).

src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/ApplicationIniGenerator.scala
```diff
@@ -0,0 +1,40 @@
+package com.typesafe.sbt.packager.archetypes.scripts
+
+import java.io.File
+
+import sbt._
+
+trait ApplicationIniGenerator {
+  def generateApplicationIni(universalMappings: Seq[(File, String)],
```
muuki88: We should add some scaladoc here:

```
Returns the existing mappings with a generated application.ini if
not already provided by the user.
```

dwickern:
Sure I'll help maintain this, although I can't make a big time commitment

PR #1044, opened 2017-10-15
Add support for --chown flag for ADD/COPY Docker commands

### LINKED ISSUE #1029: Add support for --chown flag for ADD/COPY Docker commands
As `--chown` for ADD/COPY has landed in moby here: https://github.com/moby/moby/pull/34263 I think it would be awesome if the following could be done:

 - check docker version using `sys.process.Process` (something a la `docker version --format '{{.Server.Version}}'` and if version is applicable then use `--chown` approach in order to save that notorious duplicate layer that doubles up on application size.

@muuki88 WDYT?

mrfyda:
Closes #1029


lightbend-cla-validator:
Hi @mrfyda,

muuki88:
Thanks a lot for taking the time to work on this. I'll review this when I'm back from vacation  ( 22.10 ) 😍

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -179,18 +185,34 @@ object DockerPlugin extends AutoPlugin {
...
     /**
       * This is the file path of the file in the Docker image, and does not depend on the OS where the image
       * is being built. This means that it needs to be the Unix file separator even when the image is built
       * on e.g. Windows systems.
       */
     val files = dockerBaseDirectory.split(UnixSeparatorChar)(1)
-    Cmd("ADD", s"$files /$files")
+
+    dockerVersion match {
+      case Some(DockerVersion(major, minor, _, _)) if major >= 17 && minor >= 9 =>
```
muuki88: The `if` statement looks buggy to me. 
If the server has version `18.0` it will yield false, which is not want we want, right?
It must be something like `if majro >= (17 && minor >= 9) || major > 17`

At this point I'm fine with putting this into a if statement. If we have more of this special case, we should
extract this in somethin like a `DockerSupport` object:

```scala
object DockerSupport {

   def chownFlag(version: DockerVersion): Boolean = ...

}

```

muuki88:
I updated the branch so all tests should work fine now.

mrfyda:
@muuki88 Nice catch. Fixed the condition and extracted it to a separate object.

Also, updated both the spotify docker client implementation and the docs.

ahjohannessen:
Thanks @mrfyda 💪🤗

ahjohannessen:
@muuki88 Please cut a release :)

muuki88:
Sure sure sure :stuck_out_tongue_winking_eye: 

Merged one last PR today. With sbt cross build and all scripted tests a release always
takes a while, but here it is: `1.3.0`

ahjohannessen:
👍

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -80,6 +83,9 @@ object DockerPlugin extends AutoPlugin {
     dockerEntrypoint := Seq("bin/%s" format executableScriptName.value),
     dockerCmd := Seq(),
     dockerExecCommand := Seq("docker"),
+    dockerVersion := Try(Process(dockerExecCommand.value ++ Seq("version --format '{{.Server.Version}}'")).!!)
```
NeQuissimus: Guys, I don't think this works...
I believe it needs to be `Seq("version", "--format", "'{{.Server.Version}}'")`, otherwise your shell will add another set of ticks around the whole thing and Docker will report that it does not know the command.

NeQuissimus: nvm me, already fixed somewhere else :D

NeQuissimus: #1052 and @mrfyda ftw! :D

PR #1056, opened 2017-10-25
Fix duplicate application.ini mappings when both BASH and BAT plugins are used

dwickern:
Now that the BAT plugin supports application.ini, when you use both plugins together you get duplicate mappings:
```
[info] * (target/universal/tmp/conf/application.ini,conf/application.ini)
[info] * (target/universal/tmp/conf/application.ini,conf/application.ini)
```

This isn't necessarily a problem, but unzipping the universal dist will try to create the file twice:
```
$ unzip app-1.0-SNAPSHOT.zip
Archive:  app-1.0-SNAPSHOT.zip
  inflating: app-1.0-SNAPSHOT/conf/application.ini
replace app-1.0-SNAPSHOT/conf/application.ini? [y]es, [n]o, [A]ll, [N]one, [r]ename:
```

This change filters out the duplicate mapping. Alternatively maybe it would be better to have an `ApplicationIniPlugin` which is a dependency of both the `BatStartScriptPlugin` and `BashStartScriptPlugin`

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/ApplicationIniGenerator.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/ApplicationIniGenerator.scala
@@ -21,17 +21,25 @@ trait ApplicationIniGenerator {
           val pathMapping = cleanApplicationIniPath(location)
           //Do not use writeLines here because of issue #637
           IO.write(configFile, ("# options from build" +: javaOptions).mkString("\n"))
-          val hasConflict = universalMappings.exists {
-            case (file, path) => file != configFile && path == pathMapping
+          val filteredMappings = universalMappings.filter {
+            case (`configFile`, `pathMapping`) =>
+              // ignore duplicate application.ini mappings with identical contents
+              // for example when using both the BASH and BAT plugin
+              false
+
+            case (_, `pathMapping`) =>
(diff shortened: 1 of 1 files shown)


src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/ApplicationIniGenerator.scala
```diff
@@ -21,17 +21,25 @@ trait ApplicationIniGenerator {
...
-            case (file, path) => file != configFile && path == pathMapping
-          }
-          // Warn the user if he tries to specify options
-          if (hasConflict) {
-            log.warn("--------!!! JVM Options are defined twice !!!-----------")
-            log.warn(
-              "application.ini is already present in output package. Will be overridden by 'javaOptions in Universal'"
-            )
+          val filteredMappings = universalMappings.filter {
+            case (`configFile`, `pathMapping`) =>
```
muuki88: Interesting implementation. I didn't get at first what was happening here. The second call to the method
will filter the already created config file.

muuki88:
nevermind the revert branch. I overlooked that the validateBash tests were failing, but it was a travis issue.

muuki88:
See https://travis-ci.org/sbt/sbt-native-packager/builds/292847041

PR #1058, opened 2017-10-26
Convert dockerVersion from a setting to a task

dwickern:
I get errors when starting sbt when the machine can't connect to docker
```
An error occurred trying to connect: Get http://127.0.0.1:2375/v1.22/version: dial tcp 127.0.0.1:2375: connectex: No connection could be made because the target machine actively refused it.
```

Since fetching the docker version involves starting a process and/or making an HTTP request, it seems like it should be a task rather than a setting.


muuki88:
@dwickern interested in getting commit rights? :heart_eyes:

dwickern:
@muuki88 Sure, you can add me

PR #1062, opened 2017-11-04
Remove redundant Dockerfile

### LINKED ISSUE #1060: Remove redundant docker file
## Expected behaviour

`sbt docker:stage` generates only one docker file.

## Actual behaviour

Two files are generated:

```
tree target/docker
target/docker/
├── Dockerfile <---- here
└── stage
    ├── Dockerfile  <---- here
    └── opt
        └── docker
            ├── bin
            │   ├── docker-test
            │   └── docker-test.bat
            └── lib
                ├── docker-test.docker-test-0.1.0.jar
```

This causes a lot of confusion as in #1006

mrfyda:
Fixes #1060

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -112,10 +112,6 @@ object DockerPlugin extends AutoPlugin {
     Seq(
       executableScriptName := executableScriptName.value,
       mappings ++= dockerPackageMappings.value,
-      mappings ++= {
-        val baseDir = target.value
-        Seq(dockerGenerateConfig.value) pair (file => IO.relativize(baseDir, file))
-      },
       name := name.value,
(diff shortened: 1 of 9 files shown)


src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -144,14 +140,17 @@ object DockerPlugin extends AutoPlugin {
         }
       },
       sourceDirectory := sourceDirectory.value / "docker",
-      stage := Stager.stage(Docker.name)(streams.value, stagingDirectory.value, mappings.value),
+      stage := {
+        dockerGenerateConfig.value
```
muuki88: I would rather declare this dependency with

```scala
stage := (stage dependsOn dockerGenerateConfig).value
```

I'm not sure if you have to scope the right hand side `stage` with `in Docker` or not. The `inConfig` semantics always confuse me.

PR #1066, opened 2017-11-13
Add resolver information

calvinlfer:
Without the resolver, you cannot download version 1.3.2 on Scala 2.12.4 and SBT 1.0.3


muuki88:
Thanks for providing a PR  for this. This is a bug in sbt ( sbt/sbt/issues/3432 )
and a workaround is provided in the mentioned ticket above.

PR #1068, opened 2017-11-16
Add support for -X JVM options

malliina:
If you specify JVM options starting with "-X" in `javaOptions in Universal`, such as

    -Xbootclasspath/p:/usr/share/appname/lib/org.mortbay.jetty.alpn.alpn-boot-8.1.11.v20170118.jar

Then the bash script will append the option too late in the java command, and it won't work. This commit fixes it. Similar to https://github.com/sbt/sbt-native-packager/pull/431.

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/bash-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/bash-template
@@ -194,7 +194,7 @@ process_args () {
 
      -java-home) require_arg path "$1" "$2" && jre=`eval echo $2` && java_cmd="$jre/bin/java" && shift 2 ;;
 
- -D*|-agentlib*) addJava "$1" && shift ;;
+ -D*|-X*|-agentlib*) addJava "$1" && shift ;;
             -J*) addJava "${1:2}" && shift ;;
               *) addResidual "$1" && shift ;;
     esac


muuki88:
Thanks for your pull request.

IIRC you can append JVM parameters by starting with `-J`, e.g. `-J-Xmx512mb`. If this is not working for you then we should open a bug to fix this.

malliina:
Thanks for the suggestion, I tried appending `-J`, however it didn't work. I believe in this case the example `-X` parameter above needs to be treated like app parameters similar to `-D` params.

What does work is the following:

    bashScriptExtraDefines += s"""addJava "-Xbootclasspath/p:/usr/share/app/lib/file.jar""""

I interpreted from the docs that an equivalent way to define that is:

    javaOptions in Universal += "-Xbootclasspath/p:/usr/share/app/lib/file.jar"

or then

    javaOptions in Universal += "-J-Xbootclasspath/p:/usr/share/app/lib/file.jar"

However, neither of the `javaOptions ...` approaches work in this case (but maybe the -J params do work for JVM memory settings). Effectively the desired `java` run command would look like:

    java -Xbootclasspath/p:/usr/share/appname/lib/file.jar -Dexample.prop=example.value -Dother=values and so on...

And that's only the case for me if I add the parameter using `bashScriptExtraDefines` above. Well, since the `bashScriptExtraDefines` approach works for me I'll survive, but I thought this was quite confusing to configure as it is.

dwickern:
Looking at the code, `-J-X` should be equivalent to `addJava -X` :thinking:

malliina:
Yeah you are correct, sorry for wasting your time with this PR. I had a problem in my deployment process where changes to JVM parameters didn't propagate properly in all cases.

muuki88:
No worries ☺️

PR #1075, opened 2017-11-30
Update release version in download instructions

agemooij:


dwickern:
Thanks! I've been meaning to write a sbt-release step to update the readme automatically. It's so easy to forget

PR #1085, opened 2018-01-01
"" value for default docker label values

kimxogus:
If docker label's value is null, `NullPointerException` is thrown currently.
To prevent that, this pr will put `""` as default value for label values.
`""` for empty label values seems ok according to [value guidelines in official docker doc](https://docs.docker.com/engine/userguide/labels-custom-metadata/#manage-labels-on-objects)

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -172,7 +172,11 @@ object DockerPlugin extends AutoPlugin {
     */
   private final def makeLabel(label: (String, String)): CmdLike = {
     val (variable, value) = label
-    Cmd("LABEL", variable + "=\"" + value.toString + "\"")
+    val valueString = Option(value) match {
+      case Some(v) => v.toString
+      case None => ""
+    }
+    Cmd("LABEL", variable + "=\"" + valueString + "\"")
   }
 
   /**


dwickern:
How does it get null in the first place?

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -172,7 +172,11 @@ object DockerPlugin extends AutoPlugin {
     */
   private final def makeLabel(label: (String, String)): CmdLike = {
     val (variable, value) = label
-    Cmd("LABEL", variable + "=\"" + value.toString + "\"")
+    val valueString = Option(value) match {
+      case Some(v) => v.toString
+      case None => ""
```
muuki88: An empty label seems not to be a good default. IMHO we should throw a
helpful error message here, because something else is configured incorrectly.

kimxogus:
I'm using sbt-git plugin and put label with git related values like
```scala
dockerLabels := Map(
  "gitHeadCommit" -> git.gitHeadCommit.orNull
)
```
It works ok in my personal machine, but it throws NullPointerExcpetion in circleci docker container.
I checked git installed and git commands works perfectly in circleci environment, but it failed.
I resolved that replacing it getting from circleci environment variables, but I want to make default value set for empty values.

I agree with @muuki88 , then what about giving warning messages for null value?

muuki88:
Sorry @kimxogus  for the late reply.

Using `null` is not idiomatic scala, so I wouldn't merge this change.
If a value is optional it's declare as such (with `Option[T]`). If you
pass in a `null` somewhere then you should expect nullpointer exceptions.

PR #1018, opened 2017-08-24
fix packageZipTarball defaults for UniversalDocs and UniversalSource

timcharper:
The implementation clearly fell short of the intention.

fixes #1019

PR #1086, opened 2018-01-01
fix packageZipTarball defaults for UniversalDocs and UniversalSource

muuki88:
Resurrects #1018


src/sbt-test/universal/test-zips/test
```diff
@@ -9,10 +9,18 @@ $ exists target/universal-src/simple-test-0.1.0.zip
 # Run the tgz packaging.
 > universal:packageZipTarball
 $ exists target/universal/simple-test-0.1.0.tgz
+> universal-docs:packageZipTarball
+$ exists target/universal-docs/simple-test-0.1.0.tgz
+> universal-src:packageZipTarball
```
muuki88: These commands don't seem to work at all.

Issue #1106, opened 2018-03-04
Deb package dependencies


GreyCat:
It caught my attention that .deb packages which sbt-native-packager generates lacks any dependencies, even on the JRE (i.e. something like `default-jre` or may be even something that states minimal version needed, as, for example, nowadays Scala-build .class files require Java >=8 due to 52.0 class version).

## Expected behaviour

`apt-get install my-package` installs all required stuff to make `my-package` build by sbt-native-packager work on Debian/Ubuntu.

## Actual behaviour

Package installs, but trying to run it yields

```
No java installations was detected.
Please go to http://www.java.com/getjava/ and download
```

## Information

- What sbt-native-packager are you using: 1.3.2
- What sbt version: 1.1.1
- What is your build system (e.g. Ubuntu, MacOS, Windows, Debian ): Debian
- What package are you building (e.g. docker, rpm, ...): deb
- What version has your build tool (find out with e.g. `rpm --version`): 
- What is your target system (e.g. Ubuntu 16.04, CentOS 7): Debian / Ubuntu

muuki88:
Thanks for your report 😀

We haven't set any defaults, because there is no good one IMHO. Packages and versions can vary in a wide range. This would mean we would break a lot of packages just by wrong defaults. You can [add dependencies](http://sbt-native-packager.readthedocs.io/en/latest/formats/debian.html#settings) if you like, but you don't have to.

GreyCat:
Could you provide an example of what could break exactly?

As far as I see, all Debian/Ubuntu repositories supply multiple JRE packages, but due to convention, all of them provide `java-runtime` package, or, if we'll need minimum Java 8, then `java8-runtime`. Even if one downloads an Oracle-supplies SDK and installs it using [make-jpkg from java-package](https://wiki.debian.org/JavaPackage), it would still have these `java-runtime`, `java5-runtime`, `java6-runtime`, `java7-runtime` and `java8-runtime` in Provides.

The only concern I've found is that there are two "flavours" — `javaX-runtime` vs `javaX-runtime-headless`. This one should be decided by package's creator, based on his requirements (i.e. whether the application requires AWT / Swing / fancy UI stuff or not).

I understand that it's important to not break existing user's installations, but may be we could at least recommend adding `debianPackageDependencies := "java8-runtime-headless"` or `debianPackageDependencies := "java8-runtime"` in the docs, explaining the difference? I don't see any downsides of that.

muuki88:
[...]

PR #1107, opened 2018-03-05
JRE dependencies in Debian documentation

GreyCat:
Added some documentation that recommends usage of JRE dependencies in Debian packages, and explains how to configure them properly.

Follows up discussion in https://github.com/sbt/sbt-native-packager/issues/1106.


muuki88:
Awesome :heart: 
Thanks a lot

PR #1108, opened 2018-03-15
Fix docker server version parsing

jalaziz:
Add support for numbers in the release version. This allows version
detection to parse server version of the form '18.03.0-ce-rc3'.


muuki88:
Awesome 😎 thanks for your contribution.

This is good to merge after the build is green. Had to restart one job.

PR #1109, opened 2018-03-19
Read CMD value from Docker config rather than defaulting to empty seq…

fionabianchisky:
…uence

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -81,7 +81,7 @@ object DockerPlugin extends AutoPlugin {
     ),
     dockerUpdateLatest := false,
     dockerEntrypoint := Seq(s"${(defaultLinuxInstallLocation in Docker).value}/bin/${executableScriptName.value}"),
-    dockerCmd := Seq(),
+    dockerCmd := (dockerCmd in Docker).value,
     dockerExecCommand := Seq("docker"),
     dockerVersion := Try(Process(dockerExecCommand.value ++ Seq("version", "--format", "'{{.Server.Version}}'")).!!).toOption
       .map(_.trim)
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "1.3.4-SNAPSHOT"
+version in ThisBuild := "1.3.5-SNAPSHOT"


lightbend-cla-validator:
Hi @fionabianchisky,

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "1.3.4-SNAPSHOT"
+version in ThisBuild := "1.3.5-SNAPSHOT"
```
muuki88: This not necessary :smiley: The release process updates this file during a release.

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -81,7 +81,7 @@ object DockerPlugin extends AutoPlugin {
     ),
     dockerUpdateLatest := false,
     dockerEntrypoint := Seq(s"${(defaultLinuxInstallLocation in Docker).value}/bin/${executableScriptName.value}"),
-    dockerCmd := Seq(),
+    dockerCmd := (dockerCmd in Docker).value,
```
muuki88: This change does not work as `dockerCmd in Docker` is undefined. You can see the
error in the travis log outputs

```
docker:dockerCmd from *:dockerCmd ((com.typesafe.sbt.packager.docker.DockerPlugin) DockerPlugin.scala:84)
```

fionabianchisky:
Hiya,

I have to admit I’m rather new to Scala so apologies for anything dumb in the PR.  Here we wanted to be able to set a value for the CMD value in our Dockerfile’s – to add arguments for a the script that starts Java.  On trying to set “dockerCmd” in the build.sbt settings nothing seemed to get passed through.  It *looks* from my (almost certainly flawed) reading of the code that dockerCmd is always left empty.

I got the email about the build failure but didn’t have time to investigate further – when I can “sbt test” locally everything seemed to pass – I also tested the local copy of the plugin I built and it seemed to work.  I’ll recheck my local stuff tomorrow.

Thanks,

Fiona

muuki88:
No worries :smile: We'll figure out what's wrong. There are only  two options. Either a configuration error in your build or a bug in sbt-native-packager :wink: 

> we wanted to be able to set a value for the CMD value in our Dockerfile’s – to add arguments for a the script that starts Java. 

The `dockerCmd` should be the correct option.

> On trying to set “dockerCmd” in the build.sbt settings nothing seemed to get passed through

Can you post the section of your `build.sbt` that configures the `dockerCmd`? It should look something like this

```scala
dockerCmd ++= Seq("arg1", "arg2")
```

> It *looks* from my (almost certainly flawed) reading of the code that dockerCmd is always left empty.

The `DockerPlugin` only initializes the settings. As the `dockerCmd` is a list, we initialize it with an empty list. In your `build.sbt` or another plugin you can append things to this list.

The sbt documentation gives a decent overview about the concepts of settings/tasks:
- [Task & Settings](https://www.scala-sbt.org/1.x/docs/Task-Graph.html)
- [Append values to a setting](https://www.scala-sbt.org/1.x/docs/Appending-Values.html)

PR #1111, opened 2018-03-23
fixed Java version check in bash template

jubecker:
The current Java version check in 'bash-template' relies on a string comparison for numbers.
As far as I could investigate this check fails at least for bash 4.3.48 on a xenial/sid based docker images.

It fails for the latest JDK release 10, as "10" is not greater than "1.6" when compared as string in bash.
Later bash version seem to treat this check differently.

I changed the version check to compare numbers. For versions 1.x.y, the check extracts the minor version "x" and compares agains "6". If the major version is unequal to "1", the major version is compared against "6".

Using docker openjdk images I added scripted test.

Tested with bash 4.3.48 and 4.4.19, docker 17.12.0-ce.

### DIFF
--- a/src/sbt-test/docker/jdk-versions/build.sbt
+++ b/src/sbt-test/docker/jdk-versions/build.sbt
@@ -1,6 +1,6 @@
 val basename = "jdk-versions"
 
-scalacOptions in ThisBuild := Seq("-target:jvm-1.8")
+scalacOptions in ThisBuild in (Compile, compile) := Seq("-target:jvm-1.8")
 
 lazy val `jdk8` = project
   .in(file("jdk8"))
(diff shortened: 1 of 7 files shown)


src/sbt-test/docker/jdk-versions/build.sbt
```diff
@@ -0,0 +1,31 @@
+val basename = "jdk-versions"
+
+scalacOptions in ThisBuild := Seq("-target:jvm-1.8")
```
muuki88: it seems the scala doc compiler doesn't like this setting. I never actually fought
the docs compiler (only my colleagues), so sorry for not being a greater help here :disappointed: 

IIRC you can either turn the scaladoc generation off or do some research on how
to fix this :sunglasses: I'm fine with either way :+1:

jubecker: Thanks for pointing this out. Will push a fix. Normally limiting the scope should fix this.

muuki88: Awesome, thanks :)

lightbend-cla-validator:
Hi @jubecker,

muuki88:
Awesome :)

PR #1112, opened 2018-03-23
Process -XX arguments as JVM args

longshorej:
`-XX...` arguments now added to the JVM. Without this, the following fails:

```sbt
javaOptions in Universal ++= Seq(
"-XX:+UnlockExperimentalVMOptions",
"-XX:+UseCGroupMemoryLimitForHeap"
)
```

```bash
Bad root server path: /opt/docker/-XX:+UnlockExperimentalVMOptions
```

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/bash-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/bash-template
@@ -194,9 +194,9 @@ process_args () {
 
      -java-home) require_arg path "$1" "$2" && jre=`eval echo $2` && java_cmd="$jre/bin/java" && shift 2 ;;
 
- -D*|-agentlib*) addJava "$1" && shift ;;
-            -J*) addJava "${1:2}" && shift ;;
-              *) addResidual "$1" && shift ;;
+ -D*|-agentlib*|-XX*) addJava "$1" && shift ;;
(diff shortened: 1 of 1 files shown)


muuki88:
Thanks a lot for your pull request :heart_eyes: 

To make your example work you would have to prepend `-J` to the `-XX` parameters.

```scala
javaOptions in Universal ++= Seq(
  "-J-XX:+UnlockExperimentalVMOptions",
  "-J-XX:+UseCGroupMemoryLimitForHeap"
)
```

However my understanding is that this `-J` separation was introduced to allow
a separation between JVM and application arguments. The `-XX` arguments should
always be jvm args, so IMHO this pull request makes it a bit less confusing for new comers :)

longshorej:
Ah, thanks for that tip @muuki88, that works nicely.

guizmaii:
Hi,

I don't understand.
What's the difference between using the previously accepted format:
```
javaOptions in Universal ++= Seq(
  "-J-XX:+UnlockExperimentalVMOptions",
  "-J-XX:+UseCGroupMemoryLimitForHeap"
)
```

and the new one:
```
javaOptions in Universal ++= Seq(
  "-XX:+UnlockExperimentalVMOptions",
  "-XX:+UseCGroupMemoryLimitForHeap"
)
```

Where and why should I prefer the new one ?

PR #1131, opened 2018-05-23
fix procedure syntax

xuwei-k:


muuki88:
Thanks for the code cleanup :)

PR #1138, opened 2018-07-12
[Docker] add multiple docker tags support

kimxogus:
Added additional docker image tags support.

Before, if your project's version is `1.1.0` and you want to tag `1.1` and `1`, you had to tag `1.1` and `1` manually outside of sbt task. With this PR, you can specify additional docker image tags and manage them in sbt.


kimxogus:
That sounds much better. That would be more intuitive. I'll change them

kimxogus:
- Updated `DockerAlias` with more readable member methods
- Updated `sbt-scalafmt` because it reformats differently with Intellij's scalafmt.

src/main/scala/com/typesafe/sbt/packager/docker/DockerAlias.scala
```diff
@@ -10,11 +10,18 @@ package com.typesafe.sbt.packager.docker
...
   */
 case class DockerAlias(registryHost: Option[String], username: Option[String], name: String, tag: Option[String]) {
-  protected val untagged = registryHost.map(_ + "/").getOrElse("") + username.map(_ + "/").getOrElse("") + name
 
-  /** Tag with (optional) given version */
-  val versioned = untagged + tag.map(":" + _).getOrElse("")
+  def withRegistryHost(registryHost: Option[String]): DockerAlias = copy(registryHost = registryHost)
 
-  /** Tag with version 'latest' */
-  val latest = s"$untagged:latest"
```
dwijnand: RIP binary API...

muuki88: Really time to add MiMa :(

PR #1161, opened 2018-09-25
FIX #1097 Put jdeb back in provided scope

### LINKED ISSUE #1097: Put jdeb back in provided scope
## Expected behaviour

Adding the plugin with no configuration at all does not break the build when using Coursier.

## Actual behaviour

#1053 happens out of the box with sbt-native-packager 1.3.2 and Coursier 1.0.

## Information

- sbt-native-packager 1.3.2
- sbt version 1.1.0
- sbt-coursier 1.0.0

jdeb is in provided scope for sbt 0.13, but that is currently disabled with a comment that something was broken about it in sbt 1.0 milestones (build used RC3 at the time the comment was added).

I've created this issue to track following up on that—presumably if jdeb is back in provided scope for sbt 1.x, a user won't have the bad experience of hitting #1053 and having to go discover the coursier/coursier#450 workaround as soon as they enable sbt-native-packager on their project, when they are not even using `JDebPackaging`.

This is a poor out-of-the-box experience, so I feel an issue should be kept open for it.

I don't know what the problem with provided was, but I can try to see if it still persists with later sbt 1.x.

muuki88:
This should fix #1097 and reduce some of the dependency issues in builds that don't require debian packaging.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -29,7 +29,7 @@ libraryDependencies ++= {
         // these dependencies have to be explicitly added by the user
         // FIXME temporary remove the 'provided' scope. SBT 1.0.0-M6 changed the resolving somehow
         "com.spotify" % "docker-client" % "8.9.0" /* % "provided" */,
-        "org.vafer" % "jdeb" % "1.3" /*% "provided"*/ artifacts Artifact("jdeb", "jar", "jar")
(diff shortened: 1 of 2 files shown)


timcharper:
👏 👏

PR #1166, opened 2018-10-20
Add Travis tests for the GraalVL native-image plug-in

ScalaWilliam:
Also move out the sbt downloader into the same directory as the
downloader.


.travis.yml
```diff
@@ -16,6 +16,7 @@ cache:
   - $HOME/.sbt
   - $HOME/.m2
   - $HOME/.ivy2
+  - $HOME/graalvm-ce-1.0.0-rc7
```
muuki88: rc8 has been released two days ago :smile:

.travis.yml
```diff
@@ -35,7 +36,8 @@ before_install:
     fi
 
 before_script:
-- curl -L --create-dirs -o $HOME/.sbt/launchers/1.2.6/sbt-launch.jar http://dl.bintray.com/sbt/maven-releases/org/scala-sbt/sbt-launch/1.2.6/sbt-launch.jar
+- .travis/download-sbt
+- . .travis/download-graalvm
```
muuki88: Why two dots?

muuki88:
No need for squashing 😉 GitHub has a button for this 😎

.travis.yml
muuki88: Thanks for the explanation 🤗 too long bash abstinence 😁
So this is needed in order to make the export work, right?

muuki88:
You can merge after CI passed 👍

.travis.yml
ScalaWilliam: Indeed. There could be other ways but this is the neatest.

```bash
$ bash <(echo export x=y)
$ echo $x

$ source <(echo export x=y)
$ echo $x
y
$
```

PR #1168, opened 2018-10-26
Improve the GraalVM native-image documentation

ScalaWilliam:
1. I was unable to see the documentation in the table of contents
2. GraalVM native-image plugin should be in the diagram
3. Added quickstart steps for GraalVM by reusing our Travis installer.

May need to update the URL inside of it depending on whether there're changes in #1166 - best merged only after.


muuki88:
Nice 😍

PR #1172, opened 2018-11-05
Update scaladoc in contentOf to match method signature

frosforever:
This PR updates the scaladoc for `contentOf` method to use the same signature as the code.


frosforever:
@muuki88 Thanks for your work maintaining native-packager!

PR #1177, opened 2018-11-28
WindowsPlugin: support multiple wxs sources (#1176)

### LINKED ISSUE #1176: WindowsPlugin: support multiple .wxs inputs
This would be useful when working with custom WiX scripts - for example, when harvesting a directory using `heat`. Since `heat` generates a separate `.wxs` file, the only option to use it with the current `packageBin` implementation would be transforming it to an include file, which is incredibly ugly, and would require working around the copy logic.

Both `candle` and `light` support using multiple input files, so implementing this feature can be as simple as replacing `wixFile: TaskKey[File]` with `wixFiles: TaskKey[Seq[File]]`, and updating `packageBin` implementation accordingly. If there's a need to preserve compatibility with the code that uses `wixFile`, we can keep `wixFile` and just do this:
```scala
wixFiles := List(wixFile.value)
```

nigredo-tori:
Closes #1176 

Solved this, as described in the original issue, by providing a `wixFiles` task. All the mentioned files are copied to `target/windows` in a flat structure. A more idiomatic solution would probably be to somehow add `managedSourceDirectories`/`unmanagedSourceDirectories`  mechanics to the `Windows` configuration - however, I have neither time nor energy to dig deeper into this at the moment.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/windows/WindowsPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/windows/WindowsPlugin.scala
@@ -106,12 +106,9 @@ object WindowsPlugin extends AutoPlugin {
       // to our target directory.
       val targetFlat: Path.FileMap = Path.flat(target.value)
       val wsxFiles = wsxSources.map(targetFlat(_).get)
-      IO.copy(
-        wsxSources.zip(wsxFiles).filter {
-          case (src, dest) => src.getAbsolutePath != dest.getAbsolutePath
-        },
-        CopyOptions().withOverwrite(true)
-      )
+      IO.copy(wsxSources.zip(wsxFiles).filter {
+        case (src, dest) => src.getAbsolutePath != dest.getAbsolutePath
+      }, CopyOptions().withOverwrite(true))
       IO.copy(for ((f, to) <- mappings.value) yield (f, target.value / to))
 
       // Now compile WIX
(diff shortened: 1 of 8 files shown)


lightbend-cla-validator:
Hi @nigredo-tori,

src/main/scala/com/typesafe/sbt/packager/windows/WindowsPlugin.scala
```diff
@@ -96,27 +96,45 @@ object WindowsPlugin extends AutoPlugin {
...
-      // First we have to move everything (including the wix file) to our target directory.
-      val wix = target.value / (name.value + ".wix")
-      if (wixFileValue.getAbsolutePath != wix.getAbsolutePath) IO.copyFile(wixFileValue, wix)
+
+      // First we have to move everything (including the WIX scripts)
+      // to our target directory.
+      val targetFlat: Path.FileMap = Path.flat(target.value)
+      val wsxFiles = wsxSources.map(targetFlat(_).get)
+      val wsxCopyPairs = wsxSources.zip(wsxFiles).filter {
+        case (src, dest) => src.getAbsolutePath != dest.getAbsolutePath
```
muuki88: This filter would be that `wsxSources` may reference things from `target` ?

nigredo-tori: Yes. The same check is present in the existing code, and I assumed that someone might be relying on that behavior.

muuki88:
`v1.3.15` release is on its way.

Issue #1178, opened 2018-12-10
Can't specify Epoch for the RPM Plugin


Ophirr33:
## Expected behaviour

There is a setting available to specify the Epoch of a generated rpm when using the RPM Plugin

## Actual behaviour

Rpms are always built with an epoch of 0

kardapoltsev:
Closed via #1179

PR #1179, opened 2018-12-10
 #1178 attempt at adding in rpm epochs

Ophirr33:


src/main/scala/com/typesafe/sbt/packager/rpm/RpmMetadata.scala
```diff
@@ -222,6 +223,9 @@ case class RpmSpec(meta: RpmMetadata,
     sb append ("Version: %s\n" format meta.version)
     sb append ("Release: %s\n" format meta.release)
     sb append ("Summary: %s\n" format meta.summary)
+
+    if (meta.epoch > 0) sb append ("Epoch: %d\n" format meta.epoch)
```
kardapoltsev: I think it's better to use `Option` in such cases

Ophirr33: I went back and forth on that. Doing this way also protests against negative epochs, but I suppose I could do an exists or filter for the negative check. Thoughts?

kardapoltsev:
Yes, sounds good.

-- 
Regards, Alexey

src/main/scala/com/typesafe/sbt/packager/rpm/RpmMetadata.scala
muuki88: That sounds good. Option + check.

You can also add a [validation check](https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/rpm/RpmPlugin.scala#L105) for this a log an error or warning.

Would `rpmLint` catch this?

integration-tests-ansible/test-project-play-rpm/packaging.sbt
```diff
@@ -28,3 +28,5 @@ rpmRelease := "1"
 rpmVendor := "DemoVendor"
 
 rpmLicense := Some("Apache-2.0")
+
+rpmEpoch := 1
```
muuki88: Out of curiosity? This do you acutally run this test? I haven't used this for a very long time and I'm not sure if we should remove it.

Ophirr33: Whoops, I forgot that I committed this. I mucked around with the Vagrant file (e.g., changing chef to bento) and got something half running, but tbh you could probably remove it.

src/main/scala/com/typesafe/sbt/packager/rpm/RpmMetadata.scala
Ophirr33: I don't think rpmLint has a check for this

Ophirr33:
squashed and updated with review comments

Ophirr33:
Can someone poke the travis build?

src/main/scala/com/typesafe/sbt/packager/validation/package.scala
```diff
@@ -63,4 +63,16 @@ package object validation {
     }
   }
 
+  def epochIsNaturalNumber(epoch: Int): Validation.Validator = () => {
+    sys.error(s"Passed: $epoch")
```
kardapoltsev: Did you forget to remove logging?

kardapoltsev:
LGTM

Ophirr33:
🥂

PR #1188, opened 2019-01-11
Fix #1185 Add MiMa to check binary compatibility

### LINKED ISSUE #1185: Add MiMa to check binary compatibility
We should avoid binary incompatible changes.

https://github.com/lightbend/migration-manager

muuki88:
Fixes #1185


muuki88:
Thanks for taking a look :heart:

Issue #1187, opened 2019-01-10
Docker/stage generates different Dockerfile depending on docker version


eed3si9n:
## steps

```
$ eval $(minikube docker-env)
$ sbt
> Docker/stage
```

## expected

`Docker/stage` either fails or generates an identical `Dockerfile` with or without running `eval $(minikube docker-env)`.

```dockerfile
ADD --chown=daemon:daemon opt /opt
```

## actual

The content of Dockerfile is different.

```dockerfile
ADD opt /opt
RUN ["chown", "-R", "daemon:daemon", "."]
```

This causes deployment to fail on OpenShift.

## notes

Ref https://github.com/sbt/sbt-native-packager/pull/1044

https://github.com/sbt/sbt-native-packager/blob/v1.3.15/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala#L226-L230

```scala
    if (dockerVersion.exists(DockerSupport.chownFlag)) {
      Seq(Cmd("ADD", s"--chown=$daemonUser:$daemonGroup $files /$files"))
    } else {
      Seq(Cmd("ADD", s"$files /$files"), makeChown(daemonUser, daemonGroup, "." :: Nil))
    }
```

It turns out that when I run `eval $(minikube docker-env)` it was detecting "17.12.1-ce".

muuki88:
Thanks for the detailed bug report :heart: 

From your description it looks like you have two very different docker versions on your host and in Kubernetes. If you want to support a fixed version of docker you can set the `dockerVersion` setting.

```scala
dockerVersion := Some("10.0.0")
```

or set it to `None` to use no newer features. Maybe we should point that out in the docs explicitly. 

I'm not sure what happens if you build docker images with different docker daemon version. Are they backwards compatible? What's the error on open shift. From my little docker experience I thought the `Dockerfile` is only relevant during build time.

eed3si9n:
I think it makes sense for the build to occasionally outsource its work to external system like `docker`, but as a general principle the build should be repeatable.
I suggest we create a setting for `dockerCompatibilityVersion` and set it to whatever the version that introduced `--chown`. If we detect any version below it we can error out asking the user to either bump their Docker or lower the `dockerCompatibilityVersion` to `"17.0"`.

> What's the error on open shift. From my little docker experience I thought the Dockerfile is only relevant during build time.

On OpenShift, I get `/opt/docker/bin/name-of-app: Permission denied`.

[...]

Issue #883, opened 2016-09-26
Docker Image 'chown -R daemon:daemon' duplicates application layer


gavares:
sbt-native-packager versions 1.2.0-M5 and 1.1.4

Affer adding application contents to a docker image, another layer is created with command:

```
chown -R daemon:daemon .
```

This touches every file in the application and creates new layer that has the same size as the application. In my case, after adding ~130mb layer containing my app, another 130mb layer is added as a result of touching every file in the my app layer.

`docker history` shows:
![image](https://cloud.githubusercontent.com/assets/46012/18822590/7d4770f8-8366-11e6-8bb0-34a2d59924d9.png)

Wondering if this is necessary and/or if it could be done before adding the files from the application layer to the container.

gavares:
Seems this is caused by a very annoying docker issue that has had years of discussions and github issues opened but still hasn't been fixed.

The current workaround is to perform chown/chmod locally, create a tar of the contents and then ADD the tar which docker will magically unpack, preserving permissions of the contents.

Not sure if there will be any issues getting this to work since a user (probably) won't have the `daemon` user and group locally on their dev machine and, even if they did, the uid/gid probably wouldn't match what is present in the container anyway.

muuki88:
There have already been some discussions on [gitter](https://gitter.im/sbt/sbt-native-packager) on this topic ( I couldn't find unfortunately ).

For docker we have some options to optimize this (e.g. splitting the app jars from the lib jars to enable caching). However this needs time, testing and verification. So if you have good solution with proper research we I'm happy to discuss this on a pull request from you :)

cc @fiadliel

NeQuissimus:
You'll want to follow docker/docker#30110.

If you don't have volumes, it should be reasonably safe to just remove the `USER` and `chown` from the build instructions and thereby save the space. That will, however, cause your applications to be `root`, albeit inside a container.

fiadliel:
Thanks, I'll keep an eye on it, an alternative is the "squash" command (when it becomes stable).

But I would disagree that it's safe to remove USER/chown - the proof lies in decades of escapes from chroots on Unix systems (and Docker containers are just chroots on steroids).

NeQuissimus:
Personally, I am using https://github.com/goldmann/docker-squash, which gets rid of the "root-owned files"-layer altogether.

ivantopo:
[...]

PR #1190, opened 2019-01-18
Implement dockerPermissionStrategy

### LINKED ISSUE #1189: OpenShift compatibility
OpenShift is a flavor of Kubernetes maintained by Red Hat, which has bit more tighter security defaults and requires some unique settings, described in [Guidelines](https://docs.okd.io/latest/creating_images/guidelines.html#openshift-specific-guidelines).

It would be good if sbt-native-packager could generate Docker container that is compatible with both vanilla Kubernetes and OpenShift for `daemonUser`.

### current

```Dockerfile
FROM openjdk:8
WORKDIR /opt/docker
ADD --chown=daemon:daemon opt /opt
USER daemon
ENTRYPOINT ["/opt/docker/bin/hello"]
CMD []
```

### OpenShift compatible Dockerfile

The following Dockerfile implements non-root container, but with OpenShift compatibility:

```Dockerfile
FROM openjdk:8
WORKDIR /opt/docker
RUN useradd --system --create-home --uid 1001 --gid 0 nonroot
ADD opt /opt
RUN chgrp -R 0 /opt && chmod -R g=u /opt
USER 1001
ENTRYPOINT ENTRYPOINT ["/opt/docker/bin/hello"]
```

eed3si9n:
Fixes #1189

This implements a non-root Docker container that's safer by default and compatible with [Red Hat OpenShift](https://www.openshift.com/learn/what-is-openshift/).

Current `ADD --chown=daemon:daemon opt /opt` and `USER daemon` combination nominally implements non-root image, but by giving ownership of the working directory to the `daemon` user, it reduces the safety.
Instead we should use `chmod` to default to read-only access unless the build user opts into writable working directory.

The challenge is calling `chmod` without incurring the fs layer overhead (#883).
[Multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) can be used to pre-stage the files with desired file permissions.

This adds new `dockerPermissionStrategy` setting which decides how file permissions are set for the working directory inside the Docker image generated by sbt-native-packager. The strategies are:

- `DockerPermissionStrategy.MultiStage` (default): uses multi-stage Docker build to call chmod ahead of time.
- `DockerPermissionStrategy.None`: does not attempt to change the file permissions, and use the host machine's file mode bits.
- `DockerPermissionStrategy.Run`: calls `RUN` in the Dockerfile. This has regression on the resulting Docker image file size.
- `DockerPermissionStrategy.CopyChown`: calls `COPY --chown` in the Dockerfile. Provided as a backward compatibility.

### Working directory is read-only by default

For `MultiStage` and `Run` strategies, `dockerChmodType` is used in addition to call `chmod` during Docker build.

- `DockerChmodType.UserGroupReadExecute` (default): chmod -R u=rX,g=rX
- `DockerChmodType.UserGroupWriteExecute`: chmod -R u=rwX,g=rwX
- `DockerChmodType.SyncGroupToUser`: chmod -R g=u
- `DockerChmodType.Custom`: Custom argument provided by the user.

### How to opt-in to writable working directory

Some application will require writing files to the working directory.
In that case the setting should be changed as follows:

```scala
import com.typesafe.sbt.packager.docker.DockerChmodType

dockerChmodType := DockerChmodType.UserGroupWriteExecute
```

### Docker package validation

During `docker:stage`, Docker package validation is called to check if the selected strategy is compatible with the detected Docker version.
This fixes the current repeatability issue reported as #1187. If the incompatibility is detected, the user is advised to either upgrade their Docker, pick another strategy, or override the `dockerVersion` setting.

### Use gid=0

`daemonGroup` is set to `root` instead of copying the value from the `daemonUser` setting.
This matches the semantics of `USER` as well as OpenShift, which uses gid=0.


muuki88:
Thanks for the pull request @eed3si9n . I'll take a deeper look when I have some time. Hopefully this week.

eed3si9n:
Cool. Let me squash the intermediate commits with failing tests.

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -66,6 +66,11 @@ object DockerPlugin extends AutoPlugin {
 
   override def projectConfigurations: Seq[Configuration] = Seq(Docker)
 
+  override lazy val globalSettings: Seq[Setting[_]] = Seq(
```
muuki88: I wonder if we should start using `globalSettings` at this point. I would favor a separate PR that moves the `projectSettings` defaults into the `globalSettings`. 

Or we add a small comment that we are now starting [to follow sbt best practices as documented](https://www.scala-sbt.org/release/docs/Plugins-Best-Practices.html#Provide+default+values+in)

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -153,16 +187,22 @@ object DockerPlugin extends AutoPlugin {
       stagingDirectory := (target in Docker).value / "stage",
       target := target.value / "docker",
       daemonUser := "daemon",
-      daemonGroup := daemonUser.value,
+      daemonGroup := "root",
       defaultLinuxInstallLocation := "/opt/docker",
+      validatePackage := Validation
+        .runAndThrow(validatePackageValidators.value, streams.value.log),
```
muuki88: Good catch! Thanks for fixing this.

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -153,16 +187,22 @@ object DockerPlugin extends AutoPlugin {
...
         filesExist((mappings in Docker).value),
         validateExposedPorts(dockerExposedPorts.value, dockerExposedUdpPorts.value),
-        validateDockerVersion(dockerVersion.value)
+        validateDockerVersion(dockerVersion.value),
+        validateDockerPermissionStrategy(dockerPermissionStrategy.value, dockerVersion.value)
       ),
       dockerPackageMappings := MappingsHelper.contentOf(sourceDirectory.value),
-      dockerGenerateConfig := generateDockerConfig(dockerCommands.value, stagingDirectory.value)
+      dockerGenerateConfig := {
+        val _ = validatePackage.value
```
muuki88: What are your thoughts on this pattern? I usually tend to do something like

```sbt
dockerGenerateConfig := { /* implementation */ },
// side effects
dockerGenerateConfig := (dockerGenerateConfig dependsOn validatePackage).value
```

src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala
```diff
@@ -42,3 +43,10 @@ trait DockerKeys {
 
   val dockerCommands = TaskKey[Seq[CmdLike]]("dockerCommands", "List of docker commands that form the Dockerfile")
 }
+
+// Workaround to pass mima.
+// In the next version bump we should hide DockerKeys trait to package private.
```
muuki88: Thanks for adding this comment :heart:

src/sbt-test/docker/file-permission/build.sbt
```diff
@@ -0,0 +1,96 @@
+lazy val checkDockerfile = taskKey[Unit]("")
```
muuki88: Very good test coverage :+1: 

Could you rename the `tasks` and `sbt` files to match what they test? e.g.

- `checkDockerfile` to `checkDockerfileDefaults`
- `checkDockerfile2` to `checkDockerfileWithPermissionsNone`

and also the sbt files `a.sbt` to `permissionsNone.sbt`. :smile:

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
eed3si9n: I figured we can start at least for new ones with pure values. I'll add the comment there.

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
eed3si9n: I think of `dependsOn` as part of sbt 0.12 style used together with `<<=` etc, since with `.value` you can just call it like I did.

src/sbt-test/docker/file-permission/build.sbt
eed3si9n: ok. implemented this in ef77500.

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
muuki88: Thanks for sharing :smile:

muuki88:
I always use the _Squash and merge_ button from github :grin:

muuki88:
Documentation in the `docker.rst` can be added in a separate PR.

aoboturov:
Hi, could anyone explain, how to make non-entrypoint executables executable?

here's an example of the `ls -lah`:
image

I tried different combinations of permission strategies and chmodtypes to no avail.

Issue #1198, opened 2019-02-06
Dockerfile switches to daemon user uid instead of the daemonUser


kunalkapoor:
With the addition of the new docker permission model to use Multi-stage dockerfiles (https://github.com/sbt/sbt-native-packager/commit/34c776dc868ca1bad546131c7433b38d74680adf), the user switching does not work as expected.

## Expected behaviour

Earlier, sbt docker:stage used to do the following

```
val user = (daemonUser in Docker).value
makeUser(user)
```

This creates the following Dockerfile

```
USER daemon
```

There was no check to ensure the "daemon" user exists or not.

## Actual behaviour

Now, the code does this

```
val user = (daemonUser in Docker).value
val uid = 1001
makeUserAdd(user, uid, gid)
makeUser(uid)
```

`makeUserAdd` adds the "daemon" user if it doesn't exist and then the `makeUser` switches to that user. However, if the "daemon" user does exist and with a different uid than that hardcoded 1001, then the `makeUser` switches to a uid that doesn't exist in the system.

The Dockerfile created looks like below

```
RUN id -u daemon || useradd --system --create-home --uid 1001 --gid 0 daemon
USER 1001
```

As you can see above, if the `id -u daemon` part returns a 0 exit code, then the daemon user already exists and is not created. In this case, if the daemon user has a uid other than 1001, then `USER 1001` essentially switches to a user that does not exist.

## Solution

1. The `makeUser(uid)` code in `stage1` needs to be `makeUser(user)` to ensure we switch to the `daemon` user and not be dependent on the uid.
2. We should make the `uid = 1001` a configurable parameter in `build.sbt` (just like `daemonUser`).

## Information

- What sbt-native-packager are you using
1.13.17
- What sbt version
1.2.8
- What is your build system (e.g. Ubuntu, MacOS, Windows, Debian )
MacOS and CentOS
- What package are you building (e.g. docker, rpm, ...)
sbt docker:stage
- What is your target system (e.g. Ubuntu 16.04, CentOS 7)
MacOS and CentOS 7

muuki88:
Thanks for the detailed report and the proposed PR. I'll take a look as soon as possible.

PR #1199, opened 2019-02-08
Fix USER switching in Docker

kunalkapoor:
Refers #1198.

In attempting to switch to the `daemonUser` we are instead switching to the hardcoded uid 1001. If the daemon user exists with a different uid then the docker container default user is incorrect. Correcting the `makeUser` function to switch to the username specified by `daemonUser` instead of the static uid.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -148,7 +148,7 @@ object DockerPlugin extends AutoPlugin {
         dockerEnvVars.value.map(makeEnvVar) ++
         makeExposePorts(dockerExposedPorts.value, dockerExposedUdpPorts.value) ++
         makeVolumes(dockerExposedVolumes.value, user, group) ++
-        Seq(makeUser(uid), makeEntrypoint(dockerEntrypoint.value), makeCmd(dockerCmd.value))
+        Seq(makeUser(user), makeEntrypoint(dockerEntrypoint.value), makeCmd(dockerCmd.value))
 
       stage0 ++ stage1
     }


muuki88:
cc @jw3

muuki88:
The `docker-filepermissions` test failed: 
https://travis-ci.org/sbt/sbt-native-packager/jobs/490338861#L1258

can you have a look? If you need help running the tests take a look at the [CONTRIBUTING.MD|https://github.com/sbt/sbt-native-packager/blob/master/CONTRIBUTING.md]

jw3:
Sounds like an issue in the category of working on OpenShift vs working in other places.  OpenShift doesn't need the user to exist at all, but that doesn't usually play well elsewhere.

@kunalkapoor a reference to a public Docker image that demonstrates the issue out of the box would help.

eed3si9n:
I used numeric UID in #1190 based on the OKD Creating Images [Guidelines](https://docs.okd.io/latest/creating_images/guidelines.html#openshift-specific-guidelines). (OKD, The Origin Community Distribution, is an open source fork of Kubernetes by Red Hat, which powers OpenShift):

> Lastly, the final `USER` declaration in the `Dockerfile` should specify the user ID (numeric value) and not the user name.
> ...
> If your S2I image does not include a `USER` declaration with a numeric user, your builds will fail by default.

Having said that, I think the issue brought up by @kunalkapoor in #1198 is totally fair, and I don't think non-OpenShift users would have to suffer from any drawbacks. I continue to seek for the middle ground where the Docker image is friendly to OpenShift by default.

Also, it seems the numeric `USER` requirement is limited to S2I (Source-To-Image), which is used by subset of OpenShift users, so perhaps this could be an opt-in feature. Also to avoid possible conflict with base image maybe we should consider changing from "daemon" to something more unique like "demiourgos".

kunalkapoor:
From what I gather, this issue is unrelated to the docker image used. You can try to reproduce this using any docker image as long as you have a "daemon" user already created in that image with a uid that is not 1001. Or else, we could use any random image and add docker commands to create the "daemon" user before sbt-native-packager docker commands are added.

An example excerpt of a build.sbt:
```
val userCreationCommands = Seq(Cmd("useradd --system --create-home --uid 6 --gid 0 daemon"))
dockerCommands := userCreationCommands ++ dockerCommands
```

Will result in a Dockerfile like below:
```
# The custom command to create the daemon user
RUN useradd --system --create-home --uid 1001 --gid 0 daemon

# The commands added by sbt-native-packager
USER 1001
```

The above results in a daemon user with uid 6 being created but it switches to uid 1001 which doesn't exist in the system.

If we must switch to uid for OpenSwitch users, then can we dynamically determine the uid of the daemon user and use that instead if it already exists? I'm not sure if that's possible. But an alternative might be to expose the `uid` as a configurable parameter (just like `daemonUser`).

eed3si9n:
There's a setting called `daemonUserUid`. I will send a PR to use that.

eed3si9n:
Here's my proposal - https://github.com/sbt/sbt-native-packager/pull/1200

kunalkapoor:
Closing this as it is not the correct solution. #1200 seems better suited to fix the problem.

PR #1200, opened 2019-02-08
Use daemonUserUid to opt-out of numeric USER

### LINKED ISSUE #1198: Dockerfile switches to daemon user uid instead of the daemonUser

eed3si9n:
Fixes #1198

Numeric USER directive is now controlled using `daemonUserUid in Docker`, which defaults to `Some("1001")`.

To get back to previous behavior the following can be used:

```scala
daemonUserUid in Docker := None
daemonUser in Docker    := "daemon"
```

### DIFF
--- a/src/sbt-test/docker/file-permission/build.sbt
+++ b/src/sbt-test/docker/file-permission/build.sbt
@@ -20,9 +20,10 @@ lazy val root = (project in file("."))
           |RUN ["chmod", "-R", "u=rX,g=rX", "/opt/docker"]
           |
           |FROM fabric8/java-centos-openjdk8-jdk
-          |RUN id -u daemon || useradd --system --create-home --uid 1001 --gid 0 daemon
+          |USER root
+          |RUN id -u demiourgos728 2> /dev/null || useradd --system --create-home --uid 1001 --gid 0 demiourgos728
           |WORKDIR /opt/docker
-          |COPY --from=stage0 --chown=daemon:root /opt/docker /opt/docker
+          |COPY --from=stage0 --chown=demiourgos728:root /opt/docker /opt/docker
           |USER 1001
           |ENTRYPOINT ["/opt/docker/bin/file-permission-test"]
           |CMD []""".stripMargin.linesIterator.toList)
@@ -33,7 +34,8 @@ lazy val root = (project in file("."))
       val lines = dockerfile.linesIterator.toList
       assertEquals(lines,
         """FROM fabric8/java-centos-openjdk8-jdk
-          |RUN id -u daemon || useradd --system --create-home --uid 1001 --gid 0 daemon
+          |USER root
+          |RUN id -u demiourgos728 2> /dev/null || useradd --system --create-home --uid 1001 --gid 0 demiourgos728
           |WORKDIR /opt/docker
           |COPY opt /opt
           |USER 1001
@@ -46,7 +48,8 @@ lazy val root = (project in file("."))
       val lines = dockerfile.linesIterator.toList
       assertEquals(lines,
         """FROM openjdk:8
-          |RUN id -u daemon || useradd --system --create-home --uid 1001 --gid 0 daemon
+          |USER root
+          |RUN id -u demiourgos728 2> /dev/null || useradd --system --create-home --uid 1001 --gid 0 demiourgos728
           |WORKDIR /opt/docker
           |COPY opt /opt
           |RUN ["chmod", "-R", "u=rX,g=rX", "/opt/docker"]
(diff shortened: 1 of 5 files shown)


src/sbt-test/docker/file-permission/changes/strategy-copychown.sbt
```diff
@@ -2,3 +2,7 @@ import com.typesafe.sbt.packager.docker._
 
 dockerPermissionStrategy := DockerPermissionStrategy.CopyChown
 dockerBaseImage          := "fabric8/java-centos-openjdk8-jdk"
+
+// opt-out of numeric USER
+daemonUserUid in Docker  := None
+daemonUser in Docker     := "daemon"
```
kunalkapoor: Shouldn't this be demiourgos728 too?

kunalkapoor:
Thanks for fixing this!

src/sbt-test/docker/file-permission/changes/strategy-copychown.sbt
eed3si9n: It's demiourgos728 by default, but here I am demonstrating that it can be changed back to daemon if someone is using that from base image etc.

nafg:
Isn't UserUid repetive ("user user identity") -- I'd rather daemonUid vs. daemonUser

nafg:
Couldn't it be a single setting using Either or something?

nafg:
Couldn't it be a single setting using Either or something?

eed3si9n:
The key existed since https://github.com/sbt/sbt-native-packager/pull/449.

src/sbt-test/docker/file-permission/build.sbt
```diff
@@ -79,9 +81,10 @@ lazy val root = (project in file("."))
           |RUN ["chmod", "-R", "u=rwX,g=rwX", "/opt/docker"]
           |
           |FROM fabric8/java-centos-openjdk8-jdk
-          |RUN id -u daemon || useradd --system --create-home --uid 1001 --gid 0 daemon
+          |USER root
+          |RUN id -u demiourgos728 2> /dev/null || useradd --system --create-home --uid 1001 --gid 0 demiourgos728
           |WORKDIR /opt/docker
-          |COPY --from=stage0 --chown=daemon:root /opt/docker /opt/docker
+          |COPY --from=stage0 --chown=demiourgos728:root /opt/docker /opt/docker
```
muuki88: Any particular reason for this name :laughing:

eed3si9n: In Platonism, dēmiourgos (δημιουργός) meaning craftsman, was an artisan-like figure responsible for creating / maintaining the physical universe. At some point this word got misread by someone around 350AD, got mixed up with 'daimon', and out of thin air became Demogorgon. https://twitter.com/eed3si9n/status/1094012513627394049
728 is 'sbt' spelled using telephone keypad. Hopefully this is unique enough name.

PR #1201, opened 2019-02-08
Run chmod +x explicitly in stage0

eed3si9n:
This is on top of https://github.com/sbt/sbt-native-packager/pull/1200

Using sbt-native-packager 1.3.17 with OpenShift, I still ran into some permission issues and needed to call chmod +x explicitly in the container. This will emit chmod +x for all bash scripts that was generated.

```Dockerfile
RUN ["chmod", "u=+x,g=+x", "/opt/docker/bin/file-permission-test"]
```


src/main/scala/com/typesafe/sbt/packager/docker/DockerPermissionStrategy.scala
```diff
@@ -61,6 +68,13 @@ object DockerChmodType {
     def argument: String = "u=rwX,g=rwX"
   }
 
+  /**
+    * Gives +x permission to users and groups.
+    */
+  case object UserGroupPlusExecute extends DockerChmodType {
+    def argument: String = "u=+x,g=+x"
```
eed3si9n: This is a mistake. It should be u+x,g+x.

eed3si9n:
@muuki88 Fixed a bug that the tests caught, and also updated the docs.

src/sphinx/formats/docker.rst
```diff
@@ -128,11 +128,16 @@ Environment Settings
...
-    * ``DockerChmodType.SyncGroupToUser``: chmod -R g=u
+    * ``DockerChmodType.UserGroupReadExecute`` (default): chmod u=rX,g=rX
+    * ``DockerChmodType.UserGroupRead``: chmod u=r,g=r
+    * ``DockerChmodType.UserGroupWriteExecute``: chmod u=rwX,g=rwX
+    * ``DockerChmodType.SyncGroupToUser``: chmod g=u
+    * ``DockerChmodType.UserGroupPlusExecute``: chmod u+x,g+x (This is for ``dockerAdditionalPermissions``)
     * ``DockerChmodType.Custom``: Custom argument provided by the user.
 
+  ``dockerAdditionalPermissions``
+    Additional permissions typically used to give ``chmod +x`` rights for the executable files. By default generated Bash scripts are given ``DockerChmodType``.
```
olivierlemasle: Missing permissions?

muuki88: Added this via the UI :+1: thanks for reviewing this as well :hug

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -108,6 +108,16 @@ object DockerPlugin extends AutoPlugin {
     },
     dockerRmiCommand := dockerExecCommand.value ++ Seq("rmi"),
     dockerBuildCommand := dockerExecCommand.value ++ Seq("build") ++ dockerBuildOptions.value ++ Seq("."),
+    dockerAdditionalPermissions := {
+      val scripts = makeBashScripts.value
```
muuki88: I overlooked that. This causes a circular dependency between `JavaAppPackaging` and the `DockerPlugin`, which produces a bug reported in #1206

PR #1209, opened 2019-03-14
Update Proguard example

ipostanogov:
Now it works with sbt-proguard 0.3.0.

`proguardOptions` were taken from [their example](https://github.com/sbt/sbt-proguard/blob/v0.3.0/src/sbt-test/proguard/simple/build.sbt).

#518 #721


muuki88:
Thanks for keeping the docs up-to-date :heart: 

@kardapoltsev feel free to merge :+1:

PR #1211, opened 2019-03-27
Don't write RPM standard output as error

Falmarri:
Currently stdout for building an RPM gets output at error level. 

```
[error] + umask 022
[error] + cd /home/david/test/target/rpm/BUILD
[error] + /bin/rm -rf /home/david/test/target/rpm/buildroot
[error] + exit 0
```

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/rpm/RpmHelper.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/rpm/RpmHelper.scala
@@ -105,7 +105,8 @@ object RpmHelper {
         else Seq.empty
       ) ++ Seq(spec.meta.name + ".spec")
       log.debug("Executing rpmbuild with: " + args.mkString(" "))
-      (sys.process.Process(args, Some(specsDir)) ! log) match {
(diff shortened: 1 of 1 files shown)


kardapoltsev:
Hi! Changes look good to me. Could you fix scalafmt errors please? (check travis build)

Falmarri:
Sure. I did this in the github UI 😊

muuki88:
@Falmarri thanks for fixing this! :heart_eyes:  I lived under the impression that `log` would already do this an `rpm` simply emits to `stderr`

anilkumarmyla:
does this work? Just tried this on linux (`rpm-build-4.11.3-35.el7.x86_64`) & mac (`rpm 4.14.2.1`), but the [error] log is still present

Falmarri:
Hmm. it's possible that I incorrectly identified the problem and that rpm outputting to stderr is actually the root cause. I'll inspect the rpm command more

Falmarri:
OK you guys are right. It does write the output to stderr regardless. How about this? https://github.com/sbt/sbt-native-packager/pull/1212

PR #1212, opened 2019-03-30
Log rpm output to error or info depending on exit code

Falmarri:


anilkumarmyla:
@Falmarri did you get to test this change?

Falmarri:
I didn't test it by actually running it. But I don't see how this couldn't do what is desired 😨

anilkumarmyla:
It does work @Falmarri . Did a `sbt publishLocal` and tested resulting version in a project which logged all rpm build to info instead of error

anilkumarmyla:
any actions on this?

PR #1214, opened 2019-04-09
correct sbt console example

mcanlas:


muuki88:
Thanks for keeping the docs up-to-date :heart:

PR #1216, opened 2019-04-19
Mention discoveredMainClasses in documentation

Discipe:
Added my personal use case into docs, in case someone would try to search for it.


muuki88:
Thanks for improving sbt-native-packagers documentation :heart: 

IMHO a better place to put this information would be [in the multiple applications section](https://github.com/sbt/sbt-native-packager/blob/67f06754943e1c133358eb5e2fa65df85f21c2f7/src/sphinx/archetypes/java_app/index.rst#multiple-applications). What do you think?

PR #1220, opened 2019-05-02
Add a simple jlink wrapper

nigredo-tori:
This scans the `fullClasspath` (using `jdeps`), and builds a JVM image (using `jlink`) from all the referenced modules. The image is then "mounted" to a configurable prefix in `Universal / mappings` (`jre/` by default). The image location is then passed to the start scripts with the highest priority (higher than `JAVA_HOME`).

Potential issues (feedback required):
- I'm not sure how to wire this into the existing CI setup, since the whole thing requires Java 11.
- Java 9 support is theoretically possible, but would require some fiddling with `jdeps` output. I'm not sure it is worth the trouble.
- It might be better to only limit the image to the standard modules (`java.*`, `javafx.*`, `jdk.*`), since everything else is already provided by the `JavaAppPackaging`. I can think both of pros and cons to this - maybe this warrants adding a separate incision point for transforming the detected module list?

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/bat-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/bat-template
@@ -6,6 +6,7 @@
 @REM Configuration:
 @REM @@APP_ENV_NAME@@_config.txt found in the @@APP_ENV_NAME@@_HOME.
 @setlocal enabledelayedexpansion
+@setlocal enableextensions
 
 @echo off
 
@@ -30,12 +31,27 @@ set "CFG_FILE=%APP_HOME%\@@APP_ENV_NAME@@_config.txt"
 set CFG_OPTS=
 call :parse_config "%CFG_FILE%" CFG_OPTS
 
-rem We use the value of the JAVACMD environment variable if defined
-set _JAVACMD=%JAVACMD%
+rem We use the value of the JAVA_OPTS environment variable if defined, rather than the config.
+set _JAVA_OPTS=%JAVA_OPTS%
+if "!_JAVA_OPTS!"=="" set _JAVA_OPTS=!CFG_OPTS!
 
-if "%_JAVACMD%"=="" (
-  if not "%JAVA_HOME%"=="" (
-    if exist "%JAVA_HOME%\bin\java.exe" set "_JAVACMD=%JAVA_HOME%\bin\java.exe"
+rem We keep in _JAVA_PARAMS all -J-prefixed and -D-prefixed arguments
+rem "-J" is stripped, "-D" is left as is, and everything is appended to JAVA_OPTS
+set _JAVA_PARAMS=
+set _APP_ARGS=
+
+@@APP_DEFINES@@
+
+rem Bundled JRE has priority over standard environment variables
+if defined BUNDLED_JVM (
+  set "_JAVACMD=%BUNDLED_JVM%\bin\java.exe"
+) else (
+  if "%JAVACMD%" neq "" (
+    set "_JAVACMD=%JAVACMD%"
+  ) else (
+    if "%JAVA_HOME%" neq "" (
+      if exist "%JAVA_HOME%\bin\java.exe" set "_JAVACMD=%JAVA_HOME%\bin\java.exe"
+    )
   )
 )
 
@@ -69,18 +85,6 @@ if "%JAVAOK%"=="false" (
   if defined DOUBLECLICKED pause
   exit /B 1
 )
-
-
-rem We use the value of the JAVA_OPTS environment variable if defined, rather than the config.
-set _JAVA_OPTS=%JAVA_OPTS%
-if "!_JAVA_OPTS!"=="" set _JAVA_OPTS=!CFG_OPTS!
-
-rem We keep in _JAVA_PARAMS all -J-prefixed and -D-prefixed arguments
-rem "-J" is stripped, "-D" is left as is, and everything is appended to JAVA_OPTS
-set _JAVA_PARAMS=
-set _APP_ARGS=
-
-@@APP_DEFINES@@
 
 rem if configuration files exist, prepend their contents to the script arguments so it can be processed by this runner
 call :parse_config "%SCRIPT_CONF_FILE%" SCRIPT_CONF_ARGS
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaAppKeys.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaAppKeys.scala
@@ -26,4 +26,6 @@ trait JavaAppKeys {
     "scriptClasspath",
     "A list of relative filenames (to the lib/ folder in the distribution) of what to include on the classpath."
   )
+  val bundledJvmLocation =
+    TaskKey[Option[String]]("bundledJvmLocation", "The location of the bundled JVM image")
 }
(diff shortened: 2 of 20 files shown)


src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/bat-template
```diff
@@ -6,6 +6,7 @@
 @REM Configuration:
 @REM @@APP_ENV_NAME@@_config.txt found in the @@APP_ENV_NAME@@_HOME.
 @setlocal enabledelayedexpansion
+@setlocal enableextensions
```
muuki88: What does this imply? I'm not a windows user and have almost no experience :smile:

src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkKeys.scala
```diff
@@ -0,0 +1,21 @@
...
+package jlink
+
+import sbt._
+
+import com.typesafe.sbt.packager.archetypes.jlink._
+
+/**
+  * Available settings/tasks for the [[com.typesafe.sbt.packager.archetypes.jlink.JlinkPlugin]].
+  */
+trait JlinkKeys {
```
muuki88: We are starting to make these traits private. This helps in keeping things binary compatible

src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkPlugin.scala
```diff
@@ -0,0 +1,106 @@
...
+  override lazy val projectSettings: Seq[Setting[_]] = Seq(
+    target in jlinkBuildImage := target.value / "jlink" / "output",
+    jlinkBundledJvmLocation := "jre",
+    bundledJvmLocation := Some(jlinkBundledJvmLocation.value),
+    jlinkOptions := (jlinkOptions ?? Nil).value,
+    jlinkOptions ++= {
+      val log = streams.value.log
+      val run = runJavaTool(javaHome.in(jlinkBuildImage).value, log) _
+
+      val paths = fullClasspath.in(Compile).value.map(_.data.toString)
```
muuki88: Can we rely on the `toString()` method here? Does it produce proper paths for unix/windows systemsß

muuki88:
Adding this to the`.travis.yml` should probably work

``ỳml
    - script: sbt "^validateJdkPackagerTravis"
      name: "scripted jlink tests"
      jdk: oraclejdk11
      if: type = pull_request OR (type = push AND branch = master)
```

Maybe after https://github.com/sbt/sbt-native-packager/blob/master/.travis.yml#L77

File: src/main/scala/com/typesafe/sbt/packager/archetypes/JavaAppKeys.scala [hunk 4]
```diff
@@ -26,4 +26,6 @@ trait JavaAppKeys {
     "scriptClasspath",
     "A list of relative filenames (to the lib/ folder in the distribution) of what to include on the classpath."
   )
+  val bundledJvmLocation =
```
muuki88: This is a binary incompatible change (see build error https://travis-ci.org/sbt/sbt-native-packager/jobs/527261199#L1165).

This can either be fixed by adding a line to the `build.sbt` (here https://github.com/sbt/sbt-native-packager/blob/master/build.sbt#L60) or by a new, private trait that is being extended in the `autoImport` object. I favour the second option.

src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/bat-template
nigredo-tori: > I'm not a windows user and have almost no experience 

I also have little experience with BAT scripts - this change required some digging :smile:.

`ENABLEEXTENSIONS` explicitly enables some "newer" (as in, only present in NT-based systems) *command extensions*, like extended `IF` syntax (`IF DEFINED` later in this script). These are almost always enabled anyway, but there are ways to turn them off for compatibilty reasons.

src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkKeys.scala
nigredo-tori: Made this `private[packager]`.

src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkPlugin.scala
nigredo-tori: `File.toString` seems to produce a valid path as long as the object itself represents a valid path - see its [documentation](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#toString()). There should be no escaping issues since `Process` should not require escaping. I'm not sure what other problems we can encounter here.

src/main/scala/com/typesafe/sbt/packager/archetypes/JavaAppKeys.scala
nigredo-tori: I've done this (added a `JavaAppKeys2` private trait). However there are other bincompat-breaking changes since I had to add bundled JVM defines to the script plugins. I can handle `BashStartScriptPlugin.Defines.apply` (just add an overload and delegate to it), but I had to change the `BatStartScriptPlugin.BatScriptConfig` - and it looks like working around this one would just bring more breaking changes in the future...

nigredo-tori:
> Adding this to the.travis.yml should probably work

Done this, but I can't check that it works because of the MiMa problem.

src/main/scala/com/typesafe/sbt/packager/archetypes/JavaAppKeys.scala
nigredo-tori: I've worked around other MiMa problems in d6b4b94 just in case - but I still think that it would be better to make a minor release with this.

src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/bat-template
muuki88: Oh wow! Thanks for clarifying this. That makes a lot of sense :smile:

src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkKeys.scala
muuki88: sweet

src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/bat-template
muuki88: Oh wow! Thanks for clarifying this. That makes a lot of sense :smile:

src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkPlugin.scala
muuki88: I would rather use `getPath()` to make this explicit. `toString` may change (as well as the docs)

nigredo-tori: Fair point. Done.

muuki88:
well :joy: , you have been faster then me writing my comment. `getPath` already fixed :+1:

muuki88:
version `1.3.21` is on its way

PR #1224, opened 2019-05-12
Update graalvm installation script

gurinderu:
-Update graalvm to version 1.19.0


gurinderu:
@muuki88 yep, it has a MacOS support.

PR #1225, opened 2019-05-14
Add jdeps options settings to the jlink plugin

vhiairrassary:
I just tested the jlink wrapper for my project but `jdeps` produced some `not found` for some classes:
```
Error: Missing dependencies from the module path and classpath.
To suppress this error, use --ignore-missing-deps to continue.

shapeless_2.12-2.3.3.jar
   shapeless.CachedImplicitMacros                     -> scala.reflect.macros.contexts.Context              not found
   shapeless.CachedImplicitMacros                     -> scala.tools.nsc.Global                             not found
   shapeless.CachedImplicitMacros                     -> scala.tools.nsc.typechecker.Analyzer               not found
   shapeless.CachedImplicitMacros                     -> scala.tools.nsc.typechecker.Contexts               not found
[...]
```

This PR adds a way to set or add options for `jdeps`. I tested on mac using java 12 and:
```
jlinkOptions ++= Seq("--no-header-files", "--no-man-pages", "--compress=2", "--strip-debug"),
jdepsOptions ++= Seq("--ignore-missing-deps")
```

/cc @nigredo-tori

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkKeys.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkKeys.scala
@@ -7,6 +7,8 @@ import sbt._
   * Available settings/tasks for the [[com.typesafe.sbt.packager.archetypes.jlink.JlinkPlugin]].
   */
 private[packager] trait JlinkKeys {
+  val jdepsOptions =
+    SettingKey[Seq[String]]("jdepsOptions", "Options for the jdeps utility")
 
   val jlinkBundledJvmLocation =
     TaskKey[String]("jlinkBundledJvmLocation", "The location of the resulting JVM image")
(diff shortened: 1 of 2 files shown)


nigredo-tori:
The actual problem with your build is that `shapeless` has a `Provided` dependency on `scala-reflect`, which is not satisfied. You should probably fix that instead, even more so since `--ignore-missing-deps` would also silence other similar errors.

That said, it seems like the behavior of `jdeps` has [changed](https://bugs.openjdk.java.net/browse/JDK-8213915) in Java 12, hence this issue. `jdepsOptions` is fine as a stopgap, but it ties the build to a specific JDK version. This setting also seems like an easy way to inadvertedly break things (by losing `--print-module-deps`).

Ideally, we want to catch any missing dependencies, but we also want to give users the option to ignore these with arbitrary granularity. I'm thinking something along the lines of
```scala
val jlinkSilenceMissingDependency: SettingKey[(String, String) => Boolean]
```

As for the implementation, we can leverage `--ignore-missing-deps`/`--missing-deps` for Java 12 (I think - I haven't tried these out yet). However, for Java 11 we have to parse the full `jdeps` output, so we can probably use this implementation in both cases.

Obviously, this would be more involved than `jdepsOptions`, and it would require additional tests,

vhiairrassary:
You are right 👍I will close this PR then

Issue #1202, opened 2019-02-18
Problem with docker:publishLocal for alpine


Mokrousov:
## Expected behaviour

>>  sbt docker:publishLocal
>> ...
[info] Successfully built 15d0b8df36ec
[info] Successfully tagged akka:0.2
[info] Built image akka with tags [0.2]
[success] Total time: 5 s, completed Feb 18, 2019 5:52:12 PM

## Actual behaviour

>>  sbt docker:publishLocal
>> ...
[info] Step 9/15 : RUN id -u demiourgos728 2> /dev/null || useradd --system --create-home --uid 1001 --gid 0 demiourgos728
[info]  ---> Running in 8251a433d260
[info] /bin/sh: useradd: not found
[info] Removing intermediate container 8251a433d260
[error] The command '/bin/sh -c id -u demiourgos728 2> /dev/null || useradd --system --create-home --uid 1001 --gid 0 demiourgos728' returned a non-zero code: 127
[error] java.lang.RuntimeException: Nonzero exit value: 127
[error]         at com.typesafe.sbt.packager.docker.DockerPlugin$.publishLocalDocker(DockerPlugin.scala:482)
[error]         at com.typesafe.sbt.packager.docker.DockerPlugin$.$anonfun$projectSettings$34(DockerPlugin.scala:186)
[error]         at com.typesafe.sbt.packager.docker.DockerPlugin$.$anonfun$projectSettings$34$adapted(DockerPlugin.scala:184)
[error]         at scala.Function1.$anonfun$compose$1(Function1.scala:44)
[error]         at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:40)
[error]         at sbt.std.Transform$$anon$4.work(System.scala:67)
[error]         at sbt.Execute.$anonfun$submit$2(Execute.scala:269)
[error]         at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:16)
[error]         at sbt.Execute.work(Execute.scala:278)
[error]         at sbt.Execute.$anonfun$submit$1(Execute.scala:269)
[error]         at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:178)
[error]         at sbt.CompletionService$$anon$2.call(CompletionService.scala:37)
[error]         at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #1235, opened 2019-05-30
Possible improvement for #1202

borice:
This should allow more flexibility when using images based on traditional `useradd/groupadd` and also Busybox-based systems using `adduser/addgroup`. This should provide an improvement for #1202


muuki88:
Thanks a lot @borice for your patience and initiative to implement your proposed solution. The PR is perfectly done :heart: I'll release this ASAP.

PR #1241, opened 2019-06-21
Use OpenJDK intead of Oracle JDK

### LINKED ISSUE #1239: Swtich to OpenJDK
## Expected behaviour

It's better to use OpenJDK at Travis CI.

## Actual behaviour

We use Oracle JDK at Travis CI, and it causes a build failure.

https://travis-ci.org/sbt/sbt-native-packager/jobs/545737004#L166

jiminhsieh:
Fix #1239 

* Use https://github.com/shyiko/jabba to control different JDK version
* Use Zulu OpenJDK
* Migrate Oracle JDK to OpenJDK at Travis CI and AppVeyor both


muuki88:
Thanks a lot @jiminhsieh :heart: 

Why not use the `openjdk` directly from travis and don't use _jabba_? For me it looks like this adds more complexcity and diminishes the configuration options from the travis yaml.

.travis.yml
```diff
@@ -1,7 +1,6 @@
-language: java
+language: scala
```
muuki88: Nice for trying that out! I hope that this works as expected

appveyor.yml
```diff
@@ -1,14 +1,17 @@
 version: '{build}'
 os: Windows Server 2012
 install:
+  - cmd: choco install zulu8 -ia "INSTALLDIR=""C:\zulu"""
```
muuki88: Appveyor was still working I guess. Is this necessary to do?

muuki88:
I figured that the release failed due to the oraclejdk issue and commited some tests on master as well. I updated the branch. Hope this doesn't generate too much hazle on your side :grimacing:

.travis.yml
jiminhsieh: After this change, we don't need to download sbt ourself. Travis CI will include a sbt

appveyor.yml
jiminhsieh: At AppVeyor, it used Oracle JDK, so I think it would be good to use the same JDK at different OS

jiminhsieh:
The reason I don't use OpenJDK from Travis CI directly is I think it would be better to use JDK passed TCK. After using jabba, we can use the same JDK at linux and osx. It's ok to use OpenJDK from Travis CI (I guess it was built by Ubuntu), however, I prefer to use JDK with certification. :P

PR #1242, opened 2019-06-25
Assorted JlinkPlugin improvements

### LINKED ISSUE #1240: Assorted JlinkPlugin improvements
I finally got around to using `JlinkPlugin` in my work projects, and there are some things that need work. I'll probably do these myself soon-ish, so this issue is just for documentation.

1. We need more flexible constructors in `Ignore`. I'm thinking some simple string globbing.
1. Our `jlink` call should include our classpath in `--module-path`, since there might be modular JARs there, and these are caught by our `jdeps` output parser.
1. If we have no explicit `javaHome` we use `java` from `PATH`, which is not necessarily the JVM in which SBT is running. We should use `java.home` system property instead.

nigredo-tori:
Closes #1240

Everything is as outlined in #1240, except I've decided against globbing in p.1. The semantics for that would have to be different enough to be confusing. Instead I've added simple package prefix matching.


nigredo-tori:
The CI failure doesn't seem to be related to this change.

muuki88:
Releasing in `v1.3.24`

PR #1244, opened 2019-07-10
JlinkPlugin: support multi-release dependencies

### LINKED ISSUE #1243: JlinkPlugin: support multi-release dependencies
For multi-release dependency JARs `jdeps` needs a `--multi-release` flag with the required JVM version. We can get this from the same Java toolchain - no need to expose it as a setting or to check for multiple versions. The whole point of the plugin is that we know what JVM version the application will be using.

This issue is for documentation purposes only, I'll try and do this myself tomorrow.

nigredo-tori:
Closes #1243 .

1. Changed the plugin to parse JDK's `release` file and get the value of `JAVA_VERSION` property. The "useful" part of the version number (e.g. 11 for 11.0.3, or 9 for 1.9.0), is then passed to jdeps (`--multi-release`).
1. Added a `scripted` test project for this. There's a bit of over-engineering here, so that in the future we don't have to add a whole test project for simple chages like this.
1. Changed the way we run Java tools, so that in case of an error the output is dumped to the SBT log.

### DIFF
--- a/src/sbt-test/jlink/test-jlink-misc/build.sbt
+++ b/src/sbt-test/jlink/test-jlink-misc/build.sbt
@@ -1,0 +1,23 @@
+// Various JlinkPlugin test cases that don't warrant setting up separate
+// `scripted` tests.
+
+import scala.sys.process.Process
+import com.typesafe.sbt.packager.Compat._
+
+val runChecks = taskKey[Unit]("Run checks for a specific issue")
+
+// Exclude Scala by default to simplify the test.
+autoScalaLibrary in ThisBuild := false
+
+// Should succeed for multi-release artifacts
+val issue1243 = project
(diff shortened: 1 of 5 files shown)


src/sbt-test/jlink/test-jlink-misc/build.sbt
```diff
@@ -0,0 +1,23 @@
...
+import scala.sys.process.Process
+import com.typesafe.sbt.packager.Compat._
+
+val runChecks = taskKey[Unit]("Run checks for a specific issue")
+
+// Exclude Scala by default to simplify the test.
+autoScalaLibrary in ThisBuild := false
+
+// Should succeed for multi-release artifacts
+val issue1243 = project
```
muuki88: I quite like this naming convention for the projects!

PR #1245, opened 2019-07-10
JlinkPlugin: Sort missing dependencies before logging

nigredo-tori:
This makes it easier to build the ignore expression afterwards.


nigredo-tori:
The failure is in an unrelated test group.

Issue #1246, opened 2019-07-11
Flaky rpm tests


muuki88:
The `override-start-script-systemd` (see [job on master](https://travis-ci.org/sbt/sbt-native-packager/jobs/556821998)) throws `NullPointerException`s.

We had two changes in #1212 and #1211 to the RPM output logging. Maybe we require an additional null check somehwere.

```
[info] java.lang.NullPointerException
[info] 	at sbt.ConsoleLogger.log(ConsoleLogger.scala:195)
[info] 	at sbt.ConsoleLogger.log(ConsoleLogger.scala:185)
[info] 	at sbt.AbstractLogger.log(Logger.scala:28)
[info] 	at sbt.MultiLogger$$anonfun$dispatch$1.apply(MultiLogger.scala:35)
[info] 	at sbt.MultiLogger$$anonfun$dispatch$1.apply(MultiLogger.scala:33)
[info] 	at scala.collection.immutable.List.foreach(List.scala:318)
[info] 	at sbt.MultiLogger.dispatch(MultiLogger.scala:33)
[info] 	at sbt.MultiLogger.log(MultiLogger.scala:27)
[info] 	at sbt.Logger$class.info(Logger.scala:117)
[info] 	at sbt.AbstractLogger.info(Logger.scala:11)
[info] 	at com.typesafe.sbt.packager.rpm.RpmHelper$$anonfun$buildPackage$1$$anonfun$apply$10.apply(RpmHelper.scala:114)
[info] 	at com.typesafe.sbt.packager.rpm.RpmHelper$$anonfun$buildPackage$1$$anonfun$apply$10.apply(RpmHelper.scala:114)
[info] 	at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
[info] 	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:47)
[info] 	at com.typesafe.sbt.packager.rpm.RpmHelper$$anonfun$buildPackage$1.apply(RpmHelper.scala:114)
[info] 	at com.typesafe.sbt.packager.rpm.RpmHelper$$anonfun$buildPackage$1.apply(RpmHelper.scala:88)
[info] 	at sbt.IO$.withTemporaryDirectory(IO.scala:358)
[info] 	at com.typesafe.sbt.packager.rpm.RpmHelper$.buildPackage(RpmHelper.scala:88)
[...]

PR #1248, opened 2019-07-12
JlinkPlugin: restrict linking to platform modules

### LINKED ISSUE #1247: JlinkPlugin: properly handle external modules
I finally got around to `jlink`'ing actual modular JARs (in a larger project), and the current implementation doesn't work.
1. We build `--module-path` using commas as separators. Jlink documentation for Java 9 and 10 seems to suggest semicolons instead. Documentation for Java 11 and Java 12 doesn't mention that at all, but colons (!) seem to work.
1. `--module-path` supports following items:
    - non-existent paths;
    - expanded class directories;
    - modular JARs.
    - non-modular JARs.

    For non-modular JARs `jlink` tries to generate an automatic module descriptor, using the JAR's file name. If that results in an illegal module name (like, say, `cats.core.2.12`), then it raises an error:
    > Error: Unable to derive module descriptor for foo.jar

What happens now, is that comma-separated `--module-path` list is treated as a single path. Since such file doesn't exist, the effective `--module-path` is empty. This works when using the standard modules, but fails once we start adding external modular JARs. Also, this huge path can reach OS path size limits, resulting in an exception when running `jlink`.

A short-term fix for this looks like this:
1. Fix the `--module-path` separator issue.
1. Filter out non-modular JARs from `--module-path` list. One way to do this would be `jar -d -f <file>` (though that looks troublesome, since it also generates an automatic module descriptor).

However, I have reservations about adding more command-line utilities to the mix. I think we've come to the point where it makes more sense to leverage `java.lang.module` for things like that, even though that would mean that `JlinkPlugin` would require `SBT` to run in Java 9+.

Therefore, a better fix looks like this:
1. Move `JlinkPlugin` to a separate artifact or project (`sbt-native-packager-jlink`), that would require Java 9 to build/use.
1. Use `java.lang.module` to detect module JARs.
1. Replace the `jdeps` package graph analysis with directly analyzing working with modules via `java.lang.module`.

This way we are isolated from changes in `jdeps`/`jar` output, and we have a lot of flexibility when dealing with modules. We lose `jdeps` package dependency check, but I'd argue it is not in the scope of this plugin.

nigredo-tori:
Closes #1247 .

This is a short-term solution to #1247, as outlined in https://github.com/sbt/sbt-native-packager/issues/1247#issuecomment-510532079 (point 1). The long-term solution (`JlinkPlugin` refactoring/rewrite - https://github.com/sbt/sbt-native-packager/issues/1247#issuecomment-510472779) will be a separate issue.

1. Fix `jlink` `--module-path` separator issue.
1. Filter detected dependencies - only link the platform modules.
1. Change `jlinkModulePath` to default to an empty list - we don't need it for platform modules.

Detecting platform modules has special cases - for [historical reasons](https://github.com/eclipse-ee4j/ee4j/issues/34) some Jakarta modules use `java.*` names, so we have to exclude them on a case by case basis. I'm not positive I got all of these - some seem to rely on automatic modules for their dependencies, and those are not properly supported by `jlink`/`jdeps` AFAICT... All in all, Jakarta's JPMS transition is a mess.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkPlugin.scala
@@ -100,16 +100,30 @@ object JlinkPlugin extends AutoPlugin {
         sys.error("Missing package dependencies")
       }
 
-      // Collect all the found modules
-      deps.collect {
+      val detectedModuleDeps = deps.collect {
         case PackageDependency(_, _, PackageDependency.Module(module)) =>
           module
-      }.distinct
+      }.toSet
+
+      // Some JakartaEE artifacts use `java.*` module names, even though
+      // they are not a part of the platform anymore.
+      // https://github.com/eclipse-ee4j/ee4j/issues/34
+      // This requires special handling on our part when deciding if the module
+      // is a part of the platform or not.
+      // At least the new modules shouldn't be doing this...
+      val knownJakartaJavaModules = Set("java.xml.bind", "java.xml.soap", "java.ws.rs")
+
+      val filteredModuleDeps = detectedModuleDeps
+        .filter { m =>
+          m.startsWith("jdk.") || m.startsWith("java.")
+        }
+        .filterNot(knownJakartaJavaModules.contains)
+
+      // We always want `java.base`, and `jlink` requires at least one module.
+      (filteredModuleDeps + "java.base").toSeq
     },
+    // No external modules by default: see #1247.
     jlinkModulePath := (jlinkModulePath ?? Nil).value,
-    jlinkModulePath ++= {
-      fullClasspath.in(jlinkBuildImage).value.map(_.data)
-    },
     jlinkOptions := (jlinkOptions ?? Nil).value,
     jlinkOptions ++= {
       val modules = jlinkModules.value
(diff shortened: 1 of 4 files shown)


nigredo-tori:
Test failure is due to this: #1246

src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkPlugin.scala
```diff
@@ -100,16 +100,30 @@ object JlinkPlugin extends AutoPlugin {
...
-      }.distinct
+      }.toSet
+
+      // Some JakartaEE artifacts use `java.*` module names, even though
+      // they are not a part of the platform anymore.
+      // https://github.com/eclipse-ee4j/ee4j/issues/34
+      // This requires special handling on our part when deciding if the module
+      // is a part of the platform or not.
+      // At least the new modules shouldn't be doing this...
+      val knownJakartaJavaModules = Set("java.xml.bind", "java.xml.soap", "java.ws.rs")
```
muuki88: I'm curious if this is something we should make configurable if there are more of these special cases?

nigredo-tori: Users can filter `jlinkModules` to achieve that, so we don't need another setting.

muuki88: Thanks for the clarification :+1:

PR #1251, opened 2019-08-07
Support building Graal native images in docker

### LINKED ISSUE #1250: Build Graal native images in a docker container
This has been talked about a little in #1165, I'm creating an independent issue for it.

The big reason for building native images in a docker container is that Graal native images don't support cross compiling to different platforms from the host, so the native image you build will be specific for your platform. ie, if you have a Mac, and you build a native image there, you won't be able to deploy that image to Linux.

@adamw has a great blog post on building Graal native images in a docker container with sbt [here](https://blog.softwaremill.com/small-fast-docker-images-using-graalvms-native-image-99c0bc92e70b), as a starting point.

In #1165, a good point is made that building in a docker container would be also useful for building debs and rpms, and so the feature should hopefully be somewhat reusable. That said, I'm not sure how much reuse there is to be had. The docker containers required to do the build are completely different, for Graal, you need one with Graal native-image installed. For debs, you need a Debian based distro image, and for rpms you need a RedHat based distro image. Furthermore, what needs to actually be done in those containers varies wildly depending on the task. For Graal, it's just a matter of running the container with a few directories mounted and passing a few arguments. For rpm/deb I would imagine it would first require installing a bunch of dependencies etc. I wouldn't be surprised if the only thing that makes sense to share between them is the location of the `docker` executable, from that point on, each build is going to do very different things with very different docker containers, I'd be surprised if there's anything else that could be shared.

So I think the best approach is to implement them independently, and then if any common functionality is discovered, we can refactor them to share it, but I don't expect that any will be.

jroper:
Closes #1250

This provides support for building Graal native images in a docker container.

The simplest way to enable the docker container build is to set:

```scala
graalVMNativeImageGraalVersion := Some("19.1.1")
```

When setting this, the plugin will generate a docker image based on `oracle/graalvm-ce:19.1.1` which installs the `native-image` command. It will then run this image to build the Graal native image. Options are provided to customise which base image the generated image uses, along with allowing it to use a prepackaged image rather than generating one.

The graal plugin has been restructured a little to be more idiomatic for sbt. I also introduced a dependency on the `JavaAppPackaging` plugin, because a graal vm image in this case is a java app, and this allows us to use the `scriptClasspathOrdering`, so configuring it and the behaviour of the classpath is consistent with the rest of native packager.

I also moved the setting of the `dockerExecCommand` setting from the `DockerPlugin` to the `UniversalPlugin`, this means that any native packager plugin that depends on `UniversalPlugin` can use that setting without having to depend on `DockerPlugin`.


nigredo-tori:
@jroper, as you've said in #1250, there's a benefit in making the docker/native split feature somewhat reusable. I agree that there's not much hope here for reusing implementation. However, we should strive to at least make the API predictable.
To that end, I propose we add a key like this to `UniversalPlugin`, or some other place where it will be readily available for any SNP plugin:
```scala
val sourceDockerImage = taskKey[Option[String]]("Coordinates for a docker image to use in a command")
```
This is intended to specify an image for use in an arbitrary command. E.g. in the case of the `GraalVMNativeImagePlugin`, we would have
```scala
GraalVMNativeImage / packageBin / sourceDockerImage := {
  // build an image, or provide coordinates for an existing one
}
GraalVMNativeImage / packageBin := {
  val sourceDockerImage0 := (GraalVMNativeImage / packageBin / sourceDockerImage).value
  // ...
  sourceDockerImage0 match {
    case None =>
      buildLocal(...)
    case Some(image) =>
      buildInDockerContainer(... , image, ...)
  }
}
```
This way we can provide a consistent API between plugins.

Of course, there's room for improvement. One thing that comes to mind is that there might be commands that only support Docker execution, in which case a `None` in `sourceDockerImage` would make no sense. So we can make `sourceDockerImage` a `TaskKey[String]`, and choose the execution branch depending on whether it is defined for the command is question.

jroper:
> and choose the execution branch depending on whether it is defined for the command is question

I'm not a big fan of logic based on whether something's defined or not, it's a bit magic. `Option` makes it clear. I think it's fine to just fail if it's not defined, we currently do that for `mainClass`, which is an `Option`, but is required.

jroper:
Updated. I called the setting `containerBuildImage`, which reflects that this is the image used when the build is done in a container.

src/main/scala/com/typesafe/sbt/packager/graalvmnativeimage/GraalVMNativeImagePlugin.scala
```diff
@@ -0,0 +1,191 @@
...
+        Cmd("WORKDIR", "/opt/graalvm"),
+        ExecCmd("RUN", "gu", "install", "native-image"),
+        ExecCmd("ENTRYPOINT", "native-image")
+      ).makeContent
+
+      IO.write(dockerFile, dockerContent)
+
+      DockerPlugin.publishLocalDocker(
+        target,
+        dockerCommand ++ Seq("build", "-f", dockerFile.getAbsolutePath, "-t", imageName, "."),
```
nigredo-tori: This sends the whole `target` as a context, but doesn't use it... I don't have Docker handy at the moment, but something like that might work:
```
docker build -f <dockerFile> -t <imageName> -
```
You can also avoid an unnecessary `Dockerfile` in your target, and instead pipe that data directly into
```
docker build -t <imageName> -
```
though there's no way to do this with `publishLocalDocker` as it is now.

src/main/scala/com/typesafe/sbt/packager/graalvmnativeimage/GraalVMNativeImagePlugin.scala
```diff
@@ -0,0 +1,191 @@
...
+    val dockerCommand = (DockerPlugin.autoImport.dockerExecCommand in GraalVMNativeImage).value
+    val streams = Keys.streams.value
+    val target = (Keys.target in GraalVMNativeImage).value
+
+    val (baseName, tag) = baseImage.split(":", 2) match {
+      case Array(n, t) => (n, t)
+      case Array(n)    => (n, "latest")
+    }
+
+    val imageName = s"${baseName.replace('/', '-')}-native-image:$tag"
```
nigredo-tori: Should we rely on tags here? This can be problematic, e.g. if we have two builds running in parallel with different versions of this plugin.

`docker build -q -f <dockerFile> .` builds an untagged image, and prints its ID. We can use that instead.

src/main/scala/com/typesafe/sbt/packager/graalvmnativeimage/GraalVMNativeImagePlugin.scala
nigredo-tori: Update: `docker build -f <dockerFile> -t <imageName> -` doesn't work for me: you only can build without context when you're piping the `Dockerfile`.

src/main/scala/com/typesafe/sbt/packager/graalvmnativeimage/GraalVMNativeImagePlugin.scala
jroper: The problem is, I don't want to have to rebuild the image every time I build my graal native image, nor do I want to build separate images for each project. This only rebuilds the image if the tag is not present. This is mentioned in the docs.

I don't think two builds running in parallel (or rather, one build using another builds cached image, which is the real problem) is going to cause a problem. The image produced just needs to execute `native-image` as its entrypoint. It doesn't matter what dockerfile was used to produce that, so it doesn't matter which one gets used. And, if in some rare case it does matter, then it can be overridden.

Besides, we are 100% in control of the docker file, if some future version of native packager changes it in some way that is not compatible with old images generated, then all we have to do is change the generated tag name used by the new version of native packager, and we guarantee no conflicts.

src/main/scala/com/typesafe/sbt/packager/graalvmnativeimage/GraalVMNativeImagePlugin.scala
jroper: Updated, this now uses `docker build -t <imageName> -`.

jroper: Yep, I'm piping the docker file.

src/main/scala/com/typesafe/sbt/packager/graalvmnativeimage/GraalVMNativeImagePlugin.scala
nigredo-tori: > The problem is, I don't want to have to rebuild the image every time I build my graal native image, nor do I want to build separate images for each project. This only rebuilds the image if the tag is not present. This is mentioned in the docs.

That shouldn't be much of an issue in practice, since our `Dockerfile` doesn't use `ADD` and `COPY`, so it should just resolve the layers from cache.

> It doesn't matter what dockerfile was used to produce that, so it doesn't matter which one gets used.

If we later want to add any other feature to this image, this will no longer hold. There's also the fact that the tagged source image can change (especially if the tag is `latest`). Ultimately this is future-proofing on my side, paired with paranoia about all mutable global state. :smile:

> Besides, we are 100% in control of the docker file, if some future version of native packager changes it in some way that is not compatible with old images generated, then all we have to do is change the generated tag name used by the new version of native packager, and we guarantee no conflicts.

It's a fair point that we control the docker file here, so YAGNI applies. However, changing the tag in case of a conflict seems fragile. If someone uses our known tag for some purpose (e.g. to clean up autogenerated images), we are going to have a problem in that scenario. But using the existing tag should work *for now*, so I'm resolving this.

jroper:
Something I've encountered while using this in a real project, I want to limit the amount of memory that `native-image` uses. By default it uses 80% of available physical memory on the machine (see [here](https://www.graalvm.org/docs/reference-manual/aot-compilation/)), which means if sbt is using more than 20%, (or if you have browsers/slack etc running) and you don't have swap, it's going to be OOM killed on Linux at least. [This bug](https://github.com/oracle/graal/issues/1588) makes it impossible to limit it. So, what I'd like to do is use Docker to limit it, eg by using `docker run -m4g`. To do that, I need to be able to customise the docker run command used. So, how should that be allowed? Should we add a new key to the Docker plugin for the run command? Or to the Universal plugin?

jroper:
Using this on a real project, I'm starting to think going with depending on `JavaAppPackaging` and using `scriptClasspathOrdering` is not the best idea. The Graal classpath may have a lot of dependencies that are only needed by the compiler, not needed at runtime (dependencies that contain Graal configuration). These can be marked as `provided` if you use the `Compile` classpath, but `scriptClasspathOrdering` uses the `Runtime` classpath. So I think I'll revert to just using the `Compile` classpath.

muuki88:
Sorry for my late reply. 

Thanks for tackling this issue :hugs: The general idea looks good to me. IMHO making this reusable shouldn't be a big focus. RPM and Debian are stable since years with rather less complaints/issues. On the other hand docker changes rather frequently and it's hard to keep up with those changes. Long story short: If there's a user willing to contribute a "docker-rpm build", then a shared API as @nigredo-tori suggested is the maximum we should do.

> So, what I'd like to do is use Docker to limit it, eg by using `docker run -m4g`. To do that, I need to be able to customise the docker run command used. So, how should that be allowed? Should we add a new key to the Docker plugin for the run command? Or to the Universal plugin?

We already have `dockerBuildOptions`, but those are getting appended after the `build` parameter. The `dockerExecCommand` is already a sequence, which would allow us to hijack that, which maybe is fine for an initial start.

If this seems to error prone we could introduce something like `dockerExecOptions: Setting[Seq[String]]`. WDYT?

muuki88:
I retriggered all builds as there seem to be major issues on all CI vendors -.-

muuki88:
Thanks @jroper for the implementation :heart:
I triggered a release for `1.4.0`. Should be available soon :smile:

jtjeferreira:
> Using this on a real project, I'm starting to think going with depending on `JavaAppPackaging` and using `scriptClasspathOrdering` is not the best idea. The Graal classpath may have a lot of dependencies that are only needed by the compiler, not needed at runtime (dependencies that contain Graal configuration). These can be marked as `provided` if you use the `Compile` classpath, but `scriptClasspathOrdering` uses the `Runtime` classpath. So I think I'll revert to just using the `Compile` classpath.

@jroper this was not done, i.e use `Compile` classpath instead of `scriptClasspathOrdering`, right? When upgrading to 1.4.1 this broke my native-image-compilation because I was using provided dependencies...

jroper:
Yeah, I should have backed out that change (I did in my local copy of the plugin).

muuki88:
@jtjeferreira or @jroper  would you like to open a pull request to fix this?

PR #1268, opened 2019-10-15
Layered Docker images using globs

### LINKED ISSUE #1267: docker: introduce more layers for smaller images
## Expected behaviour

`sbt docker:stage` could generate a `Dockerfile` with multiple layers:
- one for the OS with JRE
- one for the dependencies
- one for the resources
- one for the code

The idea is to generate layers that do not change frequently and to re-use them.
For example, if we change just a line of code, only the last layer has to be pushed.

## Actual behaviour

`sbt docker:stage` generate a `Dockerfile` with multiple layers:
- one for the OS with JRE
- one for all the application needs (dependencies, resources, code)

## Information

https://phauer.com/2019/no-fat-jar-in-docker-image/

CremboC:
Fixes #1267 

This is a WIP, it does something. I have no real experience working with Dockerfile or this plugin, so this may be completely off! It's quite limited right now, but if people have ideas how to make it better I'd love to hear them.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -140,7 +140,7 @@ object DockerPlugin extends AutoPlugin {
       val layers =
         fileGroupingsSorted.map {
           case (glob, layer) =>
-            makeCopySrcDest(s"$dockerBaseDirectory/lib/$glob", s"$dockerBaseDirectory/lib/$layer")
+            makeCopySrcDest(s"$dockerBaseDirectory/lib/$glob", s"$dockerBaseDirectory/lib/l$layer/")
         }
 
       val generalCommands = makeFrom(base) +: makeMaintainer((maintainer in Docker).value).toSeq
@@ -169,8 +169,8 @@ object DockerPlugin extends AutoPlugin {
             val copies = fileGroupingsSorted.map {
               case (_, layer) =>
                 makeCopyFromSrcDest(
-                  s"$dockerBaseDirectory/lib/$layer",
-                  s"$dockerBaseDirectory/lib/$layer",
+                  s"$dockerBaseDirectory/lib/l$layer/*",
+                  s"$dockerBaseDirectory/lib/",
                   stage0name,
                   user,
                   group
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -125,6 +125,7 @@ object DockerPlugin extends AutoPlugin {
           case (_, path) if path.endsWith(".sh") => DockerChmodType.UserGroupPlusExecute -> path
         }
     },
+    dockerGroupFiles := Seq(("*" -> 0)),
     dockerCommands := {
       val strategy = dockerPermissionStrategy.value
       val dockerBaseDirectory = (defaultLinuxInstallLocation in Docker).value
@@ -134,6 +135,13 @@ object DockerPlugin extends AutoPlugin {
       val gidOpt = (daemonGroupGid in Docker).value
       val base = dockerBaseImage.value
       val addPerms = dockerAdditionalPermissions.value
+      val fileGroupingsSorted = dockerGroupFiles.value.sortBy(_._2)
+
+      val layers =
+        fileGroupingsSorted.map {
+          case (glob, layer) =>
+            makeCopySrcDest(s"$dockerBaseDirectory/lib/$glob", s"$dockerBaseDirectory/lib/$layer")
+        }
 
       val generalCommands = makeFrom(base) +: makeMaintainer((maintainer in Docker).value).toSeq
       val stage0name = "stage0"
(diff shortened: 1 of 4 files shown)


CremboC:
So it does work now, but it's quite inflexible. The issue is that I don't know how to define "copy all but x" files. I think that would make it much easier.

I'm also not sure how to implement @nigredo-tori 's proposed solution with the `PartialFunction[String => Int]`. Rather, I'm not sure how to map it to a `COPY`

nigredo-tori:
> So it does work now, but it's quite inflexible.

I agree. Using globs for this is inconvenient and error-prone. I don't think this can be made more palatable while we constrain ourselves to (1) copying from a single staging directory, and  (2) copying only once per "group" (as defined by `dockerGroupFiles`).

I see two ways forward, each breaking one of those constraints:
1. Instead of staging everything together, we stage each "group" (as defined by `dockerGroupFiles`) in a directory of its own. E.g. stage `bin/app` as `1/bin/app`, `lib/app.jar` as `2/lib/app.jar`, etc. We can then do
    ```Dockerfile
    COPY 1/ my-app-dir/
    COPY 2/ my-app-dir/
    ```
1. We `COPY` each file individually. This deviates from the original issue (we would have a lot more layers), but it achieves the same caching behavior. This should be relatively easy to implement, and it shouldn't break things for people that post-process staged files (except for adding new ones, and we can add support for that).

> @nigredo-tori 's proposed solution with the PartialFunction[String => Int].

I proposed no such thing :smile:. You'll find that the function in my example is total - although there are probably issues with type inference and/or SBT syntax.

CremboC:
I did think about (2) but having checked a real-world example, that can mean even 200 jars to copy, which is insane when you think of it. 200 layers.

I was thinking about an alternative solution, but it would more far-reaching. I noticed the thing that controls the `lib` folder is not Docker, but rather `JavaServerAppPackaging` (might be wrong here?). Do you think it would be possible to change that part so that all jars generated by "our" code would be put outside the `lib` folder? This way we do very little Docker plugin. Let me know if this doesn't make sense.

> I have proposed no such thing 😄. You'll find that the function in my example is total - although there are probably issues with type inference and/or SBT syntax.

Sorry, I was in a rush when I wrote that comment. You clearly suggested just a normal `String => Int`!

nigredo-tori:
> 200 layers

Yes, that does sound bad... Actually, I don't think Docker [will be able to handle that at all](https://github.com/docker/docker.github.io/issues/8230#issuecomment-468630187). So this is not an option.

nigredo-tori:
If we go with (1), we can pass the "flat" staging directory to stage 0, split it into "group" folders there (via a generated script), and then `COPY` those folders in stage 1. The generated script would look something like this:
```bash
#!/usr/bin/env bash
set -ex
// group 1
find . -type d -exec mkdir -p '/groups/1/{}' \;
mv bin/app /groups/1/bin/app
# ...
// group 2
find . -type d -exec mkdir -p '/groups/2/{}' \;
mv lib/app.jar /groups/2/lib/app.jar
# ...
```
Generating this would be messy, but this way we won't break any post-processing people might have in the `Docker / stage`, and won't leave any extra junk behind.

muuki88:
I like (1) copying from different directories into the proper directory in the resulting docker image. This ensures that all start scripts work as before.

> If we go with (1), we can pass the "flat" staging directory to stage 0, split it into "group" folders there (via a generated script), and then COPY those folders in stage 1. 

@nigredo-tori I don't quite get why we need this. The docker target directory is not really for messing around by other sbt tasks, so I would be fine if we change the resulting structure here. I'm fine with configuring everything inside the `dockerGroupFiles` setting, e.g.

```scala
dockerGroupFiles := Seq(
  ("bin/*" -> 0),
  ("conf/*" -> 0),
  ("lib/*" -> 1),
  ("lib/com.example*" -> 2)
)
```

This would mean that we have to put a bit more thought in deduplicating stuff. A file may only appear once and it must the group with the highest ID. The algorithm should be straight forward.

1. Choose the lowest group ID
2. Union all other files from the other groups into a single set
3. Subtract the union set from the lowest group ID file set

There's only a single test failing:
https://travis-ci.org/sbt/sbt-native-packager/jobs/598043699#L1027-L1035

because the `COPY` command changed.

nigredo-tori:
> The docker target directory is not really for messing around by other sbt tasks, so I would be fine if we change the resulting structure here.

If you say so. It's just that I remember starting to use SNP in one of my projects, and the staging directory seemed like the easiest incision point to use for things that require a concrete directory. In fact my custom WiX setup still uses `Universal/stage` to "materialize" the application before running `heat` on it. So I imagine some people might depend on `Docker/stage` in a similar fashion.

muuki88:
> If you say so. It's just that I remember starting to use SNP in one of my projects, and the staging directory seemed like the easiest incision point to use for things that require a concrete directory with all the application files. In fact my custom WiX setup still uses `Universal/stage` to "materialize" the application before running `heat` on it. So I imagine some people might depend on `Docker/stage` in a similar fashion.

That's a valid point. The context for my opinion is this:

I usually recommend using sbt tasks to access values inside the target directory, e.g. `(Docker / target).value` if you need the staging directory.

Changing the output files should also be done with existing tasks, e.g. `Universal / mappings`. You can alter them in any way `Universal / mappings := (Universal / mappings).value.map(changeDestination)`

Having said that, we need to make sure that we preserve the structure defined in `Docker / mappings` in the resulting docker image. The intermediate steps can change.

CremboC:
Sorry guys, I am swamped with higher priority stuff so don't really have time to work on this further at the moment. If you guys can take this PR to the point where it works in all cases that'd be awesome!

muuki88:
> Sorry guys, I am swamped with higher priority stuff so don't really have time to work on this further at the moment. If you guys can take this PR to the point where it works in all cases that'd be awesome!

Thanks for the notice :heart: Maybe I'll find a free spot to continue to work on this :smile:

CremboC:
This has been superseded by #1310 . Thank you all!

muuki88:
Thanks for your time @CremboC even this didn't make it into native packager I'm very grateful for the time you put into this.

PR #1270, opened 2019-10-24
JlinkPlugin: add support for huge classpaths

### LINKED ISSUE #1266: JlinkPlugin: find a workaround for the jdeps command line length limit
## Expected behaviour

JlinkPlugin properly handles large classpaths.

## Actual behaviour

```
[error] (distrib-win64 / jlinkModules) java.io.IOException: Cannot run program "C:\dev\jvm\jdk-11.0.3-x64\bin\jdeps": CreateProcess error=206, Имя  файла или его расширение имеет слишком большую длину
```
("The filename or extension is too long")

## Information

- sbt-native-packager 1.4.0
- sbt 1.3.2
- Windows 10

As a short-time workaround, directly setting `jlinkModules` works, assuming you can get the module list elsewhere (e.g. by building on Linux).

This issue is mostly for documentation, and as a reminder - I expect to get to fixing this soon-ish.

nigredo-tori:
Closes #1266.

As described [here](https://github.com/sbt/sbt-native-packager/issues/1266#issuecomment-541610521), this uses a [launcher](https://github.com/eldis/java-tool-launcher) with `@argfile` support to pass large argument lists to `jdeps` (and `jlink`, just in case).

Since this approach only uses the `java` executable, I've changed the process handling to use SBT's `ForkRun`. This way we don't need to reimplement `javaOptions` and `javaHome` support. There are more hoops to jump through to get the process output out of it, but I think this is worth it.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,6 +18,7 @@ libraryDependencies ++= Seq(
   "org.apache.commons" % "commons-compress" % "1.18",
   // for jdkpackager
   "org.apache.ant" % "ant" % "1.10.5",
+  // workaround for the command line size limit
   "com.github.eldis" % "tool-launcher" % "0.2.1",
   "org.scalatest" %% "scalatest" % "3.0.5" % Test
 )
--- a/build.sbt
+++ b/build.sbt
@@ -18,6 +18,7 @@ libraryDependencies ++= Seq(
   "org.apache.commons" % "commons-compress" % "1.18",
   // for jdkpackager
   "org.apache.ant" % "ant" % "1.10.5",
+  "com.github.eldis" % "tool-launcher" % "0.2.1",
   "org.scalatest" %% "scalatest" % "3.0.5" % Test
 )
 
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkPlugin.scala
@@ -71,7 +71,7 @@ object JlinkPlugin extends AutoPlugin {
 
       // Jdeps has a few convenient options (like --print-module-deps), but those
       // are not flexible enough - we need to parse the full output.
-      val jdepsOutput = runForOutput(run("jdeps", "--multi-release" +: javaVersion +: "-R" +: paths), log)
+      val jdepsOutput = run("jdeps", "--multi-release" +: javaVersion +: "-R" +: paths)
 
       val deps = jdepsOutput.linesIterator
       // There are headers in some of the lines - ignore those.
@@ -146,7 +146,7 @@ object JlinkPlugin extends AutoPlugin {
 
       IO.delete(outDir)
 
-      runForOutput(run("jlink", jlinkOptions.value), log)
+      run("jlink", jlinkOptions.value)
 
       outDir
     },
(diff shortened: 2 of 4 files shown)


nigredo-tori:
Ugh, SBT 0.13 doesn't expose ForkOptions in `Keys` and `Defaults`, and `ScalaRun` API is different. I had to revert to `ProcessBuilder`s.

nigredo-tori:
I'm not sure what [this](https://travis-ci.org/sbt/sbt-native-packager/jobs/602243750) is about, but it's doesn't seem to be caused by this MR.

build.sbt
```diff
@@ -18,6 +18,7 @@ libraryDependencies ++= Seq(
   "org.apache.commons" % "commons-compress" % "1.18",
   // for jdkpackager
   "org.apache.ant" % "ant" % "1.10.5",
+  "com.github.eldis" % "tool-launcher" % "0.2.1",
```
muuki88: Can you add a comment why we need this dependency? :)

Also. How did you find this lib? It looks rather new and built for a specific use case. Will this work in the future?

src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkPlugin.scala
```diff
@@ -176,12 +176,26 @@ object JlinkPlugin extends AutoPlugin {
...
+  private def runJavaTool(jvm: File, log: Logger)(toolName: String, args: Seq[String]): String = {
+    log.info("Running: " + (toolName +: args).mkString(" "))
 
-    log.info("Running: " + (exe +: args).mkString(" "))
+    val toolLauncherClass = classOf[ru.eldis.toollauncher.ToolLauncher]
+    val toolLauncherJar = file(
+      // This assumes that the code source is a file or a directory (as opposed
+      // to a remote URL) - but that should hold.
+      toolLauncherClass.getProtectionDomain.getCodeSource.getLocation.getPath
+    ).getAbsolutePath
```
muuki88: Just if I got this right:
We summon the `ToolLauncher` class to get the path to the `jar` file, which we require to run it with `Process`

src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkPlugin.scala
```diff
@@ -146,7 +146,7 @@ object JlinkPlugin extends AutoPlugin {
 
       IO.delete(outDir)
 
-      runForOutput(run("jlink", jlinkOptions.value), log)
+      run("jlink", jlinkOptions.value)
```
muuki88: This is the sbt `ForkRun` method you mentioned in the ticket description, right?

muuki88:
> I'm not sure what [this](https://travis-ci.org/sbt/sbt-native-packager/jobs/602243750) is about, but it's doesn't seem to be caused by this MR.

This is due to an issue with bintray. See https://github.com/lightbend/mima/issues/422

muuki88:
Feel free to merge :smile:

build.sbt
nigredo-tori: It is new, and it is, indeed built for this exact use case. :smile: I've put it into the company github/sonatype organization so that I don't have to go through the trouble of establishing an organization for myself.

The library is pure Java, and has zero dependencies, so short of Java breaking the `ToolProvider` API, I don't see why this wouldn't work for the foreseeable future. And, since the library itself is a hundred lines of Java code under the MIT license, it should be reasonably easy to update/fix/rewrite it if the unthinkable happens. That is, if we stil have to support Java 8 by that point. :smile:

src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkPlugin.scala
nigredo-tori: Yes, that is correct. We could resolve the JAR separately (via `libraryDependencies`, or directly via `DependencyResolution`), but it's easier to just bring it as a plugin dependency. It has no transitive dependencies, so it shouldn't break anything here.

src/main/scala/com/typesafe/sbt/packager/archetypes/jlink/JlinkPlugin.scala
nigredo-tori: No, that is a convenience alias for `runJavaTool`. The `ForkRun` thing required too much `Compat` noise to be worthwhile. So I've reverted the MR to use `ProcessBuilder`s.

 That idea can be revisited once we drop SBT 0.13.

nigredo-tori:
@muuki88, I can't merge this, since I am not a maintainer. Or it could be because of the failing checks - I'm not sure what the requirements are here. This can be merged once you're OK with it.

muuki88:
Sorry @nigredo-tori . you are right. Only as an _admin_ it's possible to merge if a test is failing. I updated the branch with your latest change. This should fix it :+1:

PR #1279, opened 2019-11-14
Autoremove multi-stage intermediate image(s)

### LINKED ISSUE #1229: Why does it create an <none> image?
## steps

1. `sbt docker:publishLocal` 

## problem

There's a 703MB image named `<none>` that is created.

## expectations

sbt-native-packager cleans after itself, so `<none>` is deleted if it's only used for prep work.

## original bug report

Hi all 

When I run the command: 

`sbt docker:publishLocal`  

it always create the following images: 

```
sapmock             0.1                 db261ab225b5        8 hours ago         664MB
<none>              <none>              0940c75d8e5a        8 hours ago         703MB
openjdk             8                   bec43387959a        10 days ago         625MB
```
and I could not figure out, why it always create an `<none>` image. 

I am using the latest version of sbt-native-packager and the build.sbt looks as the following:

```
enablePlugins(JavaServerAppPackaging)
enablePlugins(DockerPlugin)
dockerExposedPorts := Seq(8080)

```
OS is UBUNTU 18.04  

What am I doing wrong? 

Thanks

mkurz:
Fixes #1229
Splitted in ~~three~~ four clean commits.

Let me know what you think. Thanks!

EDIT: Solution was inspired from here: https://github.com/moby/moby/issues/34513#issuecomment-389467566

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -1,6 +1,7 @@
 package com.typesafe.sbt.packager.docker
 
 import java.io.File
+import java.util.UUID
 import java.util.concurrent.atomic.AtomicBoolean
 
 import sbt._
@@ -96,6 +97,7 @@ object DockerPlugin extends AutoPlugin {
       Option((version in Docker).value)
     ),
     dockerUpdateLatest := false,
+    dockerAutoremoveMultiStageIntermediateImages := true,
     dockerAliases := {
       val alias = dockerAlias.value
       if (dockerUpdateLatest.value) {
@@ -134,6 +136,7 @@ object DockerPlugin extends AutoPlugin {
       val gidOpt = (daemonGroupGid in Docker).value
       val base = dockerBaseImage.value
       val addPerms = dockerAdditionalPermissions.value
+      val uniqueDockerfileId = UUID.randomUUID().toString
 
       val generalCommands = makeFrom(base) +: makeMaintainer((maintainer in Docker).value).toSeq
       val stage0name = "stage0"
@@ -141,6 +144,8 @@ object DockerPlugin extends AutoPlugin {
         case DockerPermissionStrategy.MultiStage =>
           Seq(
             makeFromAs(base, stage0name),
+            makeLabel("sbt-native-packager-multi-stage" -> "intermediate"),
+            makeLabel("sbt-native-packager-multi-stage-id" -> uniqueDockerfileId),
             makeWorkdir(dockerBaseDirectory),
             makeCopy(dockerBaseDirectory),
             makeUser("root"),
@@ -180,7 +185,7 @@ object DockerPlugin extends AutoPlugin {
         // Seq(ExecCmd("RUN", Seq("ls", "-l", "/opt/docker/bin/"): _*)) ++
         Seq(makeEntrypoint(dockerEntrypoint.value), makeCmd(dockerCmd.value))
 
-      stage0 ++ stage1
+      Seq(makeComment(s"id=${uniqueDockerfileId}")) ++ stage0 ++ stage1
     }
   ) ++ mapGenericFilesToDocker ++ inConfig(Docker)(
     Seq(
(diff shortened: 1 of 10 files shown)


src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -134,13 +136,16 @@ object DockerPlugin extends AutoPlugin {
...
+      val uniqueDockerfileId = UUID.randomUUID().toString
 
       val generalCommands = makeFrom(base) +: makeMaintainer((maintainer in Docker).value).toSeq
       val stage0name = "stage0"
       val stage0: Seq[CmdLike] = strategy match {
         case DockerPermissionStrategy.MultiStage =>
           Seq(
             makeFromAs(base, stage0name),
+            makeLabel("sbt-native-packager-multi-stage" -> "intermediate"),
+            makeLabel("sbt-native-packager-multi-stage-id" -> uniqueDockerfileId),
```
muuki88: Can you explain what issue we solve with two labels? From my limited understand we label an image so we can prune it afterwards with `--filter label=our-label`.

From my naive understand the `intermediate` label should be sufficient? Granted that some uniqueness is required so we don't accidentally prune other images. However something with `normalizedName in Docker` would be probably be enough (e.g. `val uniqueDockerfileId = (normalizedName in Docker).value + "-" + (version in Docker).value + "-intermediate".

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -497,12 +516,44 @@ object DockerPlugin extends AutoPlugin {
...
+                         strategy: DockerPermissionStrategy,
+                         removeIntermediateImages: Boolean,
+                         log: Logger): Unit = {
     log.debug("Executing Native " + buildCommand.mkString(" "))
     log.debug("Working directory " + context.toString)
 
     val ret = sys.process.Process(buildCommand, context) ! publishLocalLogger(log)
 
+    // First let's see if there was a comment that tells us the id of the generated dockerfile
+    val headComments = IO.readLines(context / "Dockerfile").takeWhile(_.startsWith("# ")).map(_.substring(2))
```
muuki88: Why do we need a comment? Can't we just search for a `LABEL` definition that starts with `id` (or make it more unique, e.g. `snp-intermediate-id`?

After reading the docs I'm not sure why we need this at all. The filter combination seems restricted enough. The only issue I could think of is when 2 parallel builds are running and cleaning things up afterwards. Is this a real issue?

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
mkurz: > Can you explain what issue we solve with two labels?

One is a very specific label, exactly for one specific build. My intention was that at the end of the build only exactly the intermediate image(s) of that build should be removed, without any possible side effects.
The general label was just added as help in case you disable the autoremove config and want to clean all images later, or in case something went wrong (and cleaning did not happen correctly) and you want to clean manually.

> From my naive understand the `intermediate` label should be sufficient? 

For the mentioned intention above, not it's not. Because we would potentially clean images that the user does not want to clean up. Who knows, maybe a user kept an intermediate image on purpose to inspect it, and when running the build again, suddenly that image would be gone.

> Granted that some uniqueness is required so we don't accidentally prune other images. 

Exactly, that's my whole point. My intention was to not have any side-effects at all.

> However something with `normalizedName in Docker` would be probably be enough (e.g. `val uniqueDockerfileId = (normalizedName in Docker).value + "-" + (version in Docker).value + "-intermediate".

I am not sure this is enough. Sometimes when setting up a build process (like we did recently via CI/CD), we just package dozens of docker packages until we have a correct build process set up. Now, these package all have the same name, (snapshot) version, etc. Now again my argument: Maybe a user wants to keep an intermediate image on purpose, however if you clean based on name + version you still would clean images a user wanted to keep.

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
mkurz: > Why do we need a comment? Can't we just search for a `LABEL` definition that starts with `id` (or make it more unique, e.g. `snp-intermediate-id`?

I can change that (instead of a comment search for the `LABEL` definition and use `snp-intermediate-id`.

> After reading the docs I'm not sure why we need this at all. The filter combination seems restricted enough. The only issue I could think of is when 2 parallel builds are running and cleaning things up afterwards. Is this a real issue?

See my other comment. My point here is I do not want to introduce any sideffects _at all_ when packaging a docker image. I don't think it's a good idea that packaging could potentially clean images that have nothing to do with the **current** build process.

mkurz:
@muuki88 I pushed commits to fix the failing test and to search for the `LABEL` definition instead of the comment approach. Let me know what you think.

muuki88:
`v1.5.0` built is triggered

mkurz:
Thanks a lot @muuki88!

Issue #1035, opened 2017-09-25
missing files cause 'java.io.IOException: This archive contains unclosed entries.'


OlegYch:
## Expected behaviour

`dist` errors out with 'missing file'

## Actual behaviour

`dist` fails with `java.io.IOException: This archive contains unclosed entries.`, and creates a corrupt zip file

## Information

plugins.sbt

`addSbtPlugin("com.typesafe.sbt" % "sbt-native-packager" % "1.2.2")`

build.sbt

```
enablePlugins(JavaAppPackaging)
mappings in Universal in packageBin += file("nothing") -> "nothing"
```

muuki88:
This is also something we should integrate in #1026

I guess at the moment there is no error handling at all.

retronym:
Fixed by #1124. Since version 1.3.9 the `dist` task for the project above fails fast with a better error message:

```
1] Not found: /Users/jz/code/native-package-test/nothing (mapped to nothing)
```

retronym:
Also improving the underlying exception handling: https://github.com/sbt/sbt-native-packager/pull/1290/files

Issue #1026, opened 2017-08-29
Add validatePackageConfiguration task


muuki88:
Newcomers to native-packager sometimes start a the wrong place, e.g. by googling for a certain solution and copy pasting code from a format documentation page, instead of reading the _Getting Started Guide_ or the _readme.md_.

## Solution

To help the users in different situations we should add a `validatePackageConfiguration` task that can be implemented by formats and archetypes and provide helpful warnings / errors.

The validation should run by default and can be turned off by overriding the setting.

**Examples**

1. If the `mappings in Universal` is empty, fail validation
2. If the `JavaServerArchetype` is enabled and `docker` is used, emit a warning
3. If multiple main classes have the same name, fail validation

Related issues:

- #634 
- #1020
- #1030

PR #1290, opened 2019-12-11
Improve error handling in ZIP utilities

retronym:
This was the root cause of #1035. It no longer is as big a problem since #1026 pre-validates that files exists, but it seems worth cleaning this up.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/universal/ZipHelper.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/universal/ZipHelper.scala
@@ -113,9 +113,17 @@ object ZipHelper {
           // Now check to see if we have permissions for this sucker.
           mode foreach (entry.setUnixMode)
           output putArchiveEntry entry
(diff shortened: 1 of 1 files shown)


src/main/scala/com/typesafe/sbt/packager/universal/ZipHelper.scala
```diff
@@ -126,7 +134,13 @@ object ZipHelper {
...
     val zipOut = new ZipArchiveOutputStream(file)
     zipOut setLevel Deflater.BEST_COMPRESSION
-    try { f(zipOut) } finally {
+    try { 
+      f(zipOut) 
+    } catch {
+      case t: Throwable => 
+        IOUtils.closeQuietly(zipOut)
+        throw t
+    } finally {
```
retronym: Oh for Java's try-with-resources.

muuki88:
Started `1.5.2` release

PR #1292, opened 2019-12-17
Add folders from Bloop and Metals to gitignore

jan0sch:
I don't know if it is intended but I saw that the `.metals` folder was included in the last commits.
So I added entries to the `.gitignore` file.

Feel free to ignore this PR if this is intended.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ log/
 target/
 .cache
 .ensime*
+.bloop/*
+.metals/*


nigredo-tori:
@muuki88, we should also delete the existing files from the repository. Also, I believe you were the one to make that commit, so you might want to add the same lines to your global `.gitignore` to avoid dealing with this in any other projects you're working on.

muuki88:
Yeiks! @nigredo-tori  you are right. I missed that I commited those.

PR #1295, opened 2019-12-20
JlinkPlugin: Add and document a workaround for automatic modules

### LINKED ISSUE #1293: jlinkModules fails for modules that require automatic modules
## Expected behaviour

When classpath includes a module that requires an automatic module, `JlinkPlugin` should properly build an image. An example of such module would be

```scala
"com.fasterxml.jackson.module" % "jackson-module-paranamer" % "2.10.1"
```

which depends on the `paranamer` module, even though the paranamer project has never switched to JPMS.

## Actual behaviour

The jdeps call fails:

```
[info] Running: jdeps --multi-release 11 -R /tmp/sbt_8df286d1/test-jlink-misc/foo/target/scala-2.12/classes /home/nigredo/.ivy2/cache/com.fasterxml.jackson.module/jackson-module-paranamer/bundles/jackson-module-paranamer-2.10.1.jar /home/nigredo/.ivy2/cache/com.fasterxml.jackson.core/jackson-databind/bundles/jackson-databind-2.10.1.jar /home/nigredo/.ivy2/cache/com.fasterxml.jackson.core/jackson-annotations/bundles/jackson-annotations-2.10.1.jar /home/nigredo/.ivy2/cache/com.fasterxml.jackson.core/jackson-core/bundles/jackson-core-2.10.1.jar /home/nigredo/.ivy2/cache/com.thoughtworks.paranamer/paranamer/bundles/paranamer-2.8.jar
[error] Exception in thread "main" java.lang.module.FindException: Module paranamer not found, required by com.fasterxml.jackson.module.paranamer
[error] 	at java.base/java.lang.module.Resolver.findFail(Resolver.java:877)
[error] 	at java.base/java.lang.module.Resolver.resolve(Resolver.java:191)
[error] 	at java.base/java.lang.module.Resolver.resolve(Resolver.java:140)
[error] 	at java.base/java.lang.module.Configuration.resolve(Configuration.java:411)
[error] 	at java.base/java.lang.module.Configuration.resolve(Configuration.java:245)
[error] 	at jdk.jdeps/com.sun.tools.jdeps.JdepsConfiguration.<init>(JdepsConfiguration.java:117)
[error] 	at jdk.jdeps/com.sun.tools.jdeps.JdepsConfiguration$Builder.build(JdepsConfiguration.java:563)
[error] 	at jdk.jdeps/com.sun.tools.jdeps.JdepsTask.buildConfig(JdepsTask.java:589)
[error] 	at jdk.jdeps/com.sun.tools.jdeps.JdepsTask.run(JdepsTask.java:543)
[error] 	at jdk.jdeps/com.sun.tools.jdeps.JdepsTask.run(JdepsTask.java:519)
[error] 	at jdk.jdeps/com.sun.tools.jdeps.Main.run(Main.java:64)
[error] 	at jdk.jdeps/com.sun.tools.jdeps.Main$JDepsToolProvider.run(Main.java:73)
[error] 	at java.base/java.util.spi.ToolProvider.run(ToolProvider.java:137)
[error] 	at ru.eldis.toollauncher.ToolLauncher.runTool(ToolLauncher.java:68)
[error] 	at ru.eldis.toollauncher.ToolLauncher.lambda$main$1(ToolLauncher.java:33)
[error] 	at ru.eldis.toollauncher.ToolLauncher.main(ToolLauncher.java:48)
```

## Information

This happens on Ubuntu, JDK 11, sbt-native-packager on master (e900351b). 

## Analysis

At the moment, the only sane way to solve this is to override `jlinkModules` (disabling the `jdeps` step entirely). This helps, but as more Java libraries switch to JPMS, we're likely to see more issues like this. So it would be nice to be able to handle at least a part of such cases out of the box.

As I've mentioned in https://github.com/sbt/sbt-native-packager/issues/1247#issue-466799089, we can't just put all of our classpath in `--module-path` (since some JAR names break automatic modules). Now it turns out that we can't put all of it into `jdeps` arguments, either, since in that case automatic modules will not be recognized. This is inherently messy, since we can't really detect whether a JAR should be a module without looking at other JARs.

I propose we partition class path items depending on whether they can be automatic modules (as in, if they are JARs with "nice" names), put those that match into `--module-path`, and put everything else into args. There are probably some edge cases that won't be solved with this - so we should at least leave an escape hatch to override/augment this heuristic.

nigredo-tori:
Closes #1293.

As described [here](https://github.com/sbt/sbt-native-packager/issues/1293#issuecomment-567818461), this doesn't fix the issue (which is difficult without the access to `ModuleDescriptor` from Java 9+). Instead, it fixes `--module-path` support for the `jdeps` call, making manually handling such issues possible without skipping the `jdeps` phase altogether. This also contains a test and documentation for this workaround.

Also, I've removed the `.metals` files, and moved `JlinkPlugin` documentation to a separate chapter.


nigredo-tori:
This labeler thing seems to be broken for forks: https://github.com/actions/labeler/issues/12.

muuki88:
> This labeler thing seems to be broken for forks: [actions/labeler#12](https://github.com/actions/labeler/issues/12).

Yeah... these github actions are kinda.. new :grin:

Issue #1294, opened 2019-12-18
docker image build not honoring API version


thurstonsand:
## Expected behaviour

When running `docker:publishLocal`, it should honor docker's API version, not version, when determining compatibility.

## Actual behaviour

`docker:publishLocal` improperly looks at version.

## Information

* sbt-native-packager version: 1.5.2
* sbt version: 1.3.5
* ubuntu 18.04
* package: docker
* target system: alpine

I am trying to use Github Actions to build my scala project's docker image. Due to [reasons](https://github.community/t5/GitHub-Actions/What-really-is-docker-3-0-6/td-p/30752), they report a wildly different version number than CE edition (at time of writing, CE reports 19.03.5, Github Actions reports 3.0.8), but correctly report a recent API version number (at time of writing, both report 1.40)

as can be [seen in this build](https://github.com/thurstonsand/scala-hello-world/commit/06b04db7fa59bb32e882e265a16f1a1743420766/checks?check_suite_id=364229480#step:4:210), sbt-native-packager fails to build the docker image because it thinks docker cannot support multi-stage builds. But [as can be seen in this later build](https://github.com/thurstonsand/scala-hello-world/runs/354166026#step:4:212), just [hard-coding a recent version in the build.sbt](https://github.com/thurstonsand/scala-hello-world/blob/7689567457870e18ff1e45b6f4a42a85a5bf044f/build.sbt#L12) causes the build to succeed.

while hard-coding the version works, it's not a very good workaround because it will require constant maintenance as API versions get updated. also, according to the github forums (linked above), Github's docker's maintainers have no plans to change their versions to match CE edition, because API version is supposed to serve as the compatibility check, so it will not be fixed upstream.

muuki88:
Fixed in [version 1.6.0](https://github.com/sbt/sbt-native-packager/releases/tag/v1.6.0)

PR #1296, opened 2020-01-08
Replace dockerVersion with dockerApiVersion

rodrigorn:
Hello!
I have the exact same issue as #1294 according to the [api version matrix](https://docs.docker.com/develop/sdk/#api-version-matrix) these could fix this problem


lightbend-cla-validator:
Hi @rodrigorn,

rodrigorn:
Hi! I have signed the CLA

rodrigorn:
Hi @muuki88 I hope this might be useful regarding your [question](https://stackoverflow.com/questions/59363727/using-a-specific-docker-version-in-github-actions)

src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala
```diff
@@ -37,7 +37,7 @@ trait DockerKeys {
     "Docker CMD. Used together with dockerEntrypoint. Arguments passed in exec form"
   )
   val dockerExecCommand = SettingKey[Seq[String]]("dockerExecCommand", "The shell command used to exec Docker")
-  val dockerVersion = TaskKey[Option[DockerVersion]]("dockerVersion", "The docker server version")
+  val dockerApiVersion = TaskKey[Option[DockerApiVersion]]("dockerApiVersion", "The docker server version")
```
muuki88: Unfortunately we can not do this in minor or patch releases as this breaks every downstream client.
We would have to 
- introduce a new setting `dockerApiVersion`
- deprecate the old setting
- introduce a new setting to switch between "server version" and "api version"  compatibility mode

src/test/scala/com/typesafe/sbt/packager/docker/DockerVersionSpec.scala
```diff
@@ -1,30 +0,0 @@
-package com.typesafe.sbt.packager.docker
```
muuki88: We should keep this as we need to keep the docker version setting

src/main/scala/com/typesafe/sbt/packager/docker/DockerVersion.scala
```diff
@@ -1,18 +0,0 @@
-package com.typesafe.sbt.packager.docker
```
muuki88: We also need to keep this

rodrigorn:
> Thanks for taking the time to make this pull requests and sorry for the late response hugs ( christmas vacation ).

Don't worry, hope you enjoyed your vacation

> Unfortunately we can not just rename this setting in a minor/patch release as this breaks all downstream users. We must
> 
>     * introduce a new setting dockerApiVersion
> 
>     * deprecate the old setting

You are right, I just wanted to see what would imply this change. I'll revert those changes
>     * introduce a new setting to switch between "server version" and "api version" compatibility mode

What about just introduce the new setting `dockerApiVersion` and test both options on `DockerSupport`?

Thanks for your feedback!

src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala
```diff
@@ -38,6 +38,7 @@ trait DockerKeys {
   )
   val dockerExecCommand = SettingKey[Seq[String]]("dockerExecCommand", "The shell command used to exec Docker")
   val dockerVersion = TaskKey[Option[DockerVersion]]("dockerVersion", "The docker server version")
+  val dockerApiVersion = TaskKey[Option[DockerApiVersion]]("dockerApiVersion", "The docker server api version")
```
muuki88: Can you move this into the trait `DockerKeysEx` below.  This will resolve the binary incompatibility issues :smiley:

src/main/scala/com/typesafe/sbt/packager/docker/DockerSpotifyClientPlugin.scala
```diff
@@ -52,7 +52,12 @@ object DockerSpotifyClientPlugin extends AutoPlugin {
 
   override lazy val projectSettings: Seq[Setting[_]] = inConfig(Docker)(clientSettings)
 
-  def clientSettings = Seq(publishLocal := publishLocalDocker.value, dockerVersion := dockerServerVersion.value)
+  def clientSettings =
+    Seq(
+      publishLocal := publishLocalDocker.value,
+      dockerVersion := dockerServerVersion.value,
+      dockerApiVersion := dockerServerApiVersion.value
```
muuki88: Nice :+1:

src/main/scala/com/typesafe/sbt/packager/docker/DockerSupport.scala
```diff
@@ -2,9 +2,9 @@ package com.typesafe.sbt.packager.docker
...
 
-  def chownFlag(version: DockerVersion): Boolean =
-    (version.major == 17 && version.minor >= 9) || version.major > 17
+  def chownFlag(version: DockerVersion, apiVersion: DockerApiVersion): Boolean =
+    ((version.major == 17 && version.minor >= 9) || version.major > 17) || (apiVersion.major == 1 && apiVersion.minor >= 32)
 
-  def multiStage(version: DockerVersion): Boolean =
-    (version.major == 17 && version.minor >= 5) || version.major > 17
+  def multiStage(version: DockerVersion, apiVersion: DockerApiVersion): Boolean =
+    ((version.major == 17 && version.minor >= 5) || version.major > 17) || (apiVersion.major == 1 && apiVersion.minor >= 29)
```
muuki88: I think this is a good path for migration. In the end it is probably saner to only check the api version

src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala
rodrigorn: thanks!

src/main/scala/com/typesafe/sbt/packager/docker/DockerSupport.scala
rodrigorn: agree

rodrigorn: I had to restore the previous versions of this methods due to compatibility issues

rodrigorn:
The test was overriding the server version expecting the plugin to fail, since the api version was still correct the plugin wasn't failing (I think this is the behavior that we are trying to achieve with this pull request). I just overrode the api version too to get the test to pass, but let me know if you want me to add more tests to check other use cases.

muuki88:
Thanks for all your time and thinking you put into this PR 🤗 I'll release this next week.

muuki88:
Released [`1.6.0`](https://github.com/sbt/sbt-native-packager/releases/tag/v1.6.0). Thanks again @rodrigorn :hugs:

rodrigorn:
Thanks to you @muuki88!!

PR #1299, opened 2020-01-26
Introduce RPM / packageBin / artifactPath setting.

### LINKED ISSUE #1287: Expose the generated rpm filepath
The [generated rpm filename](https://github.com/sbt/sbt-native-packager/blob/v1.5.1/src/main/scala/com/typesafe/sbt/packager/rpm/RpmHelper.scala#L40) should be exposed via a setting.

I generate an rpm via `sbt rpm:packageBin` in a CI which I then upload to a nexus using curl:

```bash
# This generates the rpm but does not log anything to the console
PACKAGE_OUTPUT=$(sbt "print rpm:packageBin") 
echo $PACKAGE_OUTPUT
# Extract the rpm filename full path
RPM_FILE=$(echo $PACKAGE_OUTPUT | tail -n1)
curl --user "$NEXUS_USERNAME:$NEXUS_PASSWORD" --upload-file $RPM_FILE https://nexus.example.com/repository/yum-releases/7/os/$ARCH/$(basename $RPM_FILE)
```

But its quite clunky as the `sbt "print rpm:packageBin"` is what actually generates the rpm. It is also the only way to know the generated rpm filename (apart from computing it from the other settings). The code that would be easy to read should be something like:

```bash
sbt "rpm:packageBin"
RPM_FILE=$(sbt "rpm:filename" | tail -n1)
curl --user "$NEXUS_USERNAME:$NEXUS_PASSWORD" --upload-file $RPM_FILE https://nexus.example.com/repository/yum-releases/7/os/$ARCH/$(basename $RPM_FILE)
```

## Information

- What sbt-native-packager are you using: 1.4.1
- What sbt version: 1.3.2
- What is your build system: Arch Linux + rpm-org package
- What package are you building: rpm
- What version has your build tool: 4.15.0
- What is your target system: CentOS 7

Lasering:
Closes #1287

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/rpm/RpmHelper.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/rpm/RpmHelper.scala
@@ -3,6 +3,7 @@ package com.typesafe.sbt.packager.rpm
 import sbt._
 import com.typesafe.sbt.packager.Compat._
 import com.typesafe.sbt.packager.linux.LinuxSymlink
+import sbt.Keys.{artifactPath, packageBin}
 
 object RpmHelper {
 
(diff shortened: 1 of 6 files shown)


Lasering:
@muuki88 I can't retry the AppVeyor CI. It is failing in `windows / test-bat-template` and looking at the code it does not involve RPM. Can you help?

Lasering:
I've added documentation and a sbt-test.

As for the documentation its really small, however I'm not sure if it merits more or not.

Issue #1304, opened 2020-02-05
Building Docker fails to chmod it .sh files


joan38:
I have a script in `src/docker/opt/docker/bin/launch.sh`.

The generated `Dockerfile`:
```
RUN ["chmod", "-R", "u=rX,g=rX", "/opt/docker"]
RUN ["chmod", "u+x,g+x", "/opt/docker/bin/server"]
RUN ["chmod", "u+x,g+x", "opt/docker/bin/launch.sh"]
```
See the missing `/` in front of `opt` that leads to the following error when `Docker/publishLocal`:
```
chmod: cannot access 'opt/docker/bin/launch.sh': No such file or directory
```

Thanks

muuki88:
Thanks for your bug request :smiley: 

A small `build.sbt` to reproduce this error would be nice. Also information about the system you are building the docker image.

adarshaj:
@muuki88 - I've created a repro here: https://github.com/hackcave/sbt-native-packager-docker-issue-1304 , `sbt docker:publishLocal` should give you the error. we can see the file exists at `/opt/bin/docker/entrypoint.sh` within the docker image, but the generated `chmod` command fails to add `/` prefix.

muuki88:
Thanks a lot for putting together a reproducible test case. 

To he honest I don't know if I have the time to work on this, but if you would like to do I'll give you all necessary guidance you need 😊

adarshaj:
@muuki88 - I thought about the issue deeper. I think once we set the workdir (the `defaultLinuxInstallation` value), we should only rely on relative paths from that point onwards. In a multistage build, we are anyway copying only files under this workdir, so even if the user had copied some files in `stage0`, they will not end up in `mainstage`. The current `COPY` generates an absolute path for destination, can we instead make it relative? What do you think of this?

adarshaj:
> we can see the file exists at `/opt/bin/docker/entrypoint.sh` within the docker image

Oops, I switched directories in my reproduction, I was actually wanting to put `entrypoint.sh` in `/opt/docker/bin` but I swapped `bin` and `docker` by mistake.

adarshaj:
Okay, now I understand the core of problem:
[...]

PR #1305, opened 2020-02-05
Fix dockerPackageMappings to be absolute paths

joan38:
Fixes https://github.com/sbt/sbt-native-packager/issues/1304


lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

joan38:
I don't think `MappingsHelper` has any issue since it returns relative paths by design.
Your concerne about the underlying OS would be an issue if we are using a different OS in the docker image we are building (not the host machine that is running SBT).

muuki88:
Hi @joan38 thanks again for tackeling this :smiley: 
Do you know if you have the time to add some tests for this?

joan38:
Hi @muuki88, I'm not sure I'll have much time to do this also since we moved away from SBT to Mill.

muuki88:
Thanks for the update @joan38  😊

PR #1310, opened 2020-02-20
Separate Docker Layers for Dependencies and App jars.

ppiotrow:
Fixes https://github.com/sbt/sbt-native-packager/issues/1267 
Inspired by https://phauer.com/2019/no-fat-jar-in-docker-image/

The idea is to modify docker staging directory to move application related jars from `lib` to `app-lib` directory. Then two `COPY` operations are invoked to the same destination. Dependencies are copied first as they are rarely change. 
staging

Other:
- `mapGenericFilesToDocker` exotic method was reimplemented to be similar to other functions around
- only `DockerPermissionStrategy.MultiStage` branch of code was updated. Others would need to be updated to work with new staging.

Questions:
- Convention to hardcode logic of `organization` filters was ok at first glance, but now I'd rather do it user configurable via project's build.sbt as suggested in the original issue.

Output logs from using newly published plugin in some real project.
First `docker/publishLocal`
```logs
[success] All package validations passed
[info] Sending build context to Docker daemon  26.57MB
[info] Step 1/20 : FROM docker.artifactory.the-company.com/the-team/openjre:11.0.5-0-e58ca54 as stage0
[info]  ---> 16b6e185c5cb
[info] Step 2/20 : LABEL snp-multi-stage="intermediate"
[info]  ---> Running in 539cc1636855
[info] Removing intermediate container 539cc1636855
[info]  ---> 5fe9c7b0dba8
[info] Step 3/20 : LABEL snp-multi-stage-id="ba011bf6-1ed4-4279-8ab4-937abc1a6236"
[info]  ---> Running in 749a51a0aca9
[info] Removing intermediate container 749a51a0aca9
[info]  ---> 8689656277d9
[info] Step 4/20 : WORKDIR /opt/the-project
[info]  ---> Running in 3ace818a32a7
[info] Removing intermediate container 3ace818a32a7
[info]  ---> 96e4b3bd8537
[info] Step 5/20 : COPY opt /opt
[info]  ---> d40dfada42ef
[info] Step 6/20 : USER some-user
[info]  ---> Running in 6b61481f355f
[info] Removing intermediate container 6b61481f355f
[info]  ---> 16e81c1582bd
[info] Step 7/20 : RUN ["chmod", "-R", "u=rX,g=rX", "/opt/the-project"]
[info]  ---> Running in 2f4a9f97d975
[info] Removing intermediate container 2f4a9f97d975
[info]  ---> 6ad7c76a87bf
[info] Step 8/20 : RUN ["chmod", "u+x,g+x", "/opt/the-project/bin/the-project"]
[info]  ---> Running in 6b2df7afc497
[info] Removing intermediate container 6b2df7afc497
[info]  ---> 4848ae77b613
[info] Step 9/20 : FROM docker.artifactory.the-company.com/the-team/openjre:11.0.5-0-e58ca54
[info]  ---> 16b6e185c5cb
[info] Step 10/20 : USER some-user
[info]  ---> Using cache
[info]  ---> 6af66ce6012f
[info] Step 11/20 : RUN id -u some-user 1>/dev/null 2>&1 || (( getent group 0 1>/dev/null 2>&1 || ( type groupadd 1>/dev/null 2>&1 && groupadd -g 0 some-user || addgroup -g 0 -S some-user )) && ( type useradd 1>/dev/null 2>&1 && useradd --system --create-home --uid 1001 --gid 0 some-user || adduser -S -u 1001 -G some-user some-user ))
[info]  ---> Using cache
[info]  ---> e74737ebd8f2
[info] Step 12/20 : WORKDIR /opt/the-project
[info]  ---> Using cache
[info]  ---> dbc0546c52a8
[info] Step 13/20 : COPY --from=stage0 --chown=some-user:some-user /opt/the-project/lib /opt/the-project
[info]  ---> Using cache
[info]  ---> 98297bc28b34
[info] Step 14/20 : COPY --from=stage0 --chown=some-user:some-user /opt/the-project/app-lib /opt/the-project
[info]  ---> beb82b669e4b
[info] Step 15/20 : ENV APP_CLASS_PATH="/opt/the-project/lib/*"
[info]  ---> Running in 423e10592214
[info] Removing intermediate container 423e10592214
[info]  ---> 2563bea886ff
[info] Step 16/20 : ENV APP_MAIN_CLASS="com.the-company.red.Main"
[info]  ---> Running in 59802f900a45
[info] Removing intermediate container 59802f900a45
[info]  ---> de2cbc001a35
[info] Step 17/20 : USER 1001:0
[info]  ---> Running in fa6765c1141c
[info] Removing intermediate container fa6765c1141c
[info]  ---> a26809b0b4dc
[info] Step 18/20 : ENTRYPOINT ["/usr/local/bin/run-class"]
[info]  ---> Running in b3f18864cdab
[info] Removing intermediate container b3f18864cdab
[info]  ---> 85aac1ee1e36
[info] Step 19/20 : CMD []
[info]  ---> Running in fdea9f4b0a17
[info] Removing intermediate container fdea9f4b0a17
[info]  ---> 2f2c119a32e7
[info] Step 20/20 : USER some-user
[info]  ---> Running in 12e5f019898b
[info] Removing intermediate container 12e5f019898b
[info]  ---> 5a7a508a4b22
```
Changing some lines of code and running `docker/publishLocal` again
```
[info] Step 1/20 : FROM docker.artifactory.the-company.com/the-team/openjre:11.0.5-0-e58ca54 as stage0
[info]  ---> 16b6e185c5cb
[info] Step 2/20 : LABEL snp-multi-stage="intermediate"
[info]  ---> Running in 125156cd5b67
[info] Removing intermediate container 125156cd5b67
[info]  ---> c904b531d432
[info] Step 3/20 : LABEL snp-multi-stage-id="c999d42d-5607-426f-b0c0-f207de303312"
[info]  ---> Running in 24f829d1375b
[info] Removing intermediate container 24f829d1375b
[info]  ---> a083afc95f36
[info] Step 4/20 : WORKDIR /opt/the-project
[info]  ---> Running in 2ff312e71914
[info] Removing intermediate container 2ff312e71914
[info]  ---> 75e2a8e9ff7a
[info] Step 5/20 : COPY opt /opt
[info]  ---> 6bac37070916
[info] Step 6/20 : USER some-user
[info]  ---> Running in 6dfbf0b79135
[info] Removing intermediate container 6dfbf0b79135
[info]  ---> fbed0213ad52
[info] Step 7/20 : RUN ["chmod", "-R", "u=rX,g=rX", "/opt/the-project"]
[info]  ---> Running in 95e6c38877e3
[info] Removing intermediate container 95e6c38877e3
[info]  ---> f8ca32bdfb14
[info] Step 8/20 : RUN ["chmod", "u+x,g+x", "/opt/the-project/bin/the-project"]
[info]  ---> Running in f68765152ce6
[info] Removing intermediate container f68765152ce6
[info]  ---> 77a6c19df80d
[info] Step 9/20 : FROM docker.artifactory.the-company.com/the-team/openjre:11.0.5-0-e58ca54
[info]  ---> 16b6e185c5cb
[info] Step 10/20 : USER some-user
[info]  ---> Using cache
[info]  ---> 6af66ce6012f
[info] Step 11/20 : RUN id -u some-user 1>/dev/null 2>&1 || (( getent group 0 1>/dev/null 2>&1 || ( type groupadd 1>/dev/null 2>&1 && groupadd -g 0 some-user || addgroup -g 0 -S some-user )) && ( type useradd 1>/dev/null 2>&1 && useradd --system --create-home --uid 1001 --gid 0 some-user || adduser -S -u 1001 -G some-user some-user ))
[info]  ---> Using cache
[info]  ---> e74737ebd8f2
[info] Step 12/20 : WORKDIR /opt/the-project
[info]  ---> Using cache
[info]  ---> dbc0546c52a8
[info] Step 13/20 : COPY --from=stage0 --chown=some-user:some-user /opt/the-project/lib /opt/the-project
[info]  ---> Using cache
[info]  ---> 98297bc28b34
[info] Step 14/20 : COPY --from=stage0 --chown=some-user:some-user /opt/the-project/app-lib /opt/the-project
[info]  ---> 9378c36fa0e7
[info] Step 15/20 : ENV APP_CLASS_PATH="/opt/the-project/lib/*"
[info]  ---> Running in e94424f66cbd
[info] Removing intermediate container e94424f66cbd
[info]  ---> 8d1d7a06e8f5
[info] Step 16/20 : ENV APP_MAIN_CLASS="com.the-company.red.Main"
[info]  ---> Running in 149b46c7ad6c
[info] Removing intermediate container 149b46c7ad6c
[info]  ---> 81600a029db3
[info] Step 17/20 : USER 1001:0
[info]  ---> Running in 4dd779a082a3
[info] Removing intermediate container 4dd779a082a3
[info]  ---> 719903d06e3e
[info] Step 18/20 : ENTRYPOINT ["/usr/local/bin/run-class"]
[info]  ---> Running in 694314ccaad5
[info] Removing intermediate container 694314ccaad5
[info]  ---> de3b6d226c99
[info] Step 19/20 : CMD []
[info]  ---> Running in af74538469f1
[info] Removing intermediate container af74538469f1
[info]  ---> b2bec70c0733
[info] Step 20/20 : USER some-user
[info]  ---> Running in bdeb33033857
[info] Removing intermediate container bdeb33033857
[info]  ---> 2d6d81e64ea6
```


ppiotrow:
I working on making this better by giving up magic (hardcoded organization) and use solution suggested by https://github.com/sbt/sbt-native-packager/issues/1267#issuecomment-541643279
Could you help me deciding if I should care about binary compatibility validations for `mapGenericFilesToDocker`? If yes I'll leave&deprecate it.
I also found in comments that some people might be unhappy with modification of staging directory layout. It's quite convinient at least in the current state of this PR.

ppiotrow:
Update, I decided to go with nice configurable mapping from file prefix to layer index.
The library now contains default implementation that caches bin then lib and then user organisation code.
staging2

I tried to experiment with reading staging directory with `IO.listFiles` to know what layers (/0 , /1, /17) are present. This would allow me to create perfect Dockerfile. However it would end up with circular dependencies as staging needs Dockerfiles present and Dockerfiles creation would need staging. That is why I decided to throw exception if file doesn't match any pattern.
Could you point me how to do it properly as exceptions are not common in this plugin?

Output now:
```
FROM docker.artifactory.the-company.com/the-team/openjre:11.0.5-0-e58ca54 as stage0
LABEL snp-multi-stage="intermediate"
LABEL snp-multi-stage-id="63cc94e9-7a5b-4e76-80d4-833177fa7293"
WORKDIR /opt/the-project
COPY opt /opt
USER the-user
RUN ["chmod", "-R", "u=rX,g=rX", "/opt/the-project"]

FROM docker.artifactory.the-company.com/the-team/openjre:11.0.5-0-e58ca54
USER the-user
RUN id -u the-user 1>/dev/null 2>&1 || (( getent group 0 1>/dev/null 2>&1 || ( type groupadd 1>/dev/null 2>&1 && groupadd -g 0 the-user || addgroup -g 0 -S the-user )) && ( type useradd 1>/dev/null 2>&1 && useradd --system --create-home --uid 1001 --gid 0 the-user || adduser -S -u 1001 -G the-user the-user ))
WORKDIR /opt/the-project
COPY --from=stage0 --chown=the-user:the-user /opt/the-project/0 /opt/the-project
COPY --from=stage0 --chown=the-user:the-user /opt/the-project/1 /opt/the-project
COPY --from=stage0 --chown=the-user:the-user /opt/the-project/2 /opt/the-project
ENV APP_CLASS_PATH="/opt/the-project/lib/*"
ENV APP_MAIN_CLASS="com.the-company.red.Main"
USER 1001:0
ENTRYPOINT ["/usr/local/bin/run-class"]
CMD []
USER the-user
```

nigredo-tori:
> I tried to experiment with reading staging directory with IO.listFiles to know what layers (/0 , /1, /17) are present. This would allow me to create perfect Dockerfile. However it would end up with circular dependencies as staging needs Dockerfiles present and Dockerfiles creation would need staging. That is why I decided to throw exception if file doesn't match any pattern.

We could add an intermediate task (e.g. `val dockerLayerMappings`) which would contain `Docker / mappings` with their corresponding layers as specified by `dockerLayerGrouping` (`List[(Int, File, String)]` or something like that). We can then use that both in `Docker/stage` (for transformed mappings) and for building the Dockerfile (layer indices). This should solve the circular dependency issue. It also would allow us to keep `dockerLayerGrouping`  as `SettingKey[String => Int]`), which means we wouldn't have have to deal with files not matching patterns.

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -98,6 +98,7 @@ object DockerPlugin extends AutoPlugin {
     ),
     dockerUpdateLatest := false,
     dockerAutoremoveMultiStageIntermediateImages := true,
+    dockerLayerGrouping := Seq(s"lib/${projectID.value.organization}" -> 2, "lib/" -> 1, "bin/" -> 0),
```
nigredo-tori: `projectID.value.organization` can be replaced with `organization.value`.

ppiotrow:
Thanks to @nigredo-tori tips the newest version fixes all three problems from above. The pull request is almost complete apart from this issues:

- [ ] The docker staging directory layout has changed, was it part of contract?
- [ ] When I delete docker staging repository between first and second `docker:publish` it is not using cache for `/bin` directory. The startup scripts have identical content but most likely creation date should be ignored?
> The last-modified and last-accessed times of the file(s) are not considered in these checksums
[docker](https://docs.docker.com/v17.09/engine/userguide/eng-image/dockerfile_best-practices/)

After making sure its on the right track I'll implement code for branches other than `DockerPermissionStrategy.MultiStage`. Waiting for your review now.

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -234,6 +244,18 @@ object DockerPlugin extends AutoPlugin {
...
+        val dockerGroups = dockerLayerGrouping.value
+        def buildDockerPaths(seq: Seq[(File, String)], prefix: String) =
+          for {
+            (file, path) <- seq
+            layerIdx = dockerGroups(path)
+            pathInDocker = s"$prefix/$layerIdx/$path"
+          } yield (layerIdx, file, pathInDocker)
+
+        buildDockerPaths((mappings in Universal).value, (defaultLinuxInstallLocation in Docker).value + "/") ++
+          buildDockerPaths(dockerPackageMappings.value, "")
```
nigredo-tori: It is a bit silly that we group `Universal / mappings` by their relative (to `defaultLinuxInstallLocation`) resulting paths, and `dockerPackageMappings` by absolute resulting paths, **using the same function**. Additionally, we mudden `Docker / mappings` with our layering strategy, which is not ideal. To be honest, I hadn't thought about `dockerPackageMappings` before this, and had assumed we'd only be grouping relative mappings. So it is partially my fault we've ended up here.

I propose we move the layering mechanism to **after** `Docker / mappings` in the build process, and make everything work with absolute resulting paths, as follows.

1. Keep `Docker / mappings` as it was before this PR. That is, the resulting paths (`._2` in the tuples) will still be absolute paths in the resulting Docker image, with no layering noise. This will allow us to avoid breaking anything that relies on `Docker / mappings` shape - e.g. the `dockerAdditionalPermissions` task.
1. `dockerLayerGrouping` should only work with absolute paths in the resulting Docker image. This would make writing it by hand even less convenient because of the `defaultLinuxInstallation` prefix (more on dealing with that later), but it should keep the abstraction cleaner.
1. Build `dockerLayerMappings` from `Docker / mappings` and `defaultLinuxInstallation`. The resulting paths (`._3` in the tuples) here will be absolute paths in the resulting Docker image, with no layering noise.
1. Modify `Docker / stage`. Instead of using `Docker / mappings` here, use `dockerLayerMappings`, and add layer prefixes to its resulting paths.
1. In `dockerCommands` copy each layer from its staging subdirectory with a separate command. The set of available layers can be extracted from `dockerLayerMappings`.
1. To simplify writing `dockerLayerGrouping`, introduce a helper function like this (the name is not final):

    ```scala
    def layerGroupingByStringPrefix(pairs: (String, Int)): String => Int = (s: String) =>
      pairs.find(p => s.startsWith(p._1)).map(_._2).getOrElse(0)
    ```

    which would allow us to write the default `dockerLayerGrouping` like this:

    ```scala
    dockerLayerGrouping := {
      val base = (defaultLinuxInstallLocation in Docker).value
      layerGroupingByStringPrefix(
        s"$base/lib/${organization.value}" -> 2,
        s"$base/bin/" -> 1
      )
    }
    ```

nigredo-tori:
>  When I delete docker staging repository between first and second docker:publish it is not using cache for /bin directory. The startup scripts have identical content but most likely creation date should be ignored?

Creation date [shouldn't matter at all](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache):

> For the ADD and COPY instructions, the contents of the file(s) in the image are examined and a checksum is calculated for each file. The last-modified and last-accessed times of the file(s) are **not considered** in these checksums. During the cache lookup, the checksum is compared against the checksum in the existing images. If anything has changed in the file(s), such as the contents and metadata, then the cache is invalidated.

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -98,7 +98,12 @@ object DockerPlugin extends AutoPlugin {
     ),
     dockerUpdateLatest := false,
     dockerAutoremoveMultiStageIntermediateImages := true,
-    dockerLayerGrouping := Seq(s"lib/${projectID.value.organization}" -> 2, "lib/" -> 1, "bin/" -> 0),
+    dockerLayerGrouping := {
+      case s if s.startsWith(s"lib/${organization.value}") => 2
+      case s if s.startsWith(s"bin/")                      => 1
+      case s if s.startsWith(s"lib/")                      => 0
```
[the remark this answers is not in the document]
ppiotrow: Sure, default function is to be discussed. Let's do it before merging PR.

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -169,10 +174,7 @@ object DockerPlugin extends AutoPlugin {
         Seq(makeWorkdir(dockerBaseDirectory)) ++
         (strategy match {
           case DockerPermissionStrategy.MultiStage =>
-            val layerIdsAscending = dockerLayerGrouping.value
-              .map(_._2)
-              .distinct
-              .sorted(Ordering.Int)
+            val layerIdsAscending = (dockerLayerMappings in Docker).value.map(_._1).distinct.sorted(Ordering.Int)
```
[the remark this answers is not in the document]
ppiotrow: agree

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
ppiotrow: Ok, will hack this tomorrow. I actually tried similar approach, but the grouping  function was less simple when I need to care about `defaultLinuxInstallLocation` prefix.

ppiotrow: When I follow this points I'll end up with staging directory structure like `target/docker/stage/0/opt/the-project/lib`. Notice `0` before `/opt` directory.
However first COPY command in the plugin assumes we copy everything from first directory of `defaultLinuxInstallLocation in Docker` which is /opt.
```
WORKDIR /opt/the-project
COPY opt /opt
```
[This](https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala#L326) means that the very first directory of linux installation it the only source of all files. Even files from `dockerPackageMappings` need to be mapped to /opt directory. This means I should do special handling of the first directory, and add layer_id directories directly after it.

nigredo-tori: > However first COPY command in the plugin assumes we copy everything from first directory of defaultLinuxInstallLocation in Docker which is /opt.

If we want to have separate layers, we will have to replace this with `COPY` commands for each layer. That's the main point of this PR. So we would have something like this:

```Dockerfile
WORKDIR /opt/the-project
COPY 0/* /
COPY 1/* /
...
```

> This means I should do special handling of the first directory, and add layer_id directories directly after it.

Not sure why we would need to do any extra handling. All the files we want copied should be in the layer directories (except for the `Dockerfile`), so we don't need co copy `/opt` separately. I'm probably missing something, since I'm not very familiar with the subject matter, and because it's getting late here - could you please elaborate as to why just copying layer directories as shown above is not sufficient?

ppiotrow: First part of answer is enough to solve this now.

ppiotrow:
Latest changes solve all previous problems, questions and review remarks. Some changes were required in `dockerAdditionalPermissions` as code is executed in `stage0`. 
Code got little dirty after experiments and will be improved after plugin behaviour is accepted.
I've just found a bug in `RUN ["chmod", "-R", "u=rX,g=rX", "/opt/the-project"]` while writing this comment.

staging3

Dockerfile now:
```
FROM docker.artifactory.the-company.com/the-team/openjre:11.0.5-0-e58ca54 as stage0
LABEL snp-multi-stage="intermediate"
LABEL snp-multi-stage-id="02fda8dd-1fdd-4841-9a70-6f5462f0c3a8"
WORKDIR /opt/the-project
COPY 1 /1/
COPY 2 /2/
USER the-user
RUN ["chmod", "-R", "u=rX,g=rX", "/opt/the-project"]
RUN ["chmod", "u+x,g+x", "/1/opt/the-project/bin/the-project"]

FROM docker.artifactory.the-company.com/the-team/openjre:11.0.5-0-e58ca54
USER the-user
RUN id -u demiourgos728 1>/dev/null 2>&1 || (( getent group 0 1>/dev/null 2>&1 || ( type groupadd 1>/dev/null 2>&1 && groupadd -g 0 the-user || addgroup -g 0 -S the-user )) && ( type useradd 1>/dev/null 2>&1 && useradd --system --create-home --uid 1001 --gid 0 demiourgos728 || adduser -S -u 1001 -G the-user demiourgos728 ))
WORKDIR /opt/the-project
COPY --from=stage0 --chown=demiourgos728:the-user /1/opt/the-project /opt/the-project
COPY --from=stage0 --chown=demiourgos728:the-user /2/opt/the-project /opt/the-project
USER 1001:0
ENTRYPOINT ["/opt/the-project/bin/the-project"]
CMD []
USER demiourgos728
```

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -241,20 +248,20 @@ object DockerPlugin extends AutoPlugin {
...
-          } yield (layerIdx, file, pathInDocker)
-
-        buildDockerPaths((mappings in Universal).value, (defaultLinuxInstallLocation in Docker).value + "/") ++
-          buildDockerPaths(dockerPackageMappings.value, "")
+        val dockerFinalFiles = (mappings in Docker).value
+        for {
+          (file, path) <- dockerFinalFiles
+          layerIdx = dockerGroups(path)
+          dockerLayerPath = s"/$layerIdx$path"
+        } yield (layerIdx, file, dockerLayerPath)
```
nigredo-tori: It would probably be better to use `(layerIdx, file, path)` instead, since we can always build `dockerLayerPath` from that where we need it. That way we won't have to replace `.startsWith` with less preciese `.contains` in `dockerAdditionalPermissions `.

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -327,22 +358,30 @@ object DockerPlugin extends AutoPlugin {
     Cmd("COPY", s"$files /$files")
   }
 
+  private final def makeCopy(src: String, dest: String): CmdLike =
```
nigredo-tori: This happens to overload `makeCopy` above, although the behavior differs quite a bit. I think those functions should have different names to avoid confusion. We should probably rename the original one, since it's far less intuitive.

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -192,9 +210,10 @@ object DockerPlugin extends AutoPlugin {
 
       stage0 ++ stage1
     }
-  ) ++ mapGenericFilesToDocker ++ inConfig(Docker)(
+  ) ++ inConfig(Docker)(
     Seq(
       executableScriptName := executableScriptName.value,
+      mappings := mappingsInDocker((mappings in Universal).value, defaultLinuxInstallLocation.value),
```
nigredo-tori: Since `Docker / mappings` haven't changed, we should probably roll back these changes when cleaning up, just to keep the PR focused on one thing.

ppiotrow: Makes sense, but later in separate PR I'd implement this change as previous method is used in uncommon way.

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -169,7 +185,10 @@ object DockerPlugin extends AutoPlugin {
...
         (strategy match {
           case DockerPermissionStrategy.MultiStage =>
-            Seq(makeCopyFrom(dockerBaseDirectory, stage0name, user, group))
+            val layerIdsAscending = (dockerLayerMappings in Docker).value.map(_._1).distinct.sorted
+            layerIdsAscending.map { layerId =>
+              makeCopyFrom(s"/$layerId$dockerBaseDirectory", dockerBaseDirectory, stage0name, user, group)
+            }
           case DockerPermissionStrategy.Run =>
             Seq(makeCopy(dockerBaseDirectory), makeChmodRecursive(dockerChmodType.value, Seq(dockerBaseDirectory))) ++
               (addPerms map { case (tpe, v) => makeChmod(tpe, Seq(v)) })
```
nigredo-tori: Depending on `dockerPermissionStrategy`, we do `chmod`s either in `stage0` (using layered paths) or in `stage1` (using paths without layers) - and `dockerAdditionalPermissions` in its current form can't support that. One possible solution would be to post-process `addPerms` in `stage0` to inject layer directories.

ppiotrow: I have code to solve this, just it wasn't 100% ready yesterday night to be pushed.

src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala
```diff
@@ -57,4 +57,10 @@ private[packager] trait DockerKeysEx extends DockerKeys {
   lazy val dockerAdditionalPermissions =
     taskKey[Seq[(DockerChmodType, String)]]("Explicit chmod calls to some of the paths.")
   val dockerApiVersion = TaskKey[Option[DockerApiVersion]]("dockerApiVersion", "The docker server api version")
+  val dockerLayerGrouping = settingKey[String => Int](
+    "Group files by path into in layers to increase docker cache hits. " +
+      "Lower index means the file would be a part of an earlier layer."
+  )
+  val dockerLayerMappings =
+    taskKey[Seq[(Int, File, String)]]("List of layer, source file and destination in Docker image.")
```
muuki88: Move those to [`DockerKeysEx`](https://github.com/sbt/sbt-native-packager/blob/dec9481686ae559a9a72be6a7440bca6ce6ae90a/src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala#L54), which solves the binary incompat issue.

muuki88: I would create a case class for this, which makes it easier to give the types names and extend it.

```scala
case class LayeredMapping(layerId: Int, file: file, path: String)
```
for example

ppiotrow:
I'm looking at failing tests on my computer.
`scripted docker/build-command`
This fails as it wants to invoke Dockerfiles in the staging directory.
The path of Dockerfile changed to build-command/target/docker/stage/0/Dockerfile. 
I'm not sure if I should just change test expectations or try to address it.
https://github.com/sbt/sbt-native-packager/blob/master/src/sbt-test/docker/build-command/build.sbt#L9

nigredo-tori:
> I'm not sure if I should just change test expectations or try to address it.

This test assumes the way the files are mapped to the staging directory, and we're changing that, so we should definitely change the expectations. Preferably in a way that eliminates that assumption. For example, we can replace

https://github.com/sbt/sbt-native-packager/blob/435c15551993f03bb9ae54cefec1957e68d499ab/src/sbt-test/docker/build-command/build.sbt#L8-L9

with

```scala
dockerBuildCommand := Seq(
  "docker", "build", "-t", "docker-build-command-test:0.1.0",
  (baseDirectory.value / "docker-test").getAbsolutePath
)
```

and move the `Dockerfile` to `docker-test` directory in the test project root.

However, this highlights two more potential issues. This change is breaking for people with custom build commands (only if they use a non-standard `Dockerfile` path) and, more importantly, for people with hand-written `Dockerfile`s. I'm not sure what we can do to address this without complicating the plugin even more.

ppiotrow:
To have it consistent with the past we can try to limit power of grouping function to include None as default layer.
`case class LayeredMapping(layerId: Option[Int], file: File, path: String)`
Then we define default grouping function to affect only /lib&/bin directories.

nigredo-tori:
> include None as default layer.

The thing is that, for existing `Dockerfile` to continue working, the default layer would need to be in the same directory as the `Dockerfile` (the root of the build context), where all the layers are, and it would not be easy to cleanly copy it. And there still remains some potential for breaking things (e.g. if someone copies the whole build context into their image).

If we're thinking long-term, it would make sense to just rip off the band-aid, and provide a consistent and predictable layering logic from the start. If that is unacceptable, we can provide a single opt-in (opt-out?) setting for the new layering behavior. The trade-off would be that we would need to maintain (and test) two versions of the same functionality. I defer to the maintainers about whether this would be worth the trouble.

ppiotrow:
@nigredo-tori I don't agree. It seems that layers directories on same level as main /opt directory provide with enough isolation to make accurate cuts.
Please see unit tests expected outputs as it looks like output is finally very clear now.
I bring back this convention of splitting `val files = dockerBaseDirectory.split(UnixSeparatorChar)(1)` to every copy command.

For me PR is feature complete and now ready for some refactor and final review.

muuki88:
Thanks for the good discussion :smiley: I tried to give my opinion on this, but my docker production knowledge is rather limited, so bare with me :grimacing: 

### stage dir and default layer

> The thing is that, for existing `Dockerfile` to continue working, the default layer would need to be in the same directory as the `Dockerfile` (the root of the build context), where all the layers are, and it would not be easy to cleanly copy it.

AFAIK if the user wants the old behaviour, then this would be enough

```scala
dockerLayerGrouping := (_ => None)
```
and everything will be layered out in staging as before. For me this would be what @nigredo-tori described as
_"we can provide a single opt-in (opt-out?) setting for the new layering behavior."_

> The trade-off would be that we would need to maintain (and test) two versions of the same functionality

The complexity from this PR seams reasonable to me, so I would accept the extra maintenance required to check layered and non-layered.

#854 introduce the failing build-command test. The purpose of this was not have a `stage` directory with a certain shape, but to configure the docker command to adjust to breaking changes in the CLI. So this test can be adapted as needed.

### Long term

> If we're thinking long-term, it would make sense to just rip off the band-aid, and provide a consistent and predictable layering logic from the start.

I'm open for suggestions :smile: This issue is open for years and this is the fourth attempt, which is looking promising to me. Docker changes so fast, I have lost faith in "long-term" solutions for docker :disappointed: as this constantly break or required lot of workarounds (server vs. api version, chowning, flags ).

## Conclusion

I would like to go for this solution as this solves a long standing pain point for docker users. Change the `stage` directory format is a price I'm willing to pay. My guess would be that there's quite a few people mess around with custom Dockerfiles and the stage directory due to the lack of a layering system in native-packager. As opting out is possible (or correct me if I'm wrong @ppiotrow ) the risk and migration costs are even lower for end users.

ppiotrow:
Hi, sorry for delays, I had unexpected offline time. I'm going to finish this PR in upcoming days.

> AFAIK if the user wants the old behaviour, then this would be enough
> dockerLayerGrouping := (_ => None)

That is true but need to be tested.

- [ ] write test to confirm user is able to opt-out from layering via `dockerLayerGrouping := (_ => None)`.
- [ ] write tests confirming that custom dockerPackageMappings works as before layering was introduced.
- [ ] final refactoring of code before merging.

ppiotrow:
I finished working on this PR. I've checked how people are using dockerPackageMappings on GitHub and created a test. Also opting-out is possible as proven in the test.
Documentation is up-to-date. I'm not really happy with makeCopy* methods as they repeat some code, but I wasn't able to refactor them better.
For me this is feature complete and reasonable code quality and tested.

muuki88:
Thanks a lot for all the time and passion you put into this @ppiotrow 
I'll review as soon as possible, but I'll surely merge this :heart:

muuki88:
Release `1.7.0` is on its way.

sideeffffect:
Hello guys
are you sure this feature is working as intended? I've run a small experiment, where I try to build 2 images in isolation, where only a tiny part of the code is different to see how much reuse/caching will take place.
From the results I see that actually no caching/re-use can happen, since the 2 images are different right from the layers `native-packager`/`demiourgos` creates.
I suspect that this is because of different timestamps in the layers' filesystem.

image

(visualization from `dive`)

ppiotrow:
Hi @sideeffffect, as I can see the first green layers came from your source image with JRE, right?
Then immediately you have difference in the first command that runs `id` and other user related command. I think that running same command from same parent layer should yield consistent layer. Have you run it on the same host?

I'm looking at the source code now, should not you have `USER root` at the beginning?
https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala#L188

Anyway, can you just upgrade your docker to support Multi Stage Builds?
It's intention is to prepare files and their permission and copy into final image without this permissions-related noise. This configuration for sure works best.

sideeffffect:
> I think that running same command from same parent layer should yield consistent layer.

I'm not sure it will work that way, at least not always
the previous layers are _the same_:
image

> Have you run it on the same host?

No, those were (hopefully) completely isolated instance, that was the point of the experiment.

> Anyway, can you just upgrade your docker to support Multi Stage Builds?

I can try that :)

ppiotrow:
It seems docker sha digest is not deterministic between different hosts. I've run several experiments with very simple docker files on my local node and remote machine. All gave inconsistent digests. Machines differ in the `docker --version`, but I don't think this is the issue here.
Dockerfile examples:
```
FROM anapsix/alpine-java:8
COPY test.txt /
# or ADD https://raw.githubusercontent.com/sbt/sbt-native-packager/master/README.md / 
# or RUN echo "hello"
```

I think we need to understand literally cache mechanism description.
> Starting with a parent image that is already in the cache, the next instruction is compared against all child images derived from that base image to see if one of them was built using the exact same instruction. If not, the cache is invalidated.

It will be cached only if there already exist parent-child relation between base layer and docker command. I admit this is surprise to me as I expected consistent behaviour. But this most likely has some sense in the Docker implementation.

You'll still benefit from newest [enchantment](https://medium.com/@ppiotrow/separate-docker-layers-for-dependencies-and-app-jars-e33b52ca044c) if you build your images on small pool of Docker hosts.

sideeffffect:
> But this most likely has some sense in the Docker implementation.

I think that what happens is that you copy files at different points in time, and the timestamps are reflected in the layer file system, thus having different content and hash.
That's why jibs (and the `sbt-jib` plugin) sets all file timestamps to 0 -- to make the builds/image layers 100 % deterministic.

ppiotrow:
I don't think timestamp makes digest deterministic between hosts.
See my example with `RUN echo ""` that doesn't even create a file.
Also JIB doesn't [mention](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#why-is-my-image-created-48-years-ago) reseting timestamps to boost caching.

jroper:
Was any consideration given to using `projectDependencyArtifacts` as the basis for the implementation of deciding group mappings? Unlike using the `organization` heuristic, which breaks if you have sub projects that don't use the same organization, and includes too much if you have external dependencies that are from the same organization as your build, basing the decision of inclusion in a layer using `projectDependencyArtifacts` would not be a heuristic, it robustly answers "is this artifact part of this build or not?"

The main thing though to using this is that since `dockerLayerGrouping` is a `SettingKey` it can't depend on `projectDependencyArtifacts` because that's a task, so `dockerLayerGrouping` has to be changed to a `TaskKey` for this to work. That's a non binary compatible change, but as I understand it, this change hasn't been included in a release yet, so we can still make that change.

Additionally, to make this even more robust, if the function was `(File, String) => Option[Int]`, then the matching could be done even more robustly, since the decision could be made based on the source file, rather than having to reconstruct the logic for building the path of the file.

Finally, any reason `String => Option[Int]` was chosen over `PartialFunction[String, Int]`? `PartialFunction[String, Int]` is much easier to compose, if you wanted to add a custom decision for some custom files you've added, it would be as simple as:

```scala
dockerLayerGrouping := dockerLayerGrouping.value.orElse {
  case ... => Some(...)
}
```

In contrast to now where you have to:

```scala
dockerLayerGrouping := path => dockerLayerGrouping.value(path).orElse {
  if (...) {
    Some(...)
  } else {
    None
  }
}
```

jroper:
I've submitted a PR with my suggestions here:

https://github.com/sbt/sbt-native-packager/pull/1326

ppiotrow:
Hi, I didn't know about `projectDependencyArtifacts` but it sounds very well suited for this task. That is true that inner source company artefacts are promoted as application code and I'm observing it in some of my projects.
Partial function should be fine, let me review the code in the evening.

PR #1314, opened 2020-03-08
Fix windows / test-bat-template

sshark:
* Use `Seq` for better manage command lines management.
* Cannot fix "include symbols with double q" because there is no way to escape '<' and '>' symbols during `cmd` execution. Is there a way to do so?
* Have to trim the argument for `include symbols on normal args` otherwise it won't pass the test.

### DIFF
--- a/src/sbt-test/windows/test-bat-template/build.sbt
+++ b/src/sbt-test/windows/test-bat-template/build.sbt
@@ -56,7 +56,6 @@ TaskKey[Unit]("checkScript") := {
     val pr = new StringBuilder()
     val logger = ProcessLogger((o: String) => pr.append(o + "\n"), (e: String) => pr.append("error < " + e + "\n"))
     val cmd = Seq("cmd", "/c", script.getAbsolutePath) ++ args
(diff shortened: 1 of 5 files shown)


src/sbt-test/windows/test-bat-template/build.sbt
```diff
@@ -49,13 +49,14 @@ TaskKey[Unit]("checkScript") := {
...
-                  args: String,
+                  args: Seq[String],
                   expected: String,
                   env: Map[String, String] = Map.empty,
                   expectedRC: Int = 0) = {
     val pr = new StringBuilder()
     val logger = ProcessLogger((o: String) => pr.append(o + "\n"), (e: String) => pr.append("error < " + e + "\n"))
-    val cmd = Seq("cmd", "/c", script.getAbsolutePath + " " + args)
+    val cmd = Seq("cmd", "/c", script.getAbsolutePath) ++ args
+//    val cmd = Seq("cmd", "/c", script.getAbsolutePath, "-Dtest.hoge=C:\\Program Files\\Java", "'C:\\Program Files\\Java'")
```
muuki88:
```suggestion
```

I guess this can be removed :wink:

PR #1326, opened 2020-04-06
Improved layer grouping

jroper:
The main purpose of this change is to make layer grouping more robust.  Instead of using a heuristic based on whether a library is in the same `organization` as the build (which is not always what you want, some builds may have sub projects with different organizations, and some builds may depend on external dependencies from the same organization), it checks whether the library in question is present in `projectDependencyArtifacts` (ie, whether its a dependency built by this build), if it is, then it goes in a higher layer.

Doing this change however required changing `dockerLayerGrouping` to be a `TaskKey` instead of a `SettingKey`, which is a non binary compatible change. So, `dockerLayerGrouping` has been deprecated, and a new `TaskKey`, `dockerGroupLayers` has been introduced.

Since a new task has been introduced, I also took advantage of that to improve the types, instead of just passing the destination path, which requires you to know how destination paths are built to check if a particular file should be in there, it passes the whole mapping, ie, the source file and the destination path, which means that you can match against the source file (which is what `projectDependencyArtifacts` gives you) as well. Also, I changed the type from a function that returns an `Option` to a `PartialFunction`, this makes composing it, to introduce new mappings, far simpler, as `PartialFunction[_, _]` composes trivially compared to `Function[_, Option[_]]`.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -98,19 +98,25 @@ object DockerPlugin extends AutoPlugin {
     ),
     dockerUpdateLatest := false,
     dockerAutoremoveMultiStageIntermediateImages := true,
-    dockerLayerGrouping := {
+    dockerLayerGrouping := { _: String =>
+      None
+    },
+    dockerGroupLayers := {
       val dockerBaseDirectory = (defaultLinuxInstallLocation in Docker).value
-      (path: String) =>
-        {
-          val pathInWorkdir = path.stripPrefix(dockerBaseDirectory)
-          if (pathInWorkdir.startsWith(s"/lib/${organization.value}"))
-            Some(2)
-          else if (pathInWorkdir.startsWith("/lib/"))
-            Some(1)
-          else if (pathInWorkdir.startsWith("/bin/"))
-            Some(1)
-          else None
-        }
+      // Ensure this doesn't break even if the JvmPlugin isn't enabled.
+      val artifacts = projectDependencyArtifacts.?.value.getOrElse(Nil).map(_.data).toSet
+      val oldFunction = (dockerLayerGrouping in Docker).value
+
+      // By default we set this to a function that always returns None.
+      val oldPartialFunction = Function.unlift((tuple: (File, String)) => oldFunction(tuple._2))
+
+      val libDir = dockerBaseDirectory + "/lib/"
+      val binDir = dockerBaseDirectory + "/bin/"
+
+      oldPartialFunction.orElse {
+        case (file, _) if artifacts(file)                                    => 2
+        case (_, path) if path.startsWith(libDir) || path.startsWith(binDir) => 1
+      }
     },
     dockerAliases := {
       val alias = dockerAlias.value
@@ -159,7 +165,8 @@ object DockerPlugin extends AutoPlugin {
       val multiStageId = UUID.randomUUID().toString
       val generalCommands = makeFromAs(base, "mainstage") +: makeMaintainer((maintainer in Docker).value).toSeq
       val stage0name = "stage0"
-      val layerIdsAscending = (dockerLayerMappings in Docker).value.map(_.layerId).distinct.sorted
+      val layerMappings = (dockerLayerMappings in Docker).value
+      val layerIdsAscending = layerMappings.map(_.layerId).distinct.sorted
       val stage0: Seq[CmdLike] = strategy match {
         case DockerPermissionStrategy.MultiStage =>
           Seq(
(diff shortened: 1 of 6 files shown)


src/sbt-test/docker/test-layer-groups/changes/nolayers.sbt
```diff
@@ -1 +1 @@
-dockerLayerGrouping in Docker  := (_ => None)
+dockerGroupLayers in Docker := PartialFunction.empty
```
muuki88, line 1: It would be nice to reflect this in the docs as well:

https://github.com/sbt/sbt-native-packager/blame/master/src/sphinx/formats/docker.rst#L147-L154

muuki88:
@ppiotrow any remarks from your side?

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -98,19 +98,25 @@ object DockerPlugin extends AutoPlugin {
...
-        }
+      // Ensure this doesn't break even if the JvmPlugin isn't enabled.
+      val artifacts = projectDependencyArtifacts.?.value.getOrElse(Nil).map(_.data).toSet
+      val oldFunction = (dockerLayerGrouping in Docker).value
+
+      // By default we set this to a function that always returns None.
+      val oldPartialFunction = Function.unlift((tuple: (File, String)) => oldFunction(tuple._2))
+
+      val libDir = dockerBaseDirectory + "/lib/"
+      val binDir = dockerBaseDirectory + "/bin/"
```
ppiotrow, line 114: could you also add "/conf/" to the same layer? I noticed that `/conf/application.ini` gets created when `javaOptions` are used.
```
      javaOptions in Universal := Seq(
      "-Djava.net.preferIPv4Stack=true",
      "-Dlogback.configurationFile=logback.packaged.xml"
      ),
```

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -98,19 +98,25 @@ object DockerPlugin extends AutoPlugin {
...
+
+      oldPartialFunction.orElse {
+        case (file, _) if artifacts(file)                                    => 2
```
ppiotrow, line 117: for me `artifacts.contains` would be more explicit.

src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala
```diff
@@ -57,10 +57,15 @@ private[packager] trait DockerKeysEx extends DockerKeys {
   lazy val dockerAdditionalPermissions =
     taskKey[Seq[(DockerChmodType, String)]]("Explicit chmod calls to some of the paths.")
   val dockerApiVersion = TaskKey[Option[DockerApiVersion]]("dockerApiVersion", "The docker server api version")
+  @deprecated("Use dockerGroupLayers instead", "1.7.1")
   val dockerLayerGrouping = settingKey[String => Option[Int]](
     "Group files by path into in layers to increase docker cache hits. " +
       "Lower index means the file would be a part of an earlier layer."
   )
+  val dockerGroupLayers = taskKey[PartialFunction[(File, String), Int]](
```
ppiotrow, line 65: I'm not very convinced to expose File as this function argument.
1. Not sure how plugin users will do matching on file, but most likely they end up comparing location/path
2. It will simplify chmod related code that will give up find.getOrElse and /dev/null and become simply layerToPath.lift(...)
3. To make it work, just simple change is necessary: sth like `.map(_.data.path)` in body of dockerGroupLayers

jroper: > Not sure how plugin users will do matching on file, but most likely they end up comparing location/path

I think there are use cases for either. With the path, the user has to know exactly the structure of the Docker image, including what that base directory is. The file I think makes more sense because it's the file that's changing, not the path. I know that this source file changes frequently, so every mapping that includes this source file, I want to put it in a higher layer. That's more direct and robust then having to work out what the destination path for that file is, and then matching based on that.

> It will simplify chmod related code that will give up find.getOrElse and /dev/null and become simply layerToPath.lift(...)

Yes, but that code isn't that complex, and I think the advantages outweigh the minuses here.

> To make it work, just simple change is necessary: sth like .map(_.data.path) in body of dockerGroupLayers

No, that's the point, that doesn't work, and that's why the `File` is being passed. If I want to match the slf4j api jar, the path for that is `/home/jroper/.ivy2/cache/org.slf4j/slf4j-api/jars/slf4j-api-1.7.30.jar`, meanwhile, the path that gets passed into the function is `opt/docker/lib/org.slf4j.slf4j-api-1.7.30.jar`. They are completely different paths, even the file name doesn't match. So, in order to match based on the path, I need to know the completely undocumented method that sbt native packager uses to construct that file name, I also need to know the completely undocumented method that sbt native packager uses to construct the the path to it. What if sbt native packager ever changes that logic? Then my logic for matching will break. In contrast, this logic will never break, no matter what sbt native packager does:

```scala
dockerGroupLayers := {
  val slf4jFile = for {
    artifact <- externalDependencyClasspath.value
    moduleId <- artifact.metadata.get(AttributeKey[ModuleID]("moduleID"))
    if moduleId.organization == "org.slf4j" && moduleId.name == "slf4j-api"
  } yield artifact.data
  dockerGroupLayers.value.orElse {
    case (`slf4jFile`, _) => Some(2)
  }
}
```

Perhaps a bit wordy, but if we find it's popular, we could always add a helper function to sbt and then they could use:

```scala
dockerGroupLayers := {
  val slf4jFile = getClasspathFileFor(externalDependencyClasspath.value, "org.slf4j", "slf4j-api")
  dockerGroupLayers.value.orElse {
    case (`slf4jFile`, _) => Some(2)
  }
}
```

muuki88:
```scala
dockerGroupLayers := {
  val slf4jFile = for {
    artifact <- externalDependencyClasspath.value
    moduleId <- artifact.metadata.get(AttributeKey[ModuleID]("moduleID"))
    if moduleId.organization == "org.slf4j" && moduleId.name == "slf4j-api"
  } yield artifact.data
  dockerGroupLayers.value.orElse {
    case (`slf4jFile`, _) => Some(2)
  }
}
```

maybe we can add this to the `docker.rst` docs as well?

muuki88:
@ppiotrow I'm going to merge this and we address the small hints in another pull request.

Do you have time for this :smile: ? If not, that's totally fine. I'm going to add the two notes.

ppiotrow:
Sure, I have plans to contribute more.

PR #1339, opened 2020-05-08
Add addApp function to align with bash version

olsio:
Closes https://github.com/sbt/sbt-native-packager/issues/862

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/ash-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/ash-template
@@ -32,6 +32,10 @@ addJava () {
   java_opts="$java_opts $1"
 }
 
+addApp () {
+  app_commands="$app_commands $1"
+}
+
 addResidual () {
   residual_args="$residual_args \"$1\""
 }
@@ -91,6 +95,7 @@ process_args () {
   fi
 }
 
+app_commands=""
 residual_args=""
 real_script_path="$(realpath "$0")"
 app_home="$(realpath "$(dirname "$real_script_path")")"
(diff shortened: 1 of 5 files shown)


lightbend-cla-validator:
Hi @olsio,

olsio:
Signed the CLA

muuki88:
Thanks a lot for adding this feature ♥️

It would be awesome to add one sbt-test to validate that this feature works with future changes.

https://github.com/sbt/sbt-native-packager/tree/master/src/sbt-test/ash/command-line-settings

You can checkout the [dev guidelines](https://github.com/sbt/sbt-native-packager/wiki/Developer-Guide) for pointers on how to run the tests.

olsio:
Sure will have a look.

olsio:
I added the test. It wouldn't build on MacOS so I had to find a Ubuntu machine first.

In general it might be a good idea to remove this plug-in all together and just have a single JavaAppPackaging version of the script without relying on bash. But I don't know what the implications of that would be.

muuki88:
Awesome :+1: Thanks for the additional test.

> In general it might be a good idea to remove this plug-in all together and just have a single JavaAppPackaging version of the script without relying on bash

From an engineering perspective I would agree to find the smallest common denominator and remove the additional stuff. My bare minimum bash knowledge is afraid of doing such a thing.

> But I don't know what the implications of that would be.

Me neither :joy: I have merged like 1-5 pull request each year adding a bit more "bash magic" to make certain use cases work. Bash is really string escaping hell. So at this point I'm okay with maintaining an ash and bash implementation as I have to find someone willing to backport all the weird bash hacks I don't quite understand :disappointed: 

So in general yes :heart: If you want to do it, I'll give you as much support as I can :hugs:

PR #1350, opened 2020-06-30
remove sbt 0.12 operators from documentation

mberndt123:
The sbt-native-packager docs use a lot of syntax that has been deprecated for years. Many of these will eventually end up in people's build files due to copy-pasting from the docs, so I think it's worth fixing.

It would also make sense to use sbt's unified scope syntax (`Universal / mappings` rather than `mappings in Universal`), but let's start with the less controversial things.


muuki88:
This is awesome! Thanks a lot for making the docs more accessible to sbt new comers :heart:

PR #1370, opened 2020-10-04
docker: make None the last layer group

pawitp:
Currently, if a user adds additional artifact to the Docker build
without modifying dockerGroupLayers, the artifact will have "None"
grouping and be put as the first layer in the output.

These additional artifacts are often configuration or supporting
assets. Making this the zeroth layer by default causes dependency
layer to not be re-used if any of these artifacts are changed.

Since not all users will care to modify dockerGroupLayers, the
default should be to make it such that adding additional artifact
does not break re-use of the dependency layer.


lightbend-cla-validator:
Hi @pawitp,

muuki88 (approved):
Thanks a lot for this improvement! :heart: 

Totally makes sense to have this order in this way.

muuki88:
`1.7.6` is on its way

PR #1374, opened 2020-10-14
Omit removed Java modules from detected jlinkModules

nigredo-tori:
We're already skipping detection for some Jakarta modules that have legacy `java.*` names. However, there is another source of information - the release notes for Java 11 have [a list](https://www.oracle.com/java/technologies/javase/jdk-11-relnote.html#JDK-8190378) of all removed modules. This PR filters these out as well. Although there is some overlap, I've opted to keep the lists from two sources separate, since the modules are filtered out for different reasons.

Regarding compatibility, this shouldn't impact most builds that use Java 11 and later. For such a build the removed modules could only have come from Jakarta artifacts and such, so they should be present in the project's classpath anyway.


nigredo-tori:
I've also added [`java.annotation`](https://mvnrepository.com/artifact/jakarta.annotation/jakarta.annotation-api/1.3.5) to the list of known Jakarta modules, since it came up in one of my projects.

muuki88 (approved):

PR #1376, opened 2020-10-21
Use github actions for automatic releases

muuki88:


muuki88 (commented):
I herby self-approve this :grimacing:

PR #1400, opened 2021-02-20
chown values overrides for Docker format

### LINKED ISSUE #1399: Allow override chown user and group in copy commands
Would be nice to have an option to use different user and group applied during `COPY` step. Same effect could be achieved if explicit chown command added, but it will create a new layer.

Why it could be useful. To apply hardening - to make copied files write protected by the user, which will run application (and so that it would not possible to change it by using `chmod` with the same user).

Currently we apply this via rewriting commands using map/flatMap, which is not really handy.

Proposal - add 2 more settings `chownUser` and `chownGroup` both `Option[String]`, and if they are set to `Some` value, then it will be used and in case of missing value it will fall back to `daemonUser` and `daemonGroup` respectively.

We need this applied for multistage docker, but I guess it is better to add the same logic to `ChownCopy` as well.

I will sketch a PR for this.

arixmkii:
Fixes #1399 

SettingKeys are added under Docker specific trait. I also added basic documentation about them being available.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -161,6 +161,8 @@ object DockerPlugin extends AutoPlugin {
       val uidOpt = (daemonUserUid in Docker).value
       val group = (daemonGroup in Docker).value
       val gidOpt = (daemonGroupGid in Docker).value
+      val cUser = (chownUser in Docker).value.getOrElse((daemonUser in Docker).value)
+      val cGroup = (chownGroup in Docker).value.getOrElse((daemonGroup in Docker).value)
       val base = dockerBaseImage.value
       val addPerms = dockerAdditionalPermissions.value
       val multiStageId = UUID.randomUUID().toString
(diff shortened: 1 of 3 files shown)


lightbend-cla-validator:
Hi @arixmkii,

src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala
```diff
@@ -68,4 +68,8 @@ private[packager] trait DockerKeysEx extends DockerKeys {
   )
   val dockerLayerMappings =
     taskKey[Seq[LayeredMapping]]("List of layer, source file and destination in Docker image.")
+  val chownUser =
+    SettingKey[Option[String]]("chown-user", "Optional alternative user to use in chown in COPY command")
+  val chownGroup =
+    SettingKey[Option[String]]("chown-group", "Optional alternative group to use in chown in COPY command")
```
muuki88, line 74:
```suggestion
  val dockerChownUser =
    SettingKey[Option[String]]("docker-chown-user", "Optional alternative user to use in chown in COPY command")
  val dockerChownGroup =
    SettingKey[Option[String]]("docker-chown-group", "Optional alternative group to use in chown in COPY command")
```

If the setting is specific to a packaging format it's prefixed with that format. In this case `docker<SettingKey>`. The configuration scoping is not necessary, so you can move it in the plugins definition to the other settings starting with docker.

src/sphinx/formats/docker.rst
```diff
@@ -99,6 +99,12 @@ Environment Settings
   ``daemonUser in Docker``
     The user to use when executing the application. Files below the install path also have their ownership set to this user.
 
+  ``chownUser in Docker``
```
muuki88, line 102:
```suggestion
  ``dockerChownUser``
```

src/sphinx/formats/docker.rst
```diff
@@ -99,6 +99,12 @@ Environment Settings
...
+    Optionally overrides user value used for applying to ``chown`` in ``COPY`` commands.
+
+  ``chownGroup in Docker``
```
muuki88, line 105:
```suggestion
  ``dockerChownGroup``
```

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -310,6 +312,8 @@ object DockerPlugin extends AutoPlugin {
       daemonUserUid := Some("1001"),
       daemonGroup := "root",
       daemonGroupGid := Some("0"),
+      chownUser := None,
+      chownGroup := None,
```
muuki88, line 316:
```suggestion
      dockerChownUser := None,
      dockerChownGroup := None,
```

See comment in the `Keys.scala`.

Also move up to the settings starting with `docker*`

muuki88 (changes requested):
Thanks a lot @arixmkii for your contributions :heart: 

I suggested a few changes. If possible it would be nice to have a sbt scripted test that show's this feature in action.

arixmkii:
@muuki88 Thank you for detailed feedback! I will update this PR on Thursday.

muuki88:
Looking forward 🐱

PR #1410, opened 2021-04-20
Update dockerGroupLayers documentation

onmomo:
Docs reflect the implementation since it's now based on JavaAppPackaging#projectDependencyArtifacts.
* https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/archetypes/JavaAppPackaging.scala#L59
* https://github.com/sbt/sbt-native-packager/blob/master/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala#L108


muuki88 (approved):
Thanks for making the docs better 😍

PR #1414, opened 2021-05-25
Only set executable permission on staged files if required and not set

DylanArnold:
When staging files, only set the target executable permission if the source file is executable and the target file is not executable.

It can be problematic if setting the executable permission on a target file multiple times if using inotify to watch staged files as inotify events will be triggered even when the permission is already set. This fix ensures the executable permission is only set once, to avoid unnecessary inotify events.

I am using development tools to watch the stage directory but currently they get triggered if I run the stage task even when there aren't any changes due to setting the executable permission unnecessarily.

To see this in action, assuming you had an example project with stage script at `example/target/universal/stage/bin/example`  you can run the following to see a demonstration of this problem.

In a terminal run:
`inotifywait -m example/target/universal/stage/bin/example`

In another terminal run the following:
`sbt example/stage`

You'll see this output from the first terminal:
```
example/target/universal/stage/bin/example ATTRIB 
example/target/universal/stage/bin/example OPEN 
example/target/universal/stage/bin/example ACCESS 
example/target/universal/stage/bin/example CLOSE_NOWRITE,CLOSE 
```

Run `sbt example/stage` again and you'll see the same output. 

If you run `chmod u+x example/target/universal/stage/bin/example` you will still see the same output which demonstrates that the stage task has the same behaviour as this.

After applying the change in this pull request, running `sbt example/stage` multiple times will only trigger inotify events the first time the script is created.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/Stager.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/Stager.scala
@@ -30,7 +30,9 @@ object Stager {
     // TODO - Config file user-readable permissions....
     for {
       (from, to) <- copies
-      if from.canExecute
+      // Only set executable permission if it needs to be set. Note: calling to.setExecutable(true) if it's already
+      // executable is undesirable for a developer using inotify to watch this file as it will trigger events again
+      if from.canExecute && !to.canExecute
     } to.setExecutable(true)
     stageDirectory
   }


lightbend-cla-validator:
Hi @DylanArnold,

src/main/scala/com/typesafe/sbt/packager/Stager.scala
```diff
@@ -30,7 +30,9 @@ object Stager {
     // TODO - Config file user-readable permissions....
     for {
       (from, to) <- copies
-      if from.canExecute
+      // Only set executable permission if it needs to be set. Note: calling to.setExecutable(true) if it's already
+      // executable is undesirable for a developer using inotify to watch this file as it will trigger events again
+      if from.canExecute && !to.canExecute
```
muuki88, line 35: What happens when `to` doesn't exist? Is there an existence check necessary?

muuki88 (commented):
Thanks for this improvement 😊

Ideally this change wouldn't be necessary as everything is properly cached, but that's another story.

May you explain why you are watching the staging environment?

DylanArnold:
Yes I did wonder if this could / should be solved before this code gets executed but I didn't really dig into and understand the cache implementation. I would be happy to try and dig into that if you think it makes sense.

I'm watching the staging environment because I'm using a tool called Tilt https://tilt.dev/ to continously deploy a development Kubernetes application. 

I use Tilt to build a docker image from the staging output (initially) but also to update my kubernetes pod **without** building and deploying the docker image again.

Tilt will perform a live update https://docs.tilt.dev/live_update_reference.html if the pod is already deployed by watching the staging environment for changes and copying the assets directly into the pod (without image rebuild / pod restart). I'm new to sbt-native-packager so there may be a different approach?, but I felt I wanted to avoid the overhead of any further zipping / packaging which will slow down my development cycle.

So the key thing given the above is that I want to avoid unnecessary live updates (and process restarts within my deployed pod) when the execute permission is set again.

src/main/scala/com/typesafe/sbt/packager/Stager.scala
DylanArnold: I did wonder that, and tested by removing the staged script. When running `stage` again, it correctly recreates and sets the execute permission. 

From the docs https://docs.oracle.com/javase/8/docs/api/java/io/File.html#canExecute-- 


> Returns:
> true if and only if the abstract pathname exists and the application is allowed to execute the file


So it seems it returns false if the pathname does not exist.

I did just notice this other comment on the `canExecute` docs though, which may be a slight concern? 

> On some platforms it may be possible to start the Java virtual machine with special privileges that allow it to execute files that are not marked executable.

I guess if this is the case, `canExecute` could return true, so the execute permission may never be set. I find it fairly unlikely / don't see why people would be running in this mode though. I'm not sure where to look for information on this?

muuki88 (approved):

PR #1415, opened 2021-06-08
Fix scope for rpmVendor

kardapoltsev:
I've fixed only `rpmVendor` because it's mentioned in the docs as a required setting as `rpmVendor in Rpm := "someVendor"`. But now it works only without `Rpm` scope. 

Do you think that we need to scope all other settings? I could do it if needed.

Also the default value `None` looks better for me instead of empty string. I'm not very familiar with RPM specs unfortunately


muuki88:
Thanks for your pull request 😊

The docs are wrong in this case. A setting is either prefixed ( e.g. `rpmVendor` ) or scope ( e.g. `Rpm / target` )

kardapoltsev:
Got it. I'll try to fix docs then

PR #1419, opened 2021-06-17
Migration to maven central / sonatype

muuki88:
* New groupId `io.github.sbt.sbt-native-packager`
* https://issues.sonatype.org/browse/OSSRH-70270


build.sbt
```diff
@@ -1,5 +1,5 @@
 name := "sbt-native-packager"
-organization := "com.typesafe.sbt"
+organization := "io.github.sbt.sbt-native-packager"
```
muuki88, line 2: @eed3si9n sorry for pinging your directly :pray: I would like to push sbt native packager under these new coordinates and opened a [jira ticket on sonatype OSSRH-7020](https://issues.sonatype.org/browse/OSSRH-70270). 

I'm required to create a new repository. So there are a few options

1. I get temporary access to create a repo
2. I close the ticket and move to a domain I can control
3. Keep the old coordinates and someone who controls typesafe.com adds the DNS TXT entry

What do you think?

eed3si9n: 1. I can create a repo assuming that would do
2. or temporarily give you the repo creation rights
3. or we can setup sbt-ci-release on https://repo1.maven.org/maven2/com/github/sbt/,  which I control, so you can release using tag push.

muuki88: Option 1 would be ideal I guess.

What would option 3 imply?

eed3si9n: Option 3 means using `com.github.sbt` as opposed to `io.github.sbt.sbt-native-packager`, and it also means that you have to have the release process automated via sbt-ci-release.

muuki88: Oh nice. That makes option three a lot easier.

`sbt-ci-release` is added in this PR. I'll change the maven coordinates.

eed3si9n: I've configured the Sonatype credentials so when the org name is change it should now work.

muuki88, line 2:
```suggestion
organization := "com.github.sbt"
```

PR #1425, opened 2021-06-23
Optimize docker layers

412b:
- add classpath jar to artifacts
- move start scripts to the same layer as artifacts, because those contain references to artifacts
- move /jre (JLink) to a separate layer
- move /conf to a separate layer as well, because it's not something changing often


muuki88:
Thanks for these optimization 🤩 I'll take a deeper look on Monday

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -105,18 +106,30 @@ object DockerPlugin extends AutoPlugin {
     dockerGroupLayers := {
       val dockerBaseDirectory = (defaultLinuxInstallLocation in Docker).value
       // Ensure this doesn't break even if the JvmPlugin isn't enabled.
-      val artifacts = projectDependencyArtifacts.?.value.getOrElse(Nil).map(_.data).toSet
+      var artifacts = projectDependencyArtifacts.?.value.getOrElse(Nil).map(_.data).toSet
+
+      ClasspathJarPlugin.autoImport.packageJavaClasspathJar.?.value match {
```
muuki88, line 111: This is probably the easiest way to achieve this :+1: 

I would add a comment why the DockerPlugin cross references another plugin that is unrelated at first sight:

```suggestion
      // add the classpath jar to the project artifacts to improve layer caching as it is created by
      // the ClasspathJarPlugin and is not part of the projectDependencyArtifacts
      ClasspathJarPlugin.autoImport.packageJavaClasspathJar.?.value match {
```

muuki88 (approved):
This is an amazing improvement. Thanks for the thoughtful pull request :heart:

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
412b: Added that comment :smile:

muuki88 (approved):
I'll merge and release ASAP :heart:

PR #1430, opened 2021-07-02
Don't hard-code xz compression option "-9e". If compression level was…

### LINKED ISSUE #1426: Custom options for XZ Tarball not working
## Expected behaviour

Accoridng to [**manual**](https://www.scala-sbt.org/sbt-native-packager/formats/universal.html#configurations) it should be possible to set the XZ options to be used by *packageXzTarball* via:
> `universalArchiveOptions in (Universal, packageXzTarball)`

So I added this to my "build.sbt" file:
```
universalArchiveOptions in (Universal, packageXzTarball) := Seq("-3")
```

**Expected behavior:**

Intead of the packageXzTarball *default* XZ options, which are `-9e` (extremely slow!), it should now be using option `-3`.

## Actual behaviour

It is easy to verify, e.g. via htop, that packageXzTarball still is using `xz -9e`, rather than the desired `xz -3` 😞 

This completey locks up my system due to immense memory usage!

It appears that option `universalArchiveOptions in (Universal, packageXzTarball)` is ***ignored*** or documentation is wrong.

Or am I doing something wrong?

## Information

- universal:packageXzTarball
- 1.8.0
- Linux (Ubuntu 20.04)
- XZ
- xz (XZ Utils) 5.2.5
- Linux

dEajL3kA:
Don't hard-code compression option **`-9e`** in the xz command-line used by SBT. 

If the compression options weren't hard-coded, the user would be able to set the desired compression options via environment variable `XZ_OPT`. But, currently, since **`-9e`** is hard-coded, the options in `XZ_OPT` will always be overwritten by the actual xz command-line! Also **`-9e`** is *not* a good default to begin with. Level `-9` is very slow and needs a whole lot memory (can cause OOM killer to kill your build process!) for probably only a little improvement in compression ratio, compared to the *default* compression level. Furthermore, and even worse, option **`-e`** easily doubles compression time for marginal improvement in compression ratio (if at all). Option **`-e`**,  that's really something you'd use if don't care about compression time at all. IMHO it really makes much more sense to just use the xz defaults and let the user override them, via `XZ_OPT` environment variable, if desired.

...either that, or give the use a way to set the xz compression options in the `build.sbt` file (please see my [ticket](#1426) for details).

-----------------

**Form the xz manpage:**
* On compression levels:  
  > Unless you want to maximize the compression ratio, you probably don't want a higher preset level than -7 due to speed and memory usage.
* On the '-e' option:  
  > Modify the compression preset (-0 ... -9) so that a little bit better compression ratio can be achieved without increasing memory usage of the compressor or decompressor. The downside is that the compression time will increase dramatically (it can easily double).

https://linux.die.net/man/1/xz

-----------------

This fixes #1426.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/universal/Archives.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/universal/Archives.scala
@@ -200,7 +200,7 @@ object Archives {
     */
   def xz(f: File): File = {
     val par = f.getParentFile
-    sys.process.Process(Seq("xz", "-9e", "-S", ".xz", f.getAbsolutePath), Some(par)).! match {
+    sys.process.Process(Seq("xz", /*"-9e",*/ "-S", ".xz", f.getAbsolutePath), Some(par)).! match {
       case 0 => ()
       case n => sys.error("Error xz-ing " + f + ". Exit code: " + n)
     }


lightbend-cla-validator:
Hi @dEajL3kA,

src/main/scala/com/typesafe/sbt/packager/universal/Archives.scala
```diff
@@ -200,7 +200,7 @@ object Archives {
     */
   def xz(f: File): File = {
     val par = f.getParentFile
-    sys.process.Process(Seq("xz", "-9e", "-S", ".xz", f.getAbsolutePath), Some(par)).! match {
+    sys.process.Process(Seq("xz", /*"-9e",*/ "-S", ".xz", f.getAbsolutePath), Some(par)).! match {
```
muuki88, line 203:
```suggestion
    sys.process.Process(Seq("xz", "-S", ".xz", f.getAbsolutePath), Some(par)).! match {
```

Just remove ;)

muuki88 (approved):
I commented a more generic solution #1426 , but this will do for the time being as you mentioned it can still be configured via env variables

PR #1437, opened 2021-07-15
prepend dockerCommands

### LINKED ISSUE #1417: additions to dockerCommands are run as user
For instance: to add packages to the base image you need to be root.

Worked around this by flatMapping the dockerCommands and inserting an extra item.

```
dockerCommands := dockerCommands.value.flatMap {
  case Cmd("USER", args@_*) if args.contains("1001:0") => Seq(
    Cmd("RUN", "apk add --no-cache ttf-dejavu bash"),
    Cmd("USER", args: _*)
  )
  case cmd => Seq(cmd)
}
```

Perhaps this could be made somewhat easier/prettyer.

frankivo:
Trying to fix #1417. Not sure if this is they way, feel free to comment :)

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -222,6 +222,7 @@ object DockerPlugin extends AutoPlugin {
           case Some(_) => Seq(makeUser("root"), makeUserAdd(user, group, uidOpt, gidOpt))
           case _       => Seq()
         }) ++
+      (Docker / dockerCommandsPrepend).value ++
         Seq(makeWorkdir(dockerBaseDirectory)) ++ {
         (strategy match {
           case DockerPermissionStrategy.MultiStage =>
--- a/src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala
@@ -46,6 +46,7 @@ trait DockerKeys {
   val dockerRmiCommand =
     SettingKey[Seq[String]]("dockerRmiCommand", "Command for removing the Docker image from the local registry")
 
+  val dockerCommandsPrepend = TaskKey[Seq[CmdLike]]("dockerCommandsPrepend", "List of docker commands that are run first in the final stage.")
   val dockerCommands = TaskKey[Seq[CmdLike]]("dockerCommands", "List of docker commands that form the Dockerfile")
 }
 


lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

frankivo:
Usage: 
`dockerCommandsPrepend := Seq(Cmd("RUN", "apk add --no-cache ttf-dejavu bash"))`

lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

muuki88:
Not sure if a separate tasks is the way to go here. Prepending (not inserting into an arbitrary position) should be doable with

```scala
dockerCommands := List("foo", "bar") ++ dockerCommands.value
```

Maybe this is something for the documentation?

frankivo:
> 
> 
> Not sure if a separate tasks is the way to go here. Prepending (not inserting into an arbitrary position) should be doable with
> 
> ```scala
> dockerCommands := List("foo", "bar") ++ dockerCommands.value
> ```
> 
> Maybe this is something for the documentation?

That will break the build as the commands will be before "FROM".
I do agree that the insert position is somewhat arbitrary.

lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

muuki88:
Thanks @frankivo for keeping this PR up-to-date.

I'm still unsure that this setting is a good idea. It may cause more confusion that help and creates a precedence for settings that insert docker commands in arbitrary positions.

The main issue from what I understand is finding the right spot to insert aditional docker commands. What do you think of

- adding [`Comment`](https://github.com/sbt/sbt-native-packager/blob/68600dfb60327aceb619cb9138d3c8f6e8504510/src/main/scala/com/typesafe/sbt/packager/docker/dockerfile.scala#L84) lines to the docker file with fixed names
- provide helpers to insert at this fixed "marks"

For example

```Dockerfile
FROM openjdk:8 as stage0
# stage0:after:from
WORKDIR /opt/docker
COPY opt /opt
USER root
RUN ["chmod", "-R", "u=rX,g=rX", "/opt/docker"]

FROM openjdk:8
# image:after:from
WORKDIR /opt/docker
# image:after:workdir
COPY --from=stage0 --chown=daemon:root /opt/docker /opt/docker
USER 1001
ENTRYPOINT []
CMD []
```

and the 

```sbt
dockerCommands := prependDockerCommandAfter(dockerCommands.value, Docker.ImageWorkdir, List(...))
```


this is not super elegant. But  you get the idea. This is a more extendible approach. If you have a better API, feel free :smile:

frankivo:
@muuki88 I like that idea actually. Don't have much time this week, but I'll see if I can update this PR with your suggestion.

lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

frankivo:
I implemented your suggestions @muuki88. We could add more, but first let me know what you think.

muuki88:
Thanks a lot. I'll take a look ASAP. 🤗❤️

muuki88:
I have not forgotten this pull request 😘😘

Issue #1257, opened 2019-08-26
Tini support for Docker images


jroper:
Running Java as PID 1 has some issues, for example, jstack and jstat don't work when the PID of the Java process is 1. The solution to this is to not run Java as PID 1, but then you need something to run as PID 1 that will forward SIGTERM and other signals to it when the process is shut down, otherwise the process will never be shutdown cleanly, and will always be killed with a SIGKILL after a timeout - bash doesn't forward SIGTERM for example.

The best solution is to use [tini](https://github.com/krallin/tini), a very small process that acts as an init process, forwarding signals and collecting zombies. sbt-native-packager could provide a flag that enables adding and using tini.

A few notes:

* The common way of using tini is to do an ADD from the URL to download tini. This means every time the image is built, tini needs to be redownloaded, so it's checksum can be checked to determine whether the layer needs to be rerun. Not the worst thing because tini is only 10KB, but it might be better to fetch tini from another image, then that image can be cached.
* The tini signature should probably be checked, or at least that should be allowed as an option. Though this does result in more layers being built. A multistage build of course can remedy this.

muuki88:
Thanks @jroper for the detailed feature request :smiley: 

I general I have no issues with adding this feature. Usually I start with suggesting a separate archetype plugin (like the AshScriptPlugin or LauncherPlugin) that adds the required steps to the`dockerCommands`.

>  Not the worst thing because tini is only 10KB, but it might be better to fetch tini from another image, then that image can be cached.

Agreed. Either use an image that includes tiny or we ship a version of tiny in native-packager or we use the `RUN apk add --no-cache tini` command for alpine linux distributions.

cbrunnkvist:
I don't think added it via the distributions' package is required anymore, since [Docker already bundles `tini`](https://github.com/krallin/tini#using-tini) and sets it to PID 1 when [the `--init` param](https://docs.docker.com/engine/reference/commandline/run/) is used.

jroper:
Perhaps might make sense to add a setting to enable the flag, eg `dockerBundleInit`, that if true, adds the `--init` flag so users don't have to manually add it to `dockerBuildOptions`. I would even go as far to say that this should default to `true`, since I don't see a reason not to use it for any Java application, and I can't imagine adding it breaking anything, except for my projects, where I've hacked the the `dockerCommands` to add tini manually:

```scala
    dockerCommands := {
      val oldCommands = dockerCommands.value
[...]

PR #1438, opened 2021-07-16
Added dockerBuildInit option

jroper:
See #1257.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -152,9 +152,10 @@ object DockerPlugin extends AutoPlugin {
     ).toOption
       .map(_.trim)
       .flatMap(DockerApiVersion.parse),
+    dockerBuildInit := false,
     dockerBuildOptions := Seq("--force-rm") ++ dockerAliases.value.flatMap { alias =>
       Seq("-t", alias.toString)
-    },
+    } ++ { if (dockerBuildInit.value) List("--init") else Nil },
     dockerRmiCommand := dockerExecCommand.value ++ Seq("rmi"),
     dockerBuildCommand := dockerExecCommand.value ++ Seq("build") ++ dockerBuildOptions.value ++ Seq("."),
     dockerAdditionalPermissions := {
--- a/src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala
@@ -68,7 +68,7 @@ private[packager] trait DockerKeysEx extends DockerKeys {
   )
   val dockerLayerMappings =
     taskKey[Seq[LayeredMapping]]("List of layer, source file and destination in Docker image.")
-   val dockerBuildInit = SettingKey[Boolean](
+  val dockerBuildInit = SettingKey[Boolean](
     "dockerBuildInit",
     "Whether the --init flag should be passed to Docker when building. " +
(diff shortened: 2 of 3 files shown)


src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -152,9 +152,10 @@ object DockerPlugin extends AutoPlugin {
     ).toOption
       .map(_.trim)
       .flatMap(DockerApiVersion.parse),
+    dockerBuildInit := false,
```
muuki88, line 155: Maybe we can use the `dockerVersion` or `dockerApiVersion` to set this true when possible?

muuki88 (dismissed):

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
jroper: Possible, however, `dockerVersion` and `dockerApiVersion` are task keys. `dockerBuildOptions` and `dockerBuildCommand` are setting keys, and so can't depend on `dockerVersion` or `dockerApiVersion`. We'd have to depend on this in `publishLocal`, which I think would be bad because it would mean `dockerBuildCommand` would no longer be the source of truth as to what was executed.

src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala
```diff
@@ -68,4 +68,10 @@ private[packager] trait DockerKeysEx extends DockerKeys {
   )
   val dockerLayerMappings =
     taskKey[Seq[LayeredMapping]]("List of layer, source file and destination in Docker image.")
+   val dockerBuildInit = SettingKey[Boolean](
```
muuki88, line 71: I moved those to the `DockerKeysEx`, to avoid binary incompatibilities

muuki88 (approved):

src/main/scala/com/typesafe/sbt/packager/docker/Keys.scala
```diff
@@ -68,4 +68,10 @@ private[packager] trait DockerKeysEx extends DockerKeys {
...
+    "dockerBuildInit",
+    "Whether the --init flag should be passed to Docker when building. " +
+      "Setting to true will cause Docker to bundle a tini in the container, to run as the init process, which is recommended for JVM apps. " +
+      "Requires Docker API version 1.25+"
+  )
```
muuki88, line 76:
```suggestion
  val dockerBuildInit = SettingKey[Boolean](
    "dockerBuildInit",
    "Whether the --init flag should be passed to Docker when building. " +
      "Setting to true will cause Docker to bundle a tini in the container, to run as the init process, which is recommended for JVM apps. " +
      "Requires Docker API version 1.25+"
  )
```

muuki88 (approved):

guizmaii:
Hi,

Tried to activate this new option and rebuild my image but I get the following error:

```
[error] unknown flag: --init
[error] See 'docker buildx build --help'.
[info] Removing intermediate image(s) (labeled "snp-multi-stage-id=035fd77f-20c7-432f-ae31-93913656b499") 
[info] Total reclaimed space: 0B
[error] stack trace is suppressed; run last app-embedded-docker / Docker / publishLocal for the full output
[error] (app-embedded-docker / Docker / publishLocal) Nonzero exit value: 125
[error] Total time: 51 s, completed 31 Aug. 2021, 9:57:13 pm
```

I'm on Mac, is it only available on Linux? 🤔

I have an up-to-date installation of Docker:
Screen Shot 2021-08-31 at 10 00 02 pm

PR #1440, opened 2021-07-22
Use only Unix file separators when generating Unix scripts

ghik:
When generating Unix startup scripts, `scriptClasspath` always uses Unix file separators (`/`). Therefore, `File.separator` cannot be used because the build environment may be on another system (Windows). We must always use Unix file separator when working with Unix paths.

This problem was discovered by me when I tried to build a Linux docker image on Windows. Bash startup script had wrong classpath generated.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/AshScriptPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/archetypes/scripts/AshScriptPlugin.scala
@@ -109,7 +109,7 @@ object AshScriptPlugin extends AutoPlugin {
     private[this] def makeClasspathDefine(cp: Seq[String]): String = {
       val fullString = cp map (
         n =>
-          if (n.startsWith(File.separator)) n
+          if (n.startsWith("/")) n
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:
Hi @ghik,

muuki88:
Thanks for your patience :hugs: 

I wonder if this has any other impacts :thinking: 

- building `zip` distributions on linux for linux should work the same :+1: 
- building `zip` distributions on windows for linux :thinking: , will the path be properly set? If the `scriptClasspath` uses the unix separators, this should also be broken right now, right?
- building `docker` distributions on linux should work the same :+1: 
- building `docker` distributions` on windows should now work :+1:

ghik:
Sorry for the late response, I was on vacation.

My changes only affect Bash & Ash start script generation.
As far as I understand, these start scripts are used only on Linux and are shared by `zip` and `docker` distributions. Therefore, this should fix the issue for both `zip` and `docker` while not breaking anything.

muuki88 (approved):
Well then 💪❤️

PR #1442, opened 2021-07-29
bump scala-parser-combinators to 2.0.0

arw357:


lightbend-cla-validator:
Hi @arw357,

mkurz:
@muuki88 We need that merged and release so we can upgrade scala-parser-combinators in Play: https://github.com/playframework/playframework/pull/10901

muuki88 (approved):

muuki88:
Sure sure 😃 will push a release later

mkurz:
Actually merging this pull request was a mistake:
sbt-native packager is a sbt-plugin. sbt-plugins [get published for Scala 2.12](https://repo1.maven.org/maven2/com/github/sbt/sbt-native-packager_2.12_1.0/).
The 2.12.x scala compiler depends on 1.x versions for both scala-parser-combinators and scala-xml, see https://github.com/scala/scala/blob/2.12.x/versions.properties#L21-L22
If we would use 2.x versions, we would run into annoying dependency conflicts in sbt 1.5+
There is no need to upgrade these versions to 2.x, the only reason would be if we would cross build for Scala 3, but sbt-plugins do not need to be cross build for Scala 3.

See discussion in https://github.com/scala/scala/pull/9743

PR #1453, opened 2021-08-28
Double slash fix for ash-template

darabos:
This is the change from #1451 applied to `ash-template`. I don't know much about Ash, but it works on my machine.


kardapoltsev:
LGTM, thanks @darabos!

kardapoltsev (approved):

PR #1455, opened 2021-08-31
Downgrade scala-parser-combinators

mkurz:
See explanation here: https://github.com/sbt/sbt-native-packager/pull/1442#issuecomment-909040300
Discussion in the scala repo: https://github.com/scala/scala/pull/9743
Same PR for twirl: https://github.com/playframework/twirl/pull/439

BTW I upgraded scala-xml to scala to 1.3.0, the latest release in the 1.x series, which is totally fine here (because we stay on 1.x)

@muuki88 I am pretty sure this needs to be done.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -40,8 +40,12 @@ libraryDependencies ++= {
     case "2.10" => Nil
     case _ =>
       Seq(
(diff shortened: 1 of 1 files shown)


muuki88 (approved):
Totally fine. I'm currently on vacation and can't release. Feel free to push a tag with the next patch version 🤗

PR #1465, opened 2021-10-08
Fix #1464

### LINKED ISSUE #1464: With the latest version (1.9.5), `Docker / version` is detected as "not used" by sbt
## Expected behaviour

We have some definitions for our Docker build:

```scala
object BuildHelper {
  lazy val dockerSettings = Seq(
    Docker / maintainer       := "...",
    Docker / daemonUser       := "...",
    Docker / dockerRepository := Some("..."),
    Docker / packageName      := packageName.value,
    dockerUpdateLatest        := false,
    dockerExposedPorts        := Seq(8080),
    dockerBaseImage           := "ubuntu:latest",
    Docker / version          := s"$release-$version",  // this is the line 63 that sbt talks about
  )
}
```

I just updated to the latest version of sbt-native-packager and since the update sbt tells me this:

```scala
[warn] there's a key that's not used by any other settings/tasks:
[warn]  
[warn] * app-build / Docker / version
[warn]   +- BuildHelper.scala:63
```

I had a quick look at the changes between `1.9.4` and `1.9.5`. It could be because of this change: 
https://github.com/sbt/sbt-native-packager/compare/v1.9.4...v1.9.5#diff-b866751bc91854625b6a5d821ae6a4df02736ceb3581779d81319caee1d9d2cdR100

This `Option((version in Docker).value)` piece of code have been changed for `Option((version).value)`.
Should it be `Option((Docker / version).value)` instead?

cc @frankivo

guizmaii:
Fixes https://github.com/sbt/sbt-native-packager/issues/1464

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ Global / onChangedBuildSource := ReloadOnSourceChanges
 Global / scalaVersion := "2.12.12"
 
 // crossBuildingSettings
-crossSbtVersions := Vector("0.13.17", "1.1.6")
+crossSbtVersions := Vector("1.1.6")
 
 Compile / scalacOptions ++= Seq("-deprecation")
 javacOptions ++= Seq("-source", "1.8", "-target", "1.8")


muuki88:
Time to remove the 0.13.x support....

muuki88:
1.9.6 is on its way

PR #1473, opened 2021-11-11
use proper native image command on windows

OlegYch:
previous fix assumed native-image.cmd from path won't work, but it seems to work fine


muuki88 (approved):

muuki88:
Thanks for making the life of windows devs easier :heart:

PR #1474, opened 2021-12-03
Remove unused private methods (makeComment).

takayahilton:


muuki88 (approved):
Thanks for your PR. Creating a single argument case class should really not be too hard :sweat_smile:

PR #1498, opened 2022-02-11
Update web site to have a correct addSbtPlugin

crakjie:
addSbtPlugin was missing `%` that was leading to unresolved dependencies


muuki88:
Thanks for the PR. I'm not sure if this is correct. This works as described in the documentation

SethTisue:
yeah, I think this should be closed

I don't know how to account for whatever behavior OP experienced, but the second `%` isn't needed or recommended in general for `addSbtPlugin`, so we'd need evidence that this particular occurrence is somehow different

PR #1500, opened 2022-02-16
Fix missing Graalvm native-image binary symlink in generated container image

### LINKED ISSUE #1492: GraalVMNativeImagePlugin produces broken docker images without native-image in the PATH
Try to enable  `GraalVMNativeImagePlugin` on a project and set

```
   graalVMNativeImageGraalVersion := Some("22.0.0.2"),
```

This would produce a broken docker image:

```
❯ docker run -ti --rm ghcr.io-graalvm-graalvm-ce-native-image:22.0.0.2
docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "native-image": executable file not found in $PATH: unknown.
```

```
❯ docker run -ti --rm --entrypoint /bin/bash ghcr.io-graalvm-graalvm-ce-native-image:22.0.0.2
bash-4.4# native-image
bash: native-image: command not found
bash-4.4# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
bash-4.4# find / -name "native-image" -print
/opt/graalvm-ce-java17-22.0.0.2/bin/native-image
/opt/graalvm-ce-java17-22.0.0.2/lib/svm/bin/native-image
bash-4.4# exit
```

kgston:
Closes #1492
Solution based on the monkeypatch provided by [pshirshov](https://github.com/pshirshov)
I decided to use a wildcard interpolation to get the correct folder for the symlink since there will only be one version of graalvm in the image anyway and trying to derive it from `graalVMNativeImageGraalVersion` which could contain variations on the base image version, would be difficult to parse accurately.

Also updates the graalvm base image version in the tests to the latest, otherwise the existing test would report a false positive using the old version.


muuki88 (approved):

muuki88:
THanks @kgston for fixing this long standing issue :heart:

muuki88:
1.9.9 is on its way

PR #1508, opened 2022-06-27
use modern sbt syntax in doc

SethTisue:
* instead of `foo:bar`, use `Foo / bar`
* instead of `foo in Bar`, use `Bar / foo`
* instead of `foo in (Bar, baz)`, use `Bar / baz / foo`

surely I missed at least a few occurrences 🤷‍♂️


muuki88 (approved):
Thanks a lot @SethTisue ☺️

muuki88 (approved):

PR #1512, opened 2022-07-26
Add sbt-vspp for publishing the SBT plug-in in a Maven-consistent format

ScalaWilliam:
This is to enable usage of sbt-native-packager in Enterprise environments where only the valid-POM format is accepted for JAR downloads, enabling the very powerful sbt-native-packager for thousands of developers.

More background here: https://github.com/esbeetee/sbt-vspp/blob/main/README.md


muuki88 (approved):
Thanks 😊 let's give it a try

PR #1522, opened 2022-09-29
True Multi OS support with `pathSeparator`

Rojods:
The previous version had the ":" hard-coded as a `module-path` separator, which might be fine on Linux(es) and Macs, but fails on Windows. I suggest we simply put `File.pathSeparator` instead and rely on the JVM to tell us the right path separator character. Then there is also less burden on this project to keep the path separator code up-to-date.


lightbend-cla-validator:
Hi @Rojods,

Rojods:
I am just writing here to say I've signed the CLA.

muuki88:
Thanks a lot for your contribution ☺️

muuki88 (approved):

PR #1546, opened 2023-05-23
Fixes remote debug connections closing

jackcviers:
sbt-native-packager Java App archetype remote debugging is broken when using `-jvm-debug <port>` in the docker entrypoint. It will start with `-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=<port>`, which will terminate the connection prematurely.

Adding the * before the port will allow all remote connections to connect to the debugger port, which was the old default behavior.

See https://bugs.openjdk.org/browse/JDK-8175050


muuki88:
Oh boy :smile: Thanks for fixing this. Will this break JDK 8 users (surely no one is still running this :stuck_out_tongue_winking_eye:  )

muuki88 (approved):

PR #1547, opened 2023-05-27
Use SOURCE_DATE_EPOCH for zip, tar, xz, gzip, rpm, msi

stevedlawrence:
SOURCE_DATE_EPOCH is a standardized environment variable that can be set to create reproducbile builds:

  https://reproducible-builds.org/docs/source-date-epoch/

If the SOURCE_DATE_EPOCH environment variable is defined, its value is used during the creation of zip, tar, gzip, xz, rpm, and msi files in places where timestamp information is embedded in the resulting file.

To support this capability, a new sourceDateEpoch file utility is added which executes the touch command to set the mtime of the provided file/directory and all children. This utility is called on all source files for the listed packages.

For zip, this also adds the -o option to native zip, and calls setTime() for the non-native zip.

For rpm, this also sets the %use_source_date_epoch_as_buildtime and %clamp_mtime_to_source_date_epoch tunables to 1 to ensure the build time and timestamps embedded in the RPM are set to the epoch.

For msi, the SOURCE_DATE_EPOCH variable is also used to generate reproducible name-based UUIDs, using a combination of an identifier and the epoch value. Note that this does not change the "Product" GUID or the build time that is embedded in the MSI. There does not seem to be a way to change these, so MSI builds are not 100% reproducible. However, with these changes the output of tools like msidiff is much smaller and easier to confirm similar builds.

If SOURCE_DATE_EPOCH is not set then the current behavior is used. Archives and rpms use timestamps of when the files are modified/created, and MSI uses random GUIDs.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/FileUtil.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/FileUtil.scala
@@ -1,5 +1,6 @@
 package com.typesafe.sbt
 package packager
+import sbt._
 
 import java.io.File
 import java.nio.file.Files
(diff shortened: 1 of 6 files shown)


lightbend-cla-validator:
Hi @stevedlawrence,

muuki88 (approved):
Thanks a lot @stevedlawrence for this contribution. This is super useful.

PR #1549, opened 2023-07-24
Adds cross platform architecture compilation for native image

### LINKED ISSUE #1443: How to building native image for ARM
I would like to build GraalVM native image for ARM architecture from my AMD machine.
It such feature currently supported by native-packager ?

## Information

- What sbt-native-packager are you using -> 1.8.1
- What sbt version -> 1.5.2
- What is your build system  -> MacOs AMD64
- What package are you building -> docker
- What version has your build tool -> set
- What is your target system -> any with ARM base CPU

kgston:
Closes #1443

This attempts to fill in the gap in cross platform architecture native image packaging. Added a new key `graalVMNativeImagePlatformArch: Option[String]` to specify the target platform architecture. Currently it only supports a single platform as doing multi platform would involve a much bigger change in the code base.

While this works technically, the performance is still rather lacking. When I crossed compiled (amd64 -> arm64) compile times took ~19x longer. The simple hello world test took over 5 minutes to produce a binary. 

The 5+ minutes test case also triggered a bug in [SBT #6771](https://github.com/sbt/sbt/issues/6671) causing the test to be aborted halfway through, but was recently fixed in SBT 1.9.3 hence the upgrade. Also updated the tests to use the `--no-fallback` option as I noticed that it was generating a fallback image when running the tests.

Also, it was a little tricky to correctly detect if an existing container was in the correct platform architecture when `graalVMNativeImagePlatformArch` is not defined. I used a docker image label to get around that since docker doesn't provide a way to report the host platform (using `uname -m` isn't ideal since it doesn't match docker's convention). The best it does is provide a special `local` value to signal building the image in the same platform architecture as the host.

### DIFF
--- a/.github/workflows/validate-pr.yml
+++ b/.github/workflows/validate-pr.yml
@@ -254,7 +254,7 @@ jobs:
     - name: Setup GraalVM environment
       uses: olafurpg/setup-scala@v10
       with:
-        java-version: graalvm@22.0.0=tgz+https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-22.0.0.2/graalvm-ce-java11-linux-amd64-22.0.0.2.tar.gz
+        java-version: graalvm@22.3.2=tgz+https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-22.3.2/graalvm-ce-java17-linux-amd64-22.3.2.tar.gz
     - name: Install native-image
       run: gu install native-image
     - name: Validate
--- a/src/main/scala/com/typesafe/sbt/packager/graalvmnativeimage/GraalVMNativeImagePlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/graalvmnativeimage/GraalVMNativeImagePlugin.scala
@@ -37,6 +37,7 @@ object GraalVMNativeImagePlugin extends AutoPlugin {
     target in GraalVMNativeImage := target.value / "graalvm-native-image",
     graalVMNativeImageOptions := Seq.empty,
     graalVMNativeImageGraalVersion := None,
+    graalVMNativeImagePlatformArch := None,
     graalVMNativeImageCommand := (if (scala.util.Properties.isWin) "native-image.cmd" else "native-image"),
     resourceDirectory in GraalVMNativeImage := sourceDirectory.value / "graal",
     mainClass in GraalVMNativeImage := (mainClass in Compile).value
@@ -48,7 +49,7 @@ object GraalVMNativeImagePlugin extends AutoPlugin {
     resources := resourceDirectories.value.descendantsExcept(includeFilter.value, excludeFilter.value).get,
     UniversalPlugin.autoImport.containerBuildImage := Def.taskDyn {
       graalVMNativeImageGraalVersion.value match {
-        case Some(tag) => generateContainerBuildImage(s"$GraalVMBaseImage:$tag")
+        case Some(tag) => generateContainerBuildImage(s"$GraalVMBaseImage:$tag", graalVMNativeImagePlatformArch.value)
         case None      => Def.task(None: Option[String])
       }
     }.value,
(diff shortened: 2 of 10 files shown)


lightbend-cla-validator:
Hi @kgston,

kgston:
The CI checks seems to be unhappy that I added a new parameter for specifying the target platform architecture.

.github/workflows/validate-pr.yml
```diff
@@ -254,7 +254,7 @@ jobs:
     - name: Setup GraalVM environment
       uses: olafurpg/setup-scala@v10
```
guizmaii, line 255: This action is deprecated. I think it'd be better to migrate to https://github.com/graalvm/setup-graalvm

kgston:
@muuki88 Could you take a look at the PR when you find the time? Thanks.

src/main/scala/com/typesafe/sbt/packager/graalvmnativeimage/GraalVMNativeImagePlugin.scala
```diff
@@ -166,35 +186,62 @@ object GraalVMNativeImagePlugin extends AutoPlugin {
...
+      if (!buildContainerExists) {
         streams.log.info(s"Generating new GraalVM native-image image based on $baseImage: $imageName")
 
         val dockerContent = Dockerfile(
           Cmd("FROM", baseImage),
           Cmd("WORKDIR", "/opt/graalvm"),
           ExecCmd("RUN", "gu", "install", "native-image"),
           ExecCmd("RUN", "sh", "-c", "ln -s /opt/graalvm-ce-*/bin/native-image /usr/local/bin/native-image"),
-          ExecCmd("ENTRYPOINT", "native-image")
+          ExecCmd("CMD", "native-image")
```
muuki88, line 231: I will never get the difference in my head. Why is this change from entrypoint to cmd needed?

muuki88 (approved):
Thanks for this high quality contribution 🙏

While I can't really give feedback on the graalvm implementation, the changes and tests look very good.

Only the entrypoint vs cmd is what I don't get 🫣 I'll merge if you give a short explanation here that I can reference if someone else wants to change it back 😁

muuki88:
> The CI checks seems to be unhappy that I added a new parameter for specifying the target platform architecture.

IIRC there's another, private trait where those parameters can be added. If not, I'm fine with adding those binary compatibility breakages to the mima exceptions.

kgston:
> > The CI checks seems to be unhappy that I added a new parameter for specifying the target platform architecture.
> 
> IIRC there's another, private trait where those parameters can be added. If not, I'm fine with adding those binary compatibility breakages to the mima exceptions.

Hmm, I don't know if its exactly breaking binary compatibility since the old parameters still work the same as the old versions. According to the CI, it just says that the new param that I added doesn't exist on the older version, which is to be expected. 

Is there a better, more appropriate way to specify the target platform?

muuki88:
I'm referencing this validation error: https://github.com/sbt/sbt-native-packager/actions/runs/5718905995/job/15495686495#step:7:1

which indicates binary compatibility issues. What are you referencing to?

kgston:
I've fixed the binary compat issue and all the subsequent issues that popped up from the graalvm test. Other unrelated tests are failing, and I assume it is a side effect from upgrading SBT. I'll try to debug the issue when I find some time, but taking a quick look at the github action logs, it is not immediately obvious what's the issue. Any hints would be appreciated, otherwise I might have to create a separate PR for upgrading to SBT 1.9.3 to try and isolate the issue.

muuki88:
triggered a rerun .. the appveyor thing seems super weird. 

Do the tests fail on your local machine?

kgston:
I tried running `validateDebian`, `validateDocker` and `validateWindows` on `master` branch and I do get the same errors as with on the Github Actions too.. So it might be unrelated to the SBT 1.9.3 upgrade after all? 🤷‍♂️

PR #1564, opened 2024-01-02
DockerPlugin publishTask: switch --push flag to --output=type=registry

zachdecook:
`--push` is shorthand for `--output=type=registry` (https://docs.docker.com/engine/reference/commandline/buildx_build/#push)

By specifying `--output=type=registry`, sbt can support when `docker` is a symlink to `podman`.


lightbend-cla-validator:
Hi @zachdecook,

zachdecook:
This is incorrect. (it fixes a symptom of the problem, but podman requires a second command to be run to push it).

Probably it's best fixed by patching podman.

Issue #1555, opened 2023-10-12
No releases for v1.9.17 or v1.9.18 tags


stevedlawrence:
Tags have been created for v1.9.17 and v1.9.18 but no releases have been created, in either the GitHub releases (https://github.com/sbt/sbt-native-packager/releases) or published to the maven repos (https://central.sonatype.com/artifact/com.github.sbt/sbt-native-packager)

It looks like the github actions failed due to a gpg issue with sbt-ci-release:

v1.9.17: https://github.com/sbt/sbt-native-packager/actions/runs/5641807357/job/15280486680

v1.9.18: https://github.com/sbt/sbt-native-packager/actions/runs/5642779942/job/15283307724

ekrich:
Looks like those job logs need to a gist in the future.

TheElectronWill:
What can be done to get the last version of sbt-native-packager for the last version of sbt? Should I add a dependency on the Github repo?

dwickern:
@muuki88 do you mind if I tag a new release?

muuki88:
@dwickern please do ❤️❤️

dwickern:
@muuki88 I'm trying to figure out the release workflow. It looks like we make a release by pushing a tag like `v1.9.19`. Unfortunately I don't have permission to push tags.

muuki88:
The issue seems to be that the release doesn't work anymore

```
[info] gpg: no default secret key: No secret key
[info] gpg: signing failed: No secret key
[error] java.lang.RuntimeException: Failure running 'gpg --batch --pinentry-mode loopback --passphrase *** --detach-sign --armor --use-agent --output /home/runner/work/sbt-native-packager/sbt-native-packager/target/scala-2.12/sbt-1.0/sbt-native-packager_2.12_1.0.pom.asc /home/runner/work/sbt-native-packager/sbt-native-packager/target/scala-2.12/sbt-1.0/sbt-native-packager_2.12_1.0.pom'.  Exit code: 2
[error] 	at scala.sys.package$.error(package.scala:30)
[error] 	at com.jsuereth.sbtpgp.CommandLineGpgSigner.sign(PgpSigner.scala:74)
[error] 	at com.jsuereth.sbtpgp.PgpSettings$.$anonfun$signingSettings$2(PgpSettings.scala:151)
[error] 	at scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:293)
[...]

PR #1566, opened 2024-01-30
Upgrade sbt, scala and sbt-ci-release to avoid deprecated repo.scala-sbt.org

mkurz:
I started to block [repo.scala-sbt.org](http://repo.scala-sbt.org/) and [repo.typesafe.com](http://repo.typesafe.com/) in my `/etc/hosts`.

See
- https://github.com/sbt/sbt/issues/7202

this repo could go down again.

Unfortunately you still pull in a total outdated version of sbt-ci-release which is only hosted on repo.scala-sbt.org.
Also sbt < 1.3.0 has some artifacts still hosted on that repo, so eventually I upgraded to 1.9.x+
Also there is a newer Scala patch release available already...

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,5 +1,5 @@
-addSbtPlugin("com.typesafe.sbt" % "sbt-ghpages" % "0.6.3")
-addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "1.4.0")
+addSbtPlugin("com.github.sbt" % "sbt-ghpages" % "0.8.0")
+addSbtPlugin("com.github.sbt" % "sbt-site-sphinx" % "1.7.0")
 
 // releasing
 addSbtPlugin("com.github.sbt" % "sbt-ci-release" % "1.5.12")
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -2,7 +2,7 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-ghpages" % "0.6.3")
 addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "1.4.0")
 
 // releasing
-addSbtPlugin("com.geirsson" % "sbt-ci-release" % "1.5.7")
+addSbtPlugin("com.github.sbt" % "sbt-ci-release" % "1.5.12")
 
 libraryDependencies += "org.scala-sbt" %% "scripted-plugin" % sbtVersion.value
 
--- a/src/sbt-test/windows/test-bat-template/build.sbt
+++ b/src/sbt-test/windows/test-bat-template/build.sbt
@@ -114,18 +114,16 @@ TaskKey[Unit]("checkScript") := {
   checkOutput("include symbols on -D", Seq("-Dtest.hoge=\\[]!< >%"), "property(test.hoge) is [\\[]!< >%]\nSUCCESS!")
   checkOutput("include symbols on normal args", Seq("\"\\[]!< >%\""), "arg #0 is [\\[]!< >%]\nSUCCESS!")
 
-  /* fails test because symbols '<' and '>' cannot be properly escaped during cmd execution
   checkOutput(
     "include symbols with double quote",
     Seq("-Dtest.huga=\"[]!<>%\""),
     "property(test.huga) is [[]!<>%]\nSUCCESS!"
   )
-  */
 
   checkOutput(
     "include symbols with double quote2",
-    Seq("-Dtest.hoge=\\[]!< >%", "\"\\[]!< >%\"", "-Dtest.huga=\\[]!<>%"),
-    "arg #0 is [\\[]!< >%]\nproperty(test.hoge) is [\\[]!< >%]\nproperty(test.huga) is [\\[]!<>%]\nSUCCESS!"
+    Seq("-Dtest.hoge=\\[]!< >%", "\"\\[]!< >%\"", "-Dtest.huga=\\[]!%"),
+    "arg #0 is [\\[]!< >%]\nproperty(test.hoge) is [\\[]!< >%]\nproperty(test.huga) is [\\[]!%]\nSUCCESS!"
   )
 
   // can't success include double-quote. arguments pass from Process(Seq("-Da=xx\"yy", "aa\"bb")) is parsed (%1="-Da", %2="xx\"yy aa\"bb") by cmd.exe ...
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ target/
 .ensime*
 .bloop/*
 .metals/*
+.bsp/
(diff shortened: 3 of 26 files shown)


project/plugins.sbt
```diff
@@ -1,8 +1,10 @@
-addSbtPlugin("com.typesafe.sbt" % "sbt-ghpages" % "0.6.3")
-addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "1.4.0")
+addSbtPlugin("com.github.sbt" % "sbt-ghpages" % "0.8.0")
+addSbtPlugin(
+  "com.github.sbt" % "sbt-site" % "1.5.0" // TODO: sbt-site-sphinx, see https://github.com/sbt/sbt-site/issues/186
+)
```
mkurz, line 4: @muuki88 Unfortunately, sbt-site dropped the sphinx module, at least temporary (which is needed for [the docs](https://github.com/sbt/sbt-native-packager/tree/master/src/sphinx) I guess), at least until someone steps up to revive it:
- https://github.com/sbt/sbt-site/issues/186

What do you think? Should someone invest time to fix it (not sure if that is hard, maybe it's an easy fix) or should sbt-native-packager move to some other documentation format?

mkurz:
Depends on
- https://github.com/sbt/sbt-site/pull/214

dwickern:
sbt-site 1.6.0 was just released with Sphinx integration

https://github.com/sbt/sbt-site/releases/tag/v1.6.0

project/plugins.sbt
mkurz: Fixed by https://github.com/sbt/sbt-site/pull/214

mkurz:
@muuki88 This pull request is now ready., can you take a look please if you find time? Thanks!

dwickern:
The AppVeyor test failure looks new 🤔

mkurz:
> The AppVeyor test failure looks new 🤔

Yeah, this one seems new, I will take a look when I find time...

dwickern:
I fixed the failing Docker and GraalVM tests on the main branch if you merge/rebase again.

The AppVeyor test failure doesn't seem related to your changes. I had some similar errors last time I ran those tests on my Windows machine. I suspect that your changes invalidated the cache and that's why we're seeing failures now.

mkurz:
@dwickern I rebased, but the same error occurs again. Do you have a windows machine you can test what's wrong here?

mkurz:
Or maybe the AppVeyor cache neeeds to be cleaned?

mkurz:
I found the problem, will push soon...

src/sbt-test/windows/test-bat-template/build.sbt
```diff
@@ -114,18 +114,16 @@ TaskKey[Unit]("checkScript") := {
...
     "property(test.huga) is [[]!<>%]\nSUCCESS!"
   )
-  */
 
   checkOutput(
     "include symbols with double quote2",
-    Seq("-Dtest.hoge=\\[]!< >%", "\"\\[]!< >%\"", "-Dtest.huga=\\[]!<>%"),
-    "arg #0 is [\\[]!< >%]\nproperty(test.hoge) is [\\[]!< >%]\nproperty(test.huga) is [\\[]!<>%]\nSUCCESS!"
+    Seq("-Dtest.hoge=\\[]!< >%", "\"\\[]!< >%\"", "-Dtest.huga=\\[]!%"),
+    "arg #0 is [\\[]!< >%]\nproperty(test.hoge) is [\\[]!< >%]\nproperty(test.huga) is [\\[]!%]\nSUCCESS!"
```
mkurz, line 126: So, I found what made the test fail on AppVeyor running Windows.
For the record, the error message was:
```
error < > was unexpected at this time.
```

Turns out, this has nothing to do with this pull request per se, you can easily reproduce this error even with sbt-native-packager 1.9.16 on the current main branch, without this PR applied:
1. You have to be on Windows
1. Checkout the current `main` branch 
1. Go to the failing scripted test's directory:
  ```
  cd src/sbt-test/windows/test-bat-template
  ```
1. The test succeeds with sbt 1.5.8:
  ```
  sbt --sbt-version 1.5.8 -Dproject.version=1.9.16
  sbt:windows-test> stage
  sbt:windows-test> checkScript
  # The test works
  ```
1. The test fails starting with sbt 1.6.0-M1:
  ```
  sbt --sbt-version 1.6.0-M1 -Dproject.version=1.9.16
  sbt:windows-test> stage
  sbt:windows-test> checkScript
  # The test fails with above error message
  ```

I did check and debug: The process args and environment passed [here](https://github.com/sbt/sbt-native-packager/blob/v1.9.19/src/sbt-test/windows/test-bat-template/build.sbt#L59) are 100% the same for both sbt 1.5.8 and 1.6.0-M1, nothing changed... So why does sbt 1.6+ fail?
Before we go deeper, the technical reason why it fails is because `<` and `>` are handled like normal pipes starting with sbt 1.6+ and not passed as string.
So why is that?
Let's look at the Windows specific code that runs a process, [this line](https://github.com/AdoptOpenJDK/openjdk-jdk8u/blob/master/jdk/src/windows/classes/java/lang/ProcessImpl.java#L389). You can see there is a check for a `SecurityManager`, and based on its existence things [behave differently](https://github.com/AdoptOpenJDK/openjdk-jdk8u/blob/master/jdk/src/windows/classes/java/lang/ProcessImpl.java#L393).
So if a `SecurityManager` is available, the process will be started differently: the process args will be handled and escaped differently.
Now.. it turns out until sbt 1.5.8, sbt did install a `SecurityManager`, but starting with sbt 1.6.0-M1 it does not anymore.
Look at [this commit](https://github.com/sbt/sbt/commit/0413727796adebdb25d9fdf70ea038723f354ee6) in sbt, you will see in `Main.scala` the `TrapExit.installManager()` got removed.
So this explains why the test started to fail with sbt 1.6.+

Now the thing is, assuming most people upgraded to sbt 1.6 or newer already anyway, this test now just got adjusted to the current real world sbt behaviour.
The reason why the scripted test fails with this pull request is because I upgraded sbt to 1.9.9 and by default the scripted tests are run with the sbt version of the main project. Like showed above however, this behaviour already surfaces in any project that uses sbt-native-packager with sbt 1.6+.
(I mean in theory I could make the scripted test here keep running with sbt 1.5.4, so I wouldn't have to adjust the test itself, but that does not make any sense IMHO).

mkurz:
@muuki88 This is totally ready to merge now :+1: 
Also hopefully this unblocks #1555 now so a new release can be cut.

src/sbt-test/windows/test-bat-template/build.sbt
dwickern: Wow, TIL. I agree it doesn't make sense to depend on deprecated `SecurityManager` behavior.

mkurz: > Wow, TIL.

Yeah, TIL too, took me hours to figure out what the hell is going on...

muuki88 (approved):

muuki88 (approved):

muuki88:
Thanks so much @dwickern and @mkurz for all the work and fixes you pushed in the last weeks 🙏🙏

I try to push a release ASAP. Or @dwickern you do it ☺️

dwickern:
I will try 🤞

PR #1334, opened 2020-04-30
Fix argument passing in ash template

bentucker:
Fixes #1332

PR #1303, opened 2020-02-03
Fix ash template

gregsymons:
This fixes #1302, and, in a much simpler way than #1276, #1264. Unfortunately, I don't understand the tests well enough to figure out where to hang a test for the bashisms specifically, nor the `addResidual` change. If you can give me some tips, I'd be glad to write them if you want me to.

PR #1592, opened 2024-04-05
fix support for top-level `main` methods in Scala 3

mberndt123:
app_mainclass is currently not escaped correctly. This leads to problems in Scala 3 when you define `main` as a function at the top level (as opposed to inside an `object`. In this case, the main class generated by the Scala compiler will contain $ symbols that need to be escaped.

The `LauncherJarPlugin` abuses `mainClass` to set the `-jar foo.jar` parameter instead of a main class, so this case needs to be treated specially (I find this an ugly hack but haven't found a cleaner solution).
The `scripted-docker` test failures seem to be unrelated to my changes as they also occur in [other CI runs](https://github.com/sbt/sbt-native-packager/actions/runs/8395125752)

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/ash-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/ash-template
@@ -129,4 +129,4 @@ java_cmd="$(get_java_cmd)"
 # If a configuration file exist, read the contents to $opts
 [ -f "$script_conf_file" ] && opts=$(loadConfigFile "$script_conf_file")
 
-eval "exec $java_cmd $java_opts -classpath $app_classpath $opts $app_mainclass $app_commands $residual_args"
+exec $java_cmd $java_opts -classpath $app_classpath $opts $app_mainclass $app_commands $residual_args
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/ash-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/ash-template
@@ -94,7 +94,7 @@ process_args () {
      -java-home) require_arg path "$1" "$2" && jre=`eval echo $2` && java_cmd="$jre/bin/java" && shift 2 ;;
 
  -D*|-agentlib*|-agentpath*|-javaagent*|-XX*) addJava "$1" && shift ;;
-                                         -J*) addJava "${1:2}" && shift ;;
+                                         -J*) addJava "$(printf %s "$1"  | sed s/^..//)" && shift ;;
                                            *) addResidual "$1" && shift ;;
     esac
   done
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/ash-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/ash-template
@@ -44,7 +44,7 @@ addApp () {
 }
 
 addResidual () {
-  residual_args="$residual_args \"$1\""
+  residual_args="$residual_args '$(printf %s "$1" | sed "s/'/'\\\\''/")'"
 }
 
 # Allow user to specify java options. These get listed first per bash-template.
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/ash-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/ash-template
@@ -44,7 +44,7 @@ addApp () {
 }
 
 shellEscape () {
-  printf \'%s\' "$(printf %s "$1" | sed "s/'/'\\\\''/")"
+  printf "'%s'" "$(printf %s "$1" | sed "s/'/'\\\\''/")"
 }
 
 addResidual () {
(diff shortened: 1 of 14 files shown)


lightbend-cla-validator:
At least one commit author (matthias.berndt@ttmzero.com) is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

lightbend-cla-validator:
Hi @mberndt123,

mberndt123:
OK, this doesn't work because the `LauncherJarPlugin` doesn't actually use the `mainClass` setting to store a class name but instead a string like `-jar foo.jar`.

lightbend-cla-validator:
At least one commit author (matthias.berndt@ttmzero.com) is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

lightbend-cla-validator:
At least one commit author (matthias.berndt@ttmzero.com) is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/ash-template
```diff
@@ -129,4 +129,4 @@ java_cmd="$(get_java_cmd)"
 # If a configuration file exist, read the contents to $opts
 [ -f "$script_conf_file" ] && opts=$(loadConfigFile "$script_conf_file")
 
-eval "exec $java_cmd $java_opts -classpath $app_classpath $opts $app_mainclass $app_commands $residual_args"
+exec $java_cmd $java_opts -classpath $app_classpath $opts $app_mainclass $app_commands $residual_args
```
mberndt123, line 132: I don't know why `eval` was used here. It breaks `app_mainclass` if it contains `$` symbols, so I've removed it and it doesn't seem to break any tests, so… I guess it's fine?

dwickern: #1334 added `checkComplexResidual` but never called it from the [test file](https://github.com/sbt/sbt-native-packager/blob/9e009b0acd9f5d83c58174052ce9511293399047/src/sbt-test/ash/command-line-settings/test). If we fix that, hopefully we'll see why `eval` was added.

lightbend-cla-validator:
At least one commit author (matthias.berndt@ttmzero.com) is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/ash-template
mberndt123: OK, I've added that to the test file. Ironically, the tests still pass, and they fail if I put the `eval`  back in. I think everything here is broken, both the way "residual args" are being handled and the test that is supposed to test them.

dwickern: I agree, I'm not confident that those tests work. Digging into the blame, there's a lot of history for that line of code going back to #1334, #1303, #1095 and before.

If you call `myProgram --param1 "value1 value2"`, does it correctly pass two arguments (arg[0]="--param1", arg[1]="value1 value2")?

mberndt123: I've now fixed the tests and the escaping of residual args, so it should all be passed through correctly now. In fact I've added tests for all kinds of nasty things that might confuse a shell script and they're all passed through correctly.

lightbend-cla-validator:
At least one commit author (matthias.berndt@ttmzero.com) is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/ash-template
```diff
@@ -43,8 +43,12 @@ addApp () {
   app_commands="$app_commands $1"
 }
 
+shellEscape () {
+  printf "'%s'" "$(printf %s "$1" | sed "s/'/'\\\\''/")"
+}
+
 addResidual () {
-  residual_args="$residual_args \"$1\""
+  residual_args="$residual_args $(shellEscape "$1")"
```
dwickern, line 51: If I change this line to `residual_args="$residual_args $1"` and get rid of `shellEscape`, then all the tests still pass for me locally. Am I doing something wrong, or is `shellEscape` not actually needed here? I'm running scripted tests with `scripted ash/*`.

mberndt123: Hey @dwickern,

I'm afraid that's not what I see on my machine. When I remove the `shellEscape` stuff, the `scripted ash/command-line-settings` test fails. When I put it back in, it succeeds again.

dwickern (approved):
LGTM

mberndt123:
> LGTM

Thanks @dwickern, is there anything else I need to do to get this merged?

muuki88 (approved):

PR #1594, opened 2024-04-09
Add `dockerBuildEnvVars` setting, fix docker test failure in CI

### LINKED ISSUE #1518: Option to enable Docker buildkit
Docker buildkit can be enabled by setting the environment variable `DOCKER_BUILDKIT` to `1`. Sometimes, someone might be using buildkit features as part of the build, and need this to always be enabled. It would be great if sbt native packager could provide an option to enabled this automatically (ie, by setting that environment variable when it runs docker).

dwickern:
[Docker BuildKit](https://docs.docker.com/build/buildkit) became the default in Docker v23.0 which caused the test to fail. BuildKit doesn't expose intermediate containers as images, so the cleanup setting `dockerAutoremoveMultiStageIntermediateImages` is not relevant when running BuildKit.

I've added a setting `dockerBuildEnvVars` which can be used to disable BuildKit:
```sbt
dockerBuildEnvVars += "DOCKER_BUILDKIT" -> "0"
```
I've also improved BuildKit detection-- it now checks the docker version if `DOCKER_BUILDKIT` is not set.

This incidentally fixes #1518

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -156,6 +156,17 @@ object DockerPlugin extends AutoPlugin {
     dockerBuildOptions := Seq("--force-rm") ++ dockerAliases.value.flatMap { alias =>
       Seq("-t", alias.toString)
     } ++ { if (dockerBuildInit.value) List("--init") else Nil },
+    dockerBuildEnvVars := Map.empty,
+    dockerBuildkitEnabled := {
+      dockerBuildEnvVars.value
+        .get("DOCKER_BUILDKIT")
+        .orElse(sys.env.get("DOCKER_BUILDKIT"))
+        .map(_ == "1")
+        .getOrElse {
+          // BuildKit is the default since v23.0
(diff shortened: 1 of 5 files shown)


src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -156,6 +156,17 @@ object DockerPlugin extends AutoPlugin {
     dockerBuildOptions := Seq("--force-rm") ++ dockerAliases.value.flatMap { alias =>
       Seq("-t", alias.toString)
     } ++ { if (dockerBuildInit.value) List("--init") else Nil },
+    dockerBuildEnvVars := Map.empty,
+    dockerBuildkitEnabled := {
+      dockerBuildEnvVars.value
+        .get("DOCKER_BUILDKIT")
+        .orElse(sys.env.get("DOCKER_BUILDKIT"))
+        .map(_ == "1")
```
dwickern, line 164: Docker uses [strconv.ParseBool](https://pkg.go.dev/strconv#ParseBool) which accepts some other values, but we'll assume the value is either `1` or `0`.

muuki88 (approved):
Awesome 😎 thanks a lot

PR #1608, opened 2024-04-27
fix shellEscape function in ash-template

mberndt123:
It currently only escapes the first `'` character, but they should all be escaped.

How are shell scripts still a thing in 2024?


dwickern (approved):

dwickern:
Thanks!

Issue #1523, opened 2022-10-13
AshScriptPlugin generates script without function


sanyer:
## Expected behaviour

/opt/docker/bin/<script> works

## Actual behaviour

/opt/docker/bin/<script>: line 71: require_arg: not found

## Information

- What sbt-native-packager are you using
1.3.20
- What sbt version
1.5.5
- What is your build system (e.g. Ubuntu, MacOS, Windows, Debian )
debian
- What package are you building (e.g. docker, rpm, ...)
docker
- What is your target system (e.g. Ubuntu 16.04, CentOS 7)
alpine

muuki88:
Does this also happen in the latest version? There has been quite some changes to the ash script @sanyer

sanyer:
I give it a try @muuki88

PR #1630, opened 2024-08-22
Fix the -jvm-debug and -h flags in the ash template

qwe2:
This should fix issues like https://github.com/sbt/sbt-native-packager/issues/1523

### DIFF
--- a/.github/.scala-steward.conf
+++ b/.github/.scala-steward.conf
@@ -1,0 +1,3 @@
+updates.pin = [
+  { groupId = "org.scala-lang", artifactId = "scala-library", version = "2.12." }
+]
--- a/build.sbt
+++ b/build.sbt
@@ -30,7 +30,7 @@ classpathTypes += "maven-plugin"
 libraryDependencies ++= Seq(
   // these dependencies have to be explicitly added by the user
   "com.spotify" % "docker-client" % "8.16.0" % Provided,
-  "org.vafer" % "jdeb" % "1.12" % Provided artifacts Artifact("jdeb", "jar", "jar"),
+  "org.vafer" % "jdeb" % "1.13" % Provided artifacts Artifact("jdeb", "jar", "jar"),
   "org.apache.commons" % "commons-compress" % "1.27.1",
   // for jdkpackager
   "org.apache.ant" % "ant" % "1.10.15",
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ Global / onChangedBuildSource := ReloadOnSourceChanges
 
 // crossBuildingSettings
 lazy val scala212 = "2.12.20"
-lazy val scala3 = "3.6.2"
+lazy val scala3 = "3.6.3"
 Global / scalaVersion := scala3
 crossScalaVersions := Seq(scala3, scala212)
 (pluginCrossBuild / sbtVersion) := {
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -2,7 +2,7 @@ addSbtPlugin("com.github.sbt" % "sbt-ghpages" % "0.8.0")
 addSbtPlugin("com.github.sbt" % "sbt-site-sphinx" % "1.7.0")
 
 // releasing
-addSbtPlugin("com.github.sbt" % "sbt-ci-release" % "1.6.1")
+addSbtPlugin("com.github.sbt" % "sbt-ci-release" % "1.9.2")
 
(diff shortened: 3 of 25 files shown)


lightbend-cla-validator:
At least one commit author (igansperger@cloudera.com, igansperger@cloudera.com) is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

lightbend-cla-validator:
At least one commit author (igansperger@cloudera.com, igansperger@cloudera.com) is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

lightbend-cla-validator:
Hi @qwe2,

qwe2:
Interestingly the `addDebugger` syntax from [src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/bash-template](https://github.com/sbt/sbt-native-packager/blob/4ce488721a1c6864795f682449e03050195047b8/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/bash-template) is only compatible with Java 11+, but the GH workflow is still pulling JDK 1.8. [java_version_check](https://github.com/sbt/sbt-native-packager/blob/4ce488721a1c6864795f682449e03050195047b8/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/bash-template#L287) also checks for Java 8, not 11. Not sure if this is intentional...

muuki88:
Thanks for tackling this 💪

I'll be on vacation . Maybe @dwickern can take a look 😊

qwe2:
I've been a bit busy and wasn't sure what to do about Java 8 compat, but I ended up adding some checks so that `-jvm-debug` works on both `ash` and `bash`  for Java 8 as well.

muuki88:
Hi @qwe2 

Sorry for the waiting - I'll try to find some time to take a deeper look!

dwickern:
Re: Java 8 compatibility, I found the relevant change in the [JDK 9 release notes](https://www.oracle.com/java/technologies/javase/9-relnotes.html#JDK-8041435) (bold mine)

> #### JDWP socket connector accept only local connections by default 
>The JDWP socket connector has been changed to bind to localhost only if no ip address or hostname is specified on the agent command line. A hostname of asterisk (*) may be used to achieve the old behavior which is to bind the JDWP socket connector to all available interfaces; **this is not secure and not recommended**.
> JDK-8041435

IMO we should keep the secure defaults and let users decide what address to bind to:
`-jvm-debug 5005` - listen on all addresses (Java 8) or only localhost (Java 9+)
`-jvm-debug localhost:5005` - works the same on all Java versions
`-jvm-debug '*:5005'` - listen on all addresses for Java 9+ (quotes needed in some shells to prevent * expansion)

That means we can remove all the Java version detection. We should also revert #1546 to keep bash and ash consistent.

qwe2:
Sounds good, but this would be a breaking change then:

> `-jvm-debug 5005` - listen on all addresses (Java 8) or only localhost (Java 9+)

This currently means `*:5005` on JDK 9+, but going forward, it would mean `localhost:5005` instead. Is that acceptable?

muuki88:
I'm with @dwickern on the secure setting by default and also reflecting what the JVM does. If you are not using sbt-native-packager and you use `-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005` it will also break when you upgrade the JVM. 

After all, upgrading a major version of the JVM must be tested and I think it's better if sbt-native-packager doesn't try to magically add a migration path.

qwe2:
I've reverted https://github.com/sbt/sbt-native-packager/pull/1546 and removed the major version checks. Note that this would be a breaking change for anyone currently using `-jvm-debug 5005` on JDK 9+. They would have to change it to `-jvm-debug '*:5005'`.

muuki88:
Thanks for you patience @qwe2 and thanks for stating the explicit breaking  change.

>  Note that this would be a breaking change for anyone currently using `-jvm-debug 5005` on JDK 9+

I have never used this flag before. My assumption would be, that this is only used in testing environments and not in production, right?

qwe2:
> that this is only used in testing environments and not in production, right?

Yeah, I think that nobody _should_ be using this flag in production. (Unfortunately that doesn't necessarily mean that nobody does 😞 )

muuki88 (approved):
@dwickern do you have any remarks? Otherwise I would merge and release this change :tada:

PR #1646, opened 2024-11-09
ci: Update GitHub Actions to setup-java

eed3si9n:
## Problem

setup-scala is deprecated since setup-java can set up the JDK, and configure the cache.

## Solution

This brings up the GitHub Actions to a modern set, and collapsing down the duplicated settings for the scripted jobs by using a matrix instead.

### DIFF
--- a/.github/workflows/validate-pr.yml
+++ b/.github/workflows/validate-pr.yml
@@ -2,299 +2,146 @@ name: Validate PR
 
 on:
   pull_request:
+  push:
 
 jobs:
+  # Run the unit test ahead of the scripted tests
   validate:
-    
-    runs-on: ubuntu-latest 
     strategy:
       fail-fast: false
-
+      matrix:
+        include:
+          - os: ubuntu-latest
+          - os: macos-latest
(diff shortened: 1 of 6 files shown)


.github/workflows/validate-pr.yml
```diff
@@ -2,299 +2,146 @@ name: Validate PR
...
+  scripted:
     strategy:
       fail-fast: false
-    needs: [ validate ]
+      matrix:
+        include:
+          - os: ubuntu-latest
+            java: 8
+            distribution: zulu
+            jobtype: 1
```
muuki88, line 37: I do like the refactoring to the matrix definition. 

What about putting the sbt command argument here directly instead of the if else stuff below? Would this work?

muuki88 (approved):

muuki88:
Thanks a lot @eed3si9n 🙏😊

.github/workflows/validate-pr.yml
eed3si9n: Good idea. 8213f48920feb81847f50071fe54b02df6d901a7

PR #1651, opened 2024-11-22
Fix multi-platform builds with multiple aliases.

zarthross:
Multi-platform builds append a `-t` argument for each alias to the `dockerBuildOptions`, but the `publishTask` function was ALWAYS iterating over the aliases even if it was a multi-platform build.

https://github.com/sbt/sbt-native-packager/blob/7c6885910c382784bcd2cd9fc63c2794a9dbd0a2/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala#L157-L159

The consequence was attempts to buildx and push the tags multiple times. In some environments this leads to build failures because you can't push the same tag twice.

This change simply does not iterate over the aliases for multiplatform builds.

Note: using `dockerUpdateLatest := true` counts as having multiple aliases, which is how I encountered this bug.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
@@ -300,9 +300,12 @@ object DockerPlugin extends AutoPlugin {
               "--push"
             ) ++ dockerBuildOptions.value :+ "."
           else dockerExecCommand.value
-        alias.foreach { aliasValue =>
-          publishDocker(context, execCommand, aliasValue.toString, log, multiplatform)
-        }
+        // For multiplatform builds, the alias are part of the `dockerBuildOptions` already.
+        if (multiplatform) publishDocker(context, execCommand, dockerAlias.value.tag.getOrElse(""), log, multiplatform)
+        else
+          alias.foreach { aliasValue =>
+            publishDocker(context, execCommand, aliasValue.toString, log, multiplatform)
+          }
       } tag (Tags.Network, Tags.Publish)
 
     def cleanTask =


lightbend-cla-validator:
Hi @zarthross,

We see that you have signed the Akka Contributors License Agreement before, however, the CLA has changed since you last signed it.

zarthross:
The CLA signer process is broken... I cannot sign it... not sure what ya'll want to do about it.

zarthross:
This is basically a 1 line fix, i do not care about the CLA, i do not care about copy right to the code, just wanted others to have the fix.....

dwickern:
The CLA bot has an insecure http: URL... This link should work: https://www.lightbend.com/contribute/cla

muuki88 (approved):

src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala
```diff
@@ -300,9 +300,12 @@ object DockerPlugin extends AutoPlugin {
               "--push"
             ) ++ dockerBuildOptions.value :+ "."
           else dockerExecCommand.value
-        alias.foreach { aliasValue =>
-          publishDocker(context, execCommand, aliasValue.toString, log, multiplatform)
-        }
+        // For multiplatform builds, the alias are part of the `dockerBuildOptions` already.
+        if (multiplatform) publishDocker(context, execCommand, dockerAlias.value.tag.getOrElse(""), log, multiplatform)
```
muuki88, line 304: That empty string is weird, but the `publishDocker` method does not use it for multiplatform - so this works :grin:

muuki88:
We'll get rid of the CLA after https://github.com/sbt/sbt-native-packager/discussions/1653

muuki88:
Thanks a lot :pray:

zarthross:
@muuki88 Thanks for merging!

PR #1665, opened 2025-01-06
🐛 fix: use exec in forwarder scripts (#1664)

### LINKED ISSUE #1664: Forwarder script do not use "exec"
## Expected behaviour

When using the [forwarder script mode](https://sbt-native-packager.readthedocs.io/en/latest/archetypes/java_app/index.html#single-start-script-with-forwarders) in case of multiple main classes, the forwarder scripts should use `exec` or any other way so that the java process is not a "child" of the bash process.

Otherwise, signals (SIGTERM) are not propagated correctly to the java process.

## Actual behaviour

Given, a simple Scala application with at least 2 main classes `com.xxx.Main1` and `com.xxx.Main2` and the following `build.sbt`:
```scala
// Triggers the "forwarder scripts" mode
Compile / mainClass := Some("com.xxx.Main1")
```

Then, 2 scripts are generated: `main1` and `main2`.
- `main1` uses `exec` and thus the PID of the java process is the PID of the initial `main1` process
- `main2` calls `main1` but leaves a `bash` wrapper process

**`main1`**

```bash
#!/usr/bin/env bash

# ...

execRunner () {
  # ...
  # we use "exec" here for our pids to be accurate.
  exec "$@"
}

# ...

run() {
  # ...
  execRunner "$java_cmd" \
    ${java_opts[@]} \
    "${java_args[@]}" \
    -cp "$(fix_classpath "$app_classpath")" \
    "${mainclass[@]}" \
    "${app_commands[@]}" \
    "${residual_args[@]}"
}

# ...

declare -a app_mainclass=('com.xxx.Main1')

# ...

run "$@"
```

**`main2`**

```bash
#!/usr/bin/env bash

# ...

# execute the main start script
$SCRIPTPATH/main1 -main com.xxx.Main2 "$@"
```

Processes when running `main1`:
```
$ ps -wauxf
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
xxxxxx+       1 41.9  1.4 29654076 469704 ?     Ssl  15:56   0:30 /opt/java/openjdk/bin/java ...
```

Processes when running `main2`:
```
$ ps -wauxf
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
xxxxxx+       1  0.0  0.0   7372  3628 ?        Ss   16:03   0:00 bash /app/bin/main2 ...
xxxxxx+       7  256  0.9 28590140 317244 ?     Sl   16:03   0:07 /opt/java/openjdk/bin/java ...
```

## Information

- What sbt-native-packager are you using: 1.11.0
- What sbt version: 1.10.7
- What is your build system (e.g. Ubuntu, MacOS, Windows, Debian ): Linux (Fedora)
- What package are you building (e.g. docker, rpm, ...): `JavaAppPackaging` raw files (no docker, no rpm)
- What version has your build tool (find out with e.g. `rpm --version`): N/A
- What is your target system (e.g. Ubuntu 16.04, CentOS 7): Linux (CentOS)

## Workaround

A workaround is to _not_ use a forwarder script but a distinct script for each main class by setting `Compile / mainClass := None`.

This way, each script uses the `exec` approach.

## Additional notes

The example above is simplified, in my real-life use case, I'm using Play Framework which sets the `Compile / mainClass := Some("play.core.server.ProdServerStart")` and I'd prefer to not have to override myself the `Compile / mainClass` and `Compile / discoveredMainClasses` (since the Play class is not automatically found, I'd have to declare it explicitly).

gaeljw:
Fix #1664

I tested on a personal case but please consider I'm not super experienced in Bash/SH, thus I may be missing a reason for which it wouldn't be the desired behaviour.

### DIFF
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/ash-forwarder-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/ash-forwarder-template
@@ -58,4 +58,4 @@ SCRIPT=$(realpath "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
 
 # execute the main start script
-$SCRIPTPATH/${{startScript}} -main ${{qualifiedClassName}} "$@"
+exec $SCRIPTPATH/${{startScript}} -main ${{qualifiedClassName}} "$@"
--- a/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/bash-forwarder-template
+++ b/src/main/resources/com/typesafe/sbt/packager/archetypes/scripts/bash-forwarder-template
@@ -58,4 +58,4 @@ SCRIPT=$(realpath "$0")
 SCRIPTPATH=$(dirname "$SCRIPT")
 
 # execute the main start script
-$SCRIPTPATH/${{startScript}} -main ${{qualifiedClassName}} "$@"
+exec $SCRIPTPATH/${{startScript}} -main ${{qualifiedClassName}} "$@"


gaeljw:
The check "Validate PR / scripted (ubuntu-latest, 8, zulu, +validateRpm)" is failing but I don't think it's related to my changes, is it? The error is not super clear to me.

dwickern:
It's not related to your changes, there's an issue with sbt 2.x: https://github.com/sbt/sbt/issues/7873

dwickern (approved):
I agree we should use `exec` here. Thanks!

PR #1705, opened 2025-05-31
Fix typo in "Skip packageDoc" section

esamson:
Setting needs to be set in the `Compile` configuration, not the `compile` task.


muuki88:
Thanks ❤️

PR #1706, opened 2025-06-03
Update customize.rst.txt

frosforever:
Update docs to reference `-J` requirements in `javaOptions` as a prefix not a suffix to the setting.

Thanks in advance for reviewing and for maintaining this library!


guizmaii (approved):

muuki88:
Thanks a lot for keeping the docs correct and helpful 🙏

Issue #1105, opened 2018-02-28
[Docker] Use Google "distroless" images


NeQuissimus:
https://very-serio.us/2018/02/27/zen-minimalism-with-distroless-images/

It may be interesting to have these base images as an option.
Due to their nature, they would require a separate startup script etc.

muuki88:
Thanks for the information :smile: 

The multistage build shouldn't be an issue as this is a docker "internal" feature.
You can define the `dockerCommands` freely.

> Due to their nature, they would require a separate startup script etc.

Can you elaborate on this? I haven't any production experience with docker (only  for small dev environments). What changes would be required? What can't be done with current native-packager
implementation and what is boilerplate that we should provide?

NeQuissimus:
The startup scripts currently depend on either `bash` or `ash`. Neither are available in these distroless images. Applications are meant to be started with the `java` command directly.

When I have some free time, I may do some more research into the possibility of integrating the images.

muuki88:
Sounds great 🤩 thanks for spending your spare time on this if possible.

I guess the start command can be built from the classpath and `javaOptions`.

jamesward:
I got this working with a bit of hackery:
https://github.com/jamesward/hello-uzhttp/blob/master/build.sbt#L25-L37

Note the use of the `LauncherJarPlugin`

muuki88:
Thanks a lot for sharing :hugs: @jamesward Even with the new docker layering implementation :nerd_face: 

So this doesn't require toooo much, right?

- new `dockerBaseImage`
- `LauncherPlugin` enabled
- `ExecCmd` altered

This can be an archetype plugin that applies these settings, requires the `LauncherPlugin` and then this should work. Probably I'm missing something here :thinking:

jamesward:
Yup! It'd be great to have support for this somehow built-in.  Note that some of the usual user permission & intermediate layer removal stuff don't work due to the absence of usual unix tools in the base image.

jamesward:
Oh, and the `ENTRYPOINT` on `gcr.io/distroless/java` basically is `java -jar` so that is why just the jar file needs to be specified as the `CMD` but I'm not sure if that is a pattern for other slim Java base containers.

ppiotrow:
I was planing check those distroless images this weekend byt @jamesward was first :)
I played around and tied his code with the DockerPlugin.
https://github.com/jamesward/hello-uzhttp/pull/1

jamesward:
[...]

Issue #1679, opened 2025-02-05
Recommend changing the default `openjdk:8` Docker image


dwickern:
https://github.com/sbt/sbt-native-packager/blob/eecea014304e4d4d9be11f3c87b248756ba9c3f9/src/main/scala/com/typesafe/sbt/packager/docker/DockerPlugin.scala#L84

Users should change openjdk to something else. https://hub.docker.com/_/openjdk has some alternatives.

We could remove the default value (breaking change) or add some validation to warn the user

muuki88:
Fair point. Is there a good other default? Would the eclipse-temurin image be a viable solution as this is the only one backed by an open source org?

mzuehlke:
`eclipse-temurin:8` seems likely to be the most minimal change.

mzuehlke:
> Let's go for `eclipse-termurin:8` then. I mean... Java8 is EOL anyways, right?

Oracle will support it till December 2030 🙃

dwickern:
I have a hard time choosing a default... I feel like the user needs to choose which distribution and especially which JVM version

If we simply removed the default value, it would break users on `openjdk:8` which seems like a good thing. Would it break any other use-cases?

SethTisue:
I would suggest defaulting to at least 17. You could make a good argument for defaulting to latest-LTS (currently 21, soon to be 25).

PR #1726, opened 2025-09-19
Build distroless images using the DockerPlugin

nMoncho:
Hello,
I'd like to submit this PR to allow users create [Distroless](https://github.com/GoogleContainerTools/distroless) images. This was mentioned/requested in #1105.

After playing around with the keys provided by the `DockerPlugin`, I think this is the easiest approach to maintain. My original approach duplicated a lot of code from the existing plugin. With the approach outlined in this PR the work is mostly done by defining presets for existing Settings and Tasks.

I didn't add any tests nor documentation, as I hope this can be the starting point of a conversation.

### DIFF
--- a/src/main/scala/com/typesafe/sbt/packager/distroless/DistrolessPlugin.scala
+++ b/src/main/scala/com/typesafe/sbt/packager/distroless/DistrolessPlugin.scala
@@ -1,0 +1,262 @@
+package com.typesafe.sbt.packager
+package distroless
+
+import com.typesafe.sbt.SbtNativePackager.Universal
+import com.typesafe.sbt.packager.Keys.{defaultLinuxInstallLocation, maintainer, scriptClasspath}
+import com.typesafe.sbt.packager.archetypes.scripts.*
+import com.typesafe.sbt.packager.docker.DockerPlugin.UnixSeparatorChar
+import com.typesafe.sbt.packager.docker.{Cmd, CmdLike, DockerPlugin, ExecCmd}
+import sbt.*
+import sbt.Keys.{discoveredMainClasses, mainClass, mappings, streams}
+
+/**
+  * ==Distroless Plugin==
+  *
+  * This plugin helps you build docker containers with distroless images
+  *
+  * ==Configuration==
+  *
+  * In order to configure this plugin take a look at the available [[com.typesafe.sbt.packager.docker.DockerKeys]]
+  *
+  * ==Requirements==
+  *
+  * You need docker to have docker installed on your system and be able to execute commands. Check with a single
+  * command:
+  *
+  * {{{
+  * docker version
+  * }}}
+  *
+  * @example
+  *   Enable the plugin in the `build.sbt`
+  *   {{{
+  *    enablePlugins(DistrolessPlugin)
+  *   }}}
+  */
+object DistrolessPlugin extends AutoPlugin {
+
+  override def requires: Plugins = DockerPlugin
+
+  object autoImport extends Keys
+
+  import DistrolessPlugin.autoImport.*
+  import DockerPlugin.autoImport.*
+
+  override def projectConfigurations: Seq[Configuration] = Seq()
+
+  // Some of the default values are now provided in the global setting based on
+  // sbt plugin best practice: https://www.scala-sbt.org/release/docs/Plugins-Best-Practices.html#Provide+default+values+in
+  override lazy val globalSettings: Seq[Setting[?]] =
+    Seq(
+      dockerBaseImage := "gcr.io/distroless/java",
+      dockerEntrypoint := Seq("/usr/bin/java"),
+      distrolessDebuggerPort := None
+    )
+
+  override lazy val projectSettings: Seq[Setting[?]] = Seq(dockerCommands := {
+    val base = dockerBaseImage.value
+    val baseDirectory = (Docker / defaultLinuxInstallLocation).value
+    val dependencies = (Docker / scriptClasspath).value
+    val debugPort = distrolessDebuggerPort.value
+    val compileMainClass = (Compile / mainClass).value
+    val compileDiscoveredMainClasses = (Compile / discoveredMainClasses).value
+    val log = streams.value.log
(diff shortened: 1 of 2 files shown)


src/main/scala/com/typesafe/sbt/packager/distroless/DistrolessPlugin.scala
```diff
@@ -0,0 +1,262 @@
...
+  import DockerPlugin.autoImport.*
+
+  override def projectConfigurations: Seq[Configuration] = Seq()
+
+  // Some of the default values are now provided in the global setting based on
+  // sbt plugin best practice: https://www.scala-sbt.org/release/docs/Plugins-Best-Practices.html#Provide+default+values+in
+  override lazy val globalSettings: Seq[Setting[?]] =
+    Seq(
+      dockerBaseImage := "gcr.io/distroless/java",
+      dockerEntrypoint := Seq("/usr/bin/java"),
```
nMoncho, line 52: Initially I didn't want to override the entrypoint defined by the image, but by default is `java -jar`, expecting a single jar as CMD. This limits its use to simple JARs with no libraries, or to FatJAR/UberJARs

src/main/scala/com/typesafe/sbt/packager/distroless/DistrolessPlugin.scala
```diff
@@ -0,0 +1,262 @@
...
+      distrolessDebuggerPort := None
+    )
+
+  override lazy val projectSettings: Seq[Setting[?]] = Seq(dockerCommands := {
```
nMoncho, line 56: Distroless images cannot run any script, like the on coming from `BashStartScriptPlugin`. This forces most of the configuration to be done at build time.
For example, the JDWP debugger must be specified when building so its corresponding cli argument can be included in the CMD command in the Dockerfile

src/main/scala/com/typesafe/sbt/packager/distroless/DistrolessPlugin.scala
```diff
@@ -0,0 +1,262 @@
...
+    val compileDiscoveredMainClasses = (Compile / discoveredMainClasses).value
+    val log = streams.value.log
+
+    val layerMappings = (Docker / dockerLayerMappings).value
+    val layerIdsAscending = layerMappings.map(_.layerId).distinct.sortWith { (a, b) =>
+      // Make the None (unspecified) layer the last layer
+      a.getOrElse(Int.MaxValue) < b.getOrElse(Int.MaxValue)
+    }
+
+    val cmd = if (dockerCmd.value.isEmpty) {
```
nMoncho, line 71: In my original approach I duplicated `dockerCmd` to be an `Option[Seq[String]]`. This way users could define an empty CMD, or leave alone with a `None`. This in turn allowed this code to know explicitly what the user wanted.

With this preset, even if the user wants an empty CMD, the plugin will still calculate it and include in the Dockerfile. This, I feel, is a decent compromise. With a distroless image the CMD would probably always be defined to at least specify which JAR the user wants to run

src/main/scala/com/typesafe/sbt/packager/distroless/DistrolessPlugin.scala
```diff
@@ -0,0 +1,262 @@
...
+
+  /**
+    * @param exposedPorts
+    * @return
+    *   if ports are exposed the EXPOSE command
+    */
+  private final def makeExposePorts(
+    exposedPorts: Seq[Int],
+    exposedUdpPorts: Seq[Int],
+    jdwpPort: Option[Int]
```
nMoncho, line 164: This method has a slight change from the one in `DockerPlugin`. If the JDWP port is specified, then it's added to the exposed list.

src/main/scala/com/typesafe/sbt/packager/distroless/DistrolessPlugin.scala
```diff
@@ -0,0 +1,262 @@
...
+
+  /**
+    * Tries to define what main class to call when running the distroless container
+    * @param mainClass
+    *   optional main class
+    * @param discoveredMainClasses
+    *   discovered main classes
+    * @return
+    */
+  private def defineMainClass(
```
nMoncho, line 227: This method is similar to `CommonStartScriptGenerator.generateStartScripts` but here we require a single main class as it has to be defined at build time. The user cannot specify it at runtime like it was possible with the script.

src/main/scala/com/typesafe/sbt/packager/distroless/DistrolessPlugin.scala
```diff
@@ -0,0 +1,262 @@
...
+        makeMaintainer((Docker / maintainer).value).toSeq :+
+        makeWorkdir(baseDirectory)
+    ) ++
+      layerIdsAscending.map { layerId =>
+        makeCopyLayerDirect(layerId, baseDirectory)
+      } ++
+      dockerLabels.value.map(makeLabel) ++
+      dockerEnvVars.value.map(makeEnvVar) ++
+      makeExposePorts(dockerExposedPorts.value, dockerExposedUdpPorts.value, distrolessDebuggerPort.value) ++
+      Seq(makeEntrypoint(dockerEntrypoint.value), makeCmd(cmd))
```
nMoncho, line 90: The commands defined by the plugin are a subset of the ones available on `DockerPlugin`.
These are the most important changes:
- There is not multistage build. Since this was based on `DockerPermissionStrategy`, which in some cases it requires a RUN while building, it doesn't make sense in this context as there is no way to run anything in a distroless image to change permissions.
- There are no exposed volumes. Currently that's implemented with several RUN commands. I guess that if it would be desired, then defining just `VOLUME` without the RUNs would be possible.
- There are no `makeUser` or USER. This is due to how users are provided/created by the base distroless image. There are different tags for root and nonroot users.

nMoncho:
This PR started to failed due to what's mentioned in #1679. Specifically https://github.com/sbt/sbt-native-packager/issues/1679#issuecomment-3486700178

PR #1744, opened 2026-02-19
Log docker commands on info instead of debug

LaurenceWarne:
I had an issue where I was using the wrong key to add a docker flag, I spent a while derping around before I found out the full command was output on `debug`.

It's very minor change, but I thought the docker command might be important enough to make logging it on `info` ok and thus hopefully make such issues more visible.

Thanks


muuki88 (approved):

muuki88:
Thanks, makes a lot of sense 😊

PR #1776, opened 2026-08-21
Restore `MappingsHelper` DSL ergonomics for sbt 2.x virtual-file mappings

### LINKED ISSUE #1775: SBT 2.x incompatible types
Hello folks,

Upgrading a project to SBT 2.x, plugin 1.11.7, I faced the following errors:

```
[error] an error in expression: class dotty.tools.dotc.reporting.Diagnostic$Error at /home/xxx/build.sbt:<5955..5955> L48: No Append.Values[Seq[(xsbti.HashedVirtualFileRef, String)], Seq[(sbt.File, String)]] found, so Seq[(sbt.File, String)] cannot be appended to Seq[(xsbti.HashedVirtualFileRef, String)].
[error] I found:
[error] 
[error]     sbt.Append.appendSeqImplicit[(xsbti.HashedVirtualFileRef, String), V](
[error]       /* missing */summon[Conversion[V, (xsbti.HashedVirtualFileRef, String)]])
[error] 
[error] But no implicit values were found that match type Conversion[V, (xsbti.HashedVirtualFileRef, String)].
[error] failed to connect to server
```

The `build.sbt` contains:
```scala
Universal / mappings ++= directory("src/conf")
```

This is consistent with the documentation at https://www.scala-sbt.org/sbt-native-packager/formats/universal.html#mapping-a-complete-directory.

I read a bit about some changes in SBT migration guide at https://www.scala-sbt.org/2.x/docs/en/changes/migrating-from-sbt-1.x.html?highlight=FileConverter#migrating-classpath-type but I'm not sure as a user of the plugin, how am I supposed to handle the error above?

AI suggested the following which seems to work:
```scala
Universal / mappings ++= {
  implicit val conv: FileConverter = fileConverter.value
  directory("src/conf").map { case (f, p) => conv.toVirtualFile(f.toPath) -> p }
}
```

Is it the right way? It feels a bit "heavy" to have to do this in all places.

gaeljw:
_Disclaimer: this was mostly generated by GitHub Copilot._

## Context

Aims to fix #1775.

sbt 2.x changed `mappings` to virtual-file keys, while `MappingsHelper.directory/contentOf` still produced `File`-keyed mappings. This broke the long-standing DSL usage and pushed users into manual `FileConverter`/mapping conversion code in `build.sbt`.

## Options to fix it

For the record, here are the options Copilot suggested:
- **Add sbt-2-specific overloads that return virtual-file mappings directly** :arrow_left: I picked this one
- Provide a plugin-level `given Conversion[File, HashedVirtualFileRef]` (or equivalent) in auto imports
- Wrap mapping append behind a dedicated DSL method (e.g. `addDirectoryMappings(...)`) and document migration
- Compatibility shim in `directory(...)` internals keyed by sbt version

## Implementation changes

- **Cross-build API split (sbt 1.x vs sbt 2.x)**
  - Moved `MappingsHelper` into versioned source partitions.
  - `scala-2.12` keeps existing behavior/signatures for sbt 1.x compatibility.
  - `scala-3` provides sbt-2-compatible overloads for the same helper names.

<details><summary>Diff SBT 1.x / 2.x</summary>
<p>

```diff
--- <sans nom>
+++ <sans nom>
@@ -20,8 +20,13 @@
     * @return
     *   mappings
     */
-  def directory(sourceDir: String): Seq[(File, String)] =
-    directory(file(sourceDir))
+  def directory(sourceDir: String): Def.Initialize[Seq[(FileRef, String)]] =
+    Def.setting {
+      implicit val conv: FileConverter = fileConverter.value
+      directory(file(sourceDir)).map { case (f, p) =>
+        toFileRef(f) -> p
+      }
+    }
 
   /**
     * It lightens the build file if one wants to give a string instead of file.
@@ -36,8 +41,11 @@
     * @return
     *   mappings
     */
-  def contentOf(sourceDir: String): Seq[(File, String)] =
-    contentOf(file(sourceDir))
+  def contentOf(sourceDir: String): Def.Initialize[Seq[(FileRef, String)]] =
+    Def.setting {
+      implicit val conv: FileConverter = fileConverter.value
+      contentOf(sourceDir = file(sourceDir), conv0 = conv)
+    }
 
   def contentOf(sourceDir: File, conv0: FileConverter): Seq[(FileRef, String)] = {
     implicit val conv: FileConverter = conv0

```

</p>
</details> 

- **sbt 2.x helper behavior**
  - `directory(String)` and `contentOf(String)` can now be used as in SBT 1.x with the extra addition of a `.value`

- **Scripted coverage update**
  - Created scripted tests folder specific to SBT 1.x and 2.x and the associated alias (run in the CI)
  - mapping helpers test moved from the common tests folder (SBT 1.x & 2.x) to one variant for each SBT version


gaeljw:
I thought Copilot had it but it turns out that the integration/scripted test are failing because:
```
[error] -- [E172] Type Error: /tmp/sbt_6c8c5de/build.sbt:14:62 -------------------------
[error] 14 |Universal / mappings ++= directory("src/main/resources/cache")
[error]    |                                                              ^
[error]    |No Append.Values[Seq[(xsbti.HashedVirtualFileRef, String)], sbt.Def.Initialize[Seq[(sbtcompat.PluginCompat.FileRef, String)]]] found, so sbt.Def.Initialize[Seq[(sbtcompat.PluginCompat.FileRef, String)]] cannot be appended to Seq[(xsbti.HashedVirtualFileRef, String)]
[info] [error] an error in expression: class dotty.tools.dotc.reporting.Diagnostic$Error at /tmp/sbt_6c8c5de/build.sbt:<5783..5783> L14: No Append.Values[Seq[(xsbti.HashedVirtualFileRef, String)], sbt.Def.Initialize[Seq[(sbtcompat.PluginCompat.FileRef, String)]]] found, so sbt.Def.Initialize[Seq[(sbtcompat.PluginCompat.FileRef, String)]] cannot be appended to Seq[(xsbti.HashedVirtualFileRef, String)]
```

One fix is to have `directory`/`contentOf` returns a `Def.Initialize[Seq[(File, String)]]` rather than a plain `Seq[(File, String)]` but it forces users to add an extra `.value` like `Universal / mappings ++= directory("src/main/resources/cache").value`. This doesn't feel great for users.

Then, I'm lost at how this should be handled ; I don't know enough of SBT 2.x to understand all of this.

Claude is suggesting something that apparently works but doesn't sound right ; it wants to use `val unmappedFileConverter: FileConverter = sbt.internal.inc.MappedFileConverter.empty`.

<details><summary>Claude's suggestion</summary>
<p>
Here's the mechanism end to end.

**The problem it solves**

Under sbt2, `Universal / mappings` is `Seq[(HashedVirtualFileRef, String)]` — not `Seq[(File, String)]` like sbt1. Turning a `File` into a `HashedVirtualFileRef` normally requires an `xsbti.FileConverter`, and the *only* one sbt normally hands you is the `fileConverter` **setting**, which you can only read inside a task/setting body via `fileConverter.value`. That's the trap: `directory("cache")` is called directly inside `build.sbt` as a plain expression passed to `++=`, not inside a `.value`-context, so it has no legal way to fetch that setting. That's exactly why the previous fix had to fall back to returning `Def.Initialize[...]` and forcing `.value` at every call site — which broke the "usage identical to before" requirement.

**The trick**

`unmappedFileConverter` sidesteps needing the *setting* at all, by constructing a `FileConverter` instance directly:

```scala
private val unmappedFileConverter: FileConverter = sbt.internal.inc.MappedFileConverter.empty
```

`MappedFileConverter.empty` is `new MappedFileConverter(rootPaths = Map.empty, allowMachinePath = true)` (confirmed by decompiling it). It's a fully standalone, stateless-enough object — no build context needed, just a plain `val`. Because `allowMachinePath = true`, when you ask it to convert a `Path` it doesn't recognize as belonging to any mapped root, it doesn't error out — it just wraps the *raw absolute path* as the virtual file's id (backslashes normalized to `/`). So `directory(...)`/`contentOf(...)` can call `unmappedFileConverter.toVirtualFile(file.toPath)` synchronously, right there in the plain method body, no task graph involved.

**Why the result is still safe to hand off**

The `HashedVirtualFileRef` this produces is a `MappedVirtualFile`, which implements `xsbti.PathBasedFile` — a sub-interface that exposes its own `.toPath()` directly. I checked the bytecode of `FileConverter.toPath(ref)` (the method any *other* converter uses to turn a ref back into a real path): the very first thing it does is `if (ref.isInstanceOf[PathBasedFile]) return ref.toPath()` — it short-circuits before even looking at its own root mappings.

So later, when the real packaging task does `fileConverter.value.toPath(ourRef)` to actually copy the file into the zip, it never consults its own (possibly different) root-mapping table at all — it just calls back into the ref we built, which already remembers the correct absolute path. That's why mixing "my throwaway converter at mapping-declaration time" with "the project's real converter at packaging time" is safe: the ref is self-describing, not dependent on which converter reads it.

**Net effect**

- `directory`/`contentOf` are ordinary functions again, callable straight from `build.sbt` exactly like sbt1 (`Universal / mappings ++= directory("cache")`, no `.value`).
- The `FileRef`s they emit are real, resolvable references to the actual files on disk, verified end-to-end by the scripted test (which unzips the built archive and checks the files actually landed at the right paths).
- Content hashing (`contentHashStr`) is still computed lazily by `MappedVirtualFile` on first access, so there's no eager file-hashing cost just from building the mapping list.


</p>
</details>

muuki88:
Thanks for giving this a shot. I agree with your assements, but am not much of help 😢 
I requested reinforcements 😁

eed3si9n:
It should use

```scala
val conv = fileConverter.value
```

Maybe define something like:

```scala
Universal / mappings ++= directory("src/main/resources/cache", fileConverter.value)
```

?

/cc @anatoliykmetyuk

gaeljw:
Is there no way we can preserve the original DSL for end users so that they don't have to know anything about "converter"?

eed3si9n:
Then you'd probably need:

```scala
Universal / mappings ++= directory("src/main/resources/cache").value
```

with `.value`?

gaeljw:
Thanks. I think the `.value` is more acceptable than the other options. If you all agree, I can take over the PR with this approach.

muuki88:
Tha absolutely works 👍  Thanks for fixing this

src/sbt-test/universal-sbt1/test-mapping-helpers/build.sbt
```diff
@@ -11,16 +11,10 @@ name := "simple-test"
 version := "0.1.0"
 
 // or just place your cache folder in /src/universal/
-Universal / mappings ++= {
-  implicit val converter: FileConverter = fileConverter.value
-  PluginCompat.toFileRefsMapping(directory("src/main/resources/cache"))
-}
+Universal / mappings ++= directory("src/main/resources/cache")
```
gaeljw, line 14: ℹ️  This restores the original syntax for SBT 1.x

src/sbt-test/universal-sbt2/test-mapping-helpers/build.sbt
```diff
@@ -0,0 +1,32 @@
...
+import xsbti.FileConverter
+
+enablePlugins(JavaAppPackaging)
+
+name := "simple-test"
+
+version := "0.1.0"
+
+// or just place your cache folder in /src/universal/
+Universal / mappings ++= directory("src/main/resources/cache").value
```
gaeljw, line 14: ℹ️  Notice the extra `.value` here to make it work in SBT 2.x

src/sphinx/recipes/sbt2-migration-mappings-helper.rst
```diff
@@ -0,0 +1,52 @@
+.. _sbt2-migration-mappings-helper:
+
+Updating MappingsHelper.directory/contentOf to SBT 2.x
```
gaeljw, line 3: This is a bit verbose, feedbacks appreciated.

gaeljw:
@muuki88 ready for review

gaeljw:
Oh actually there's 1 failing test because docker integration tests are run both from SBT 1.x and 2.x. I need to duplicate them in the respective folders.

EDIT: fixed.

muuki88 (approved):

muuki88:
Awesome! Looking good

